PR #83, opened 2015-12-09
Change 'enterprise-client' to 'usage-client'

timraymond:
`enterprise-client` will soon be repurposed for the on-prem solution, so we have renamed the existing `enterprise-client` to `usage-client`. It provides the same interface, but the backend endpoints for everything except anonymous usage reporting have been no-oped. I attempted to remove stats reporting from Kapacitor, however it seems that `client.StatsData` is used internally to facilitate writing TICK scripts against Kapacitor's own stats. Since this will require a bit more care than originally thought, for the time being this patch only switches the repos so that everything is reported to `usage.influxdata.com` instead of `enterprise.influxdata.com` (since we are repurposing that subdomain as well).

### DIFF
--- a/services/reporting/service.go
+++ b/services/reporting/service.go
@@ -7,19 +7,13 @@ import (
 	"sync"
 	"time"
 
-	"github.com/influxdb/enterprise-client/v1"
 	"github.com/influxdb/kapacitor"
+	"github.com/influxdb/usage-client/v1"
 )
 
 const reportingInterval = time.Hour * 12
 
-// Sends periodic information to Enterprise.
-// If not registered with Enterprise just
-// registers the server on startup and sends anonymous
-// stats every 12 hours.
-//
-// If registered with Enterprise also sends
-// all expvar statistics at the Config.StatsInterval.
+// Sends anonymous usage information every 12 hours.
 type Service struct {
 	tags client.Tags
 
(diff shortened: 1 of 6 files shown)


nathanielc:
@timraymond Thanks this looks good, yeah when I heard things were changing I realized I was going to need to do some rework. It won't be too hard. Can we get together for chat soon to make sure I understand the direction so I can make appropriate changes?

timraymond:
@nathanielc Certainly! I'll find you in Slack and we'll set something up.

services/reporting/service.go
```diff
@@ -39,9 +39,9 @@ type Service struct {
 	wg            sync.WaitGroup
 }
 
-func NewService(c Config, token string, l *log.Logger) *Service {
-	client := client.New(token)
-	client.URL = c.EnterpriseURL
+func NewService(c Config, l *log.Logger) *Service {
+	client := client.New("")
```
nathanielc: @timraymond  I don't need the token any more correct?

nathanielc:
@timraymond Overall is this what you expected.

services/reporting/service.go
timraymond: @nathanielc Yep, that's fine :scissors: . The empty string token is the same approach we've taken in Chronograf and Core for usage-only reporting.

timraymond:
@nathanielc Yep! This looks good to me :+1:

Issue #72, opened 2015-12-01
Custom User Functions


nathanielc:
Right now the only functions that can be applied to the data stream are the ones available in the InfluxQL language or can be expressed via lambda expressions. We need a way to allow users to define their own functions without having to compile them into Kapacitor.

Basic Plan so far:

* Use RPC and external processes.
* The RPC API will have basic Collect/Emit calls to stream data in and out off the process.
* Initial candidate for RPC is using gRPC
* Snapshot state periodically so that processes can resume where they left off after restart.

Questions:

* How to configure such a custom function and expose it via TICKscript, Maybe separate lookup for custom chaining methods based off registered custom functions.
* Which language do we want to support, the choice of RPC system will greatly effect this. Maybe we want a fast robust system like gRPC and a simple JSON one for other languages.

nathanielc:
I have a working prototype using gRPC. Initial thoughts:

* gRPC is very new and so installation and use is difficult since it requires building lots of pre-released code from source.
* There are not very many languages supported yet. Key languages are the scientific languages. C, Python, R, maybe even Fortran. 
* The streaming is very nice. Specially since the bi-directional streams are independent and the order of reads vs writes doesn't matter.

SkyRocknRoll:
How about adding support for lua runtime in kapacitor? 
If we add then the users can define their functions in lua . 
Heka https://hekad.readthedocs.org/en/v0.9.2/ allow user to write custom filter in lua and load it to run time. Heka also written in Go. 

https://github.com/mozilla-services/lua_sandbox lua snabox is opensoure.

nathanielc:
@SkyRocknRoll we thought a lot about using lua initially and it would be able to solve a  piece of the problem, but we also decided that it would not solve enough.

The goal of custom functions is not necessarily so that user can `write` their own functions but rather can `use` their existing algorithms/code for processing their data. Many users will already have their algorithms already written whatever language C/C++ Python etc and many of them will not be able to rewrite the code effectively in any other language. By creating a method for input and output of Kapacitor users can just create wrappers as needed.

Having multiple language support is important enough that we decided away from Lua.

nathanielc:
I finished a survey of serialization formats and RPC frameworks. Here are my thoughts:

Serialization:
[...]

PR #99, opened 2015-12-16
Initial User Defined Functions (UDFs) proto example. [WIP]

nathanielc:
Initial Protocol definition for communicating with child processes.

Includes a working example in `udf_example` where the go code spawns the python server and communicates with it via STDIN and STDOUT. 

The proto file here `udf/udf.proto` is the meat of this PR.

This example uses varint + protobuf encoded streams of data over the sockets to the process. 

I decided not to use the RPC framework in proto3 since using it would require writing our own version of gRPC. More thoughts on the design are in the issue #72.


udf/udf.proto
```diff
@@ -0,0 +1,183 @@
...
+// Meaning that ordering or synchronizing STDIN and STDOUT in anyway
+// is not necessary.
+// For example if Kapacitor requests a state snapshot and the
+// process is in the middle of writing a previous response or
+// data points those can continue. Eventually Kapacitor will receive
+// the state snapshot response and act accordingly.
+//
+// A KeepAliveRequest/KeepAliveResponse system is used to ensure that 
+// the process is responsive. Every time that a KeepAliveRequest is sent
+// a KeepAliveResponse must be returned within a timeout. If too many heartbeats
```
pauldix: I would modify this so that any message received by Kapacitor from STDOUT should reset the timer, because the process is obviously still alive?

udf/udf.proto
```diff
@@ -0,0 +1,183 @@
...
+}
+
+// Message containing information about a single data point.
+// Can be sent on it's own or bookended by BeginBatch and EndBatch messages.
+message Point {
+    int64 Time                       = 1;
+    string Name                      = 2;
+    string Database                  = 3;
+    string RetentionPolicy           = 4;
+    string Group                     = 5;
```
pauldix: What's group?

udf/udf.proto
```diff
@@ -0,0 +1,183 @@
...
+message BeginBatch {
+}
+
+// Message containing information about a single data point.
+// Can be sent on it's own or bookended by BeginBatch and EndBatch messages.
+message Point {
+    int64 Time                       = 1;
+    string Name                      = 2;
+    string Database                  = 3;
+    string RetentionPolicy           = 4;
```
pauldix: Would be good to have a `Points` message that could pull these variables out. That way Kapacitor could just send the batch (or chunks of it) in a single go without repeating all these.

udf/udf.proto
```diff
@@ -0,0 +1,183 @@
...
+// snapshot of the running state.
+message StateResponse {
+    int32 version  = 1;
+    bytes snapshot = 2;
+}
+
+// Request that the process restore its state from a snapshot.
+message RestoreRequest {
+    int32 version  = 1;
+    bytes snapshot = 2;
```
pauldix: Is it going to be easy in every language to send bytes? Maybe have it be a `oneof` of either string or bytes?

udf/udf.proto
nathanielc: Agreed

udf/udf.proto
nathanielc: Hmm, thats a good point, I'll look into it and see what I learned. For languages like python as string is a bytes so it doesn't really matter but I'll do a survey and see what I find.

udf/udf.proto
nathanielc: Its a unique identifier for the group by dimensions.  For example with  a group by `host` and `service` the `Group` would be something like `host=serverA,service=cart`. Each node has to treat groups independently based on its needs, some processing doesn't care since it applies a transformation uniformly while other transformations need to handle groups explicitly. So each point received is tagged with its `Group`.

udf/udf.proto
nathanielc: I like that idea, proto is sparse encoding so if the field is not set it doesn't take any space on the wire but maybe grouping things into messages would make that easier to manage. As opposed to having to check each field explicitly.

nathanielc:
This is nearly done. Just need to:

- [ ] Add docs for how to use UDFs, much of this already exists in udf.proto.
- [ ] Add tests for the agent code. I want a utility that can test your custom written UDFS external to the Kapacitor process so you can iterate on them easily.

Issue #71, opened 2015-12-01
Add OpsGenie as notifications system


panda87:
Please add output integration with OpsGenie.
If you need any help with their documentation or best practices for the API calls, just let me know.

D.

nathanielc:
Kapacitor's alerting system is simple: 

It tracks the state of an alert one of OK,INFO,WARN,CRITICAL 
It calls registered alert handlers whenever the state is not OK or when it has just changed to the state OK from a non OK state.
Looking at the OpsGenie api it seems to have a Create/Close API calls which map to the alert entering and leaving a non OK state. What about for changes in state like WARN->CRITICAL?

Contributing the handler is pretty straight forward:

1. Create a new service in the `services` package and implement methods for interacting with OpsGenie.
2. Add the OpsGenie service to the `cmd/kapacitord/run` package so the server can open/close and configure the service.
3. Update the AlertNode in `pipeline` so the TICKscripts can declare OpsGenie options
4. Update the AlertNode in `kapacitor` to register the handler when configured via the TICKscript.
5. Update the TaskMaster to have an interface for communicating with the OpsGenie service.

You can follow the examples of Slack, PagerDuty, and VictorOps as they are all very similar.

schmurfy:
instead of implementing every services out there that can receive alerts it would be wiser and easier to maintain to implement something flexible enough to allow doing it  directly in a user script. I imagine all these services have an http(s) endpoint and except some kind of parameters in the query or in the post body.

nathanielc:
@schmurfy I agree that implementing every service out there will get out of hand. Right now it is easier to just implement them directly in Go instead of building a sandbox env to do so. Already the differences between the implemented services is enough that a simple HTTP POST template with URLs will not be sufficient. For now simplicity wins. That said the current organization of each Alert Handler will not scale and I am looking into ways to clean that up.

PR #113, opened 2015-12-22
added OpsGenie service for alerting

ericiles:
Implements #71 

I am running this code in our production environment, and have been getting OpsGenie alerts since 12/18/2015.

Usage:
```
stream
	.from().measurement('cpu')
	.groupBy('host')
	.window()
		.period(10s)
		.every(10s)
	.alert()
		.info(lambda: "count" > 6.0)
		.warn(lambda: "count" > 7.0)
		.crit(lambda: "count" > 8.0)
		.opsGenie()
		.ogTeams('test_team,another_team')
		.ogRecipients('test_recipient,another_recipient')
```

Configuration:
```
[opsgenie]
    # Configure OpsGenie with your API key and default routing key.
    enabled = false
    # Your OpsGenie API Key.
    api-key = ""
    # Default OpsGenie teams, can be overridden per alert.
    teams = ""
    # Default OpsGenie recipients, can be overridden per alert.
    recipients = ""
    # The OpsGenie API URL should not need to be changed.
    url = "https://api.opsgenie.com/v1/json/alert"
    # The OpsGenie Recovery URL, you can change this
    # based on which behavior you want a recovery to
    # trigger (Add Notes, Close Alert, etc.)
    recovery_url = "https://api.opsgenie.com/v1/json/alert/note"
    # If true the all alerts will be sent to OpsGenie
    # without explicity marking them in the TICKscript.
    # The routing key can still be overridden.
    global = false
```

This is my first run at working with Go, so some changes may be necessary. This is mainly a copy of the VictorOps service, modified to work properly with OpsGenie.

- [ ] CHANGELOG.md updated
- [x] Rebased/mergable
- [X] Tests pass
- [X] Sign [CLA](http://influxdb.com/community/cla.html) (if not already signed)

### DIFF
--- a/etc/kapacitor/kapacitor.conf
+++ b/etc/kapacitor/kapacitor.conf
@@ -91,6 +91,26 @@ data_dir = "/var/lib/kapacitor"
   # Close idle connections after timeout
   idle-timeout = "30s"
 
+[opsgenie]
+    # Configure OpsGenie with your API key and default routing key.
+    enabled = false
+    # Your OpsGenie API Key.
+    api-key = ""
+    # Default OpsGenie teams, can be overridden per alert.
+    # teams = ["team1", "team2"]
+    # Default OpsGenie recipients, can be overridden per alert.
+    # recipients = ["recipient1", "recipient2"]
+    # The OpsGenie API URL should not need to be changed.
+    url = "https://api.opsgenie.com/v1/json/alert"
+    # The OpsGenie Recovery URL, you can change this
+    # based on which behavior you want a recovery to
+    # trigger (Add Notes, Close Alert, etc.)
+    recovery_url = "https://api.opsgenie.com/v1/json/alert/note"
+    # If true then all alerts will be sent to OpsGenie
+    # without explicity marking them in the TICKscript.
+    # The team and recipients can still be overridden.
+    global = false
+
 [victorops]
   # Configure VictorOps with your API key and default routing key.
   enabled = false
--- a/pipeline/alert.go
+++ b/pipeline/alert.go
@@ -19,7 +19,7 @@ const defaultMessageTmpl = "{{ .ID }} is {{ .Level }}"
 // See AlertNode.Info, AlertNode.Warn, and AlertNode.Crit below.
 //
 // Different event handlers can be configured for each AlertNode.
-// Some handlers like Email, Slack, VictorOps and PagerDuty have a configuration
+// Some handlers like Email, Slack, OpsGenie, VictorOps and PagerDuty have a configuration
 // option 'global' that indicates that all alerts implicitly use the handler.
 //
 // Available event handlers:
@@ -29,6 +29,7 @@ const defaultMessageTmpl = "{{ .ID }} is {{ .Level }}"
 //    * email -- Send and email with alert data.
 //    * exec -- Execute a command passing alert data over STDIN.
 //    * Slack -- Post alert message to Slack channel.
+//	  * OpsGenie -- Send alert to OpsGenie.
 //    * VictorOps -- Send alert to VictorOps.
 //    * PagerDuty -- Send alert to PagerDuty.
 //
@@ -169,6 +170,18 @@ type AlertNode struct {
 	// Log JSON alert data to file. One event per line.
 	Log string
 
+	// Send alert to OpsGenie.
+	// tick:ignore
+	UseOpsGenie bool
+
+	// OpsGenie Teams.
+	// tick:ignore
+	OpsGenieTeams []string
+
+	// OpsGenie Recipients.
+	// tick:ignore
+	OpsGenieRecipients []string
+
 	// Send alert to VictorOps.
 	// tick:ignore
 	UseVictorOps bool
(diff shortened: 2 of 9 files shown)


etc/kapacitor/kapacitor.conf
```diff
@@ -91,6 +91,26 @@ data_dir = "/var/lib/kapacitor"
...
+    recipients = ""
+    # The OpsGenie API URL should not need to be changed.
+    url = "https://api.opsgenie.com/v1/json/alert"
+    # The OpsGenie Recovery URL, you can change this
+    # based on which behavior you want a recovery to
+    # trigger (Add Notes, Close Alert, etc.)
+    recovery_url = "https://api.opsgenie.com/v1/json/alert/note"
+    # If true the all alerts will be sent to OpsGenie
+    # without explicity marking them in the TICKscript.
+    # The routing key can still be overridden.
```
nathanielc: This should say `team and recipients` instead of `routing key`.

services/opsgenie/service.go
```diff
@@ -0,0 +1,125 @@
...
+	if teams == "" {
+		teams = s.teams
+	}
+
+	if recipients == "" {
+		recipients = s.recipients
+	}
+
+	if teams != "" {
+		ogData["teams"] = strings.Split(teams, ",")
```
nathanielc: Can we make the `teams` and `recipients` vars a list of strings instead of splitting on commas?

pipeline/alert.go
```diff
@@ -169,6 +170,18 @@ type AlertNode struct {
 	// Log JSON alert data to file. One event per line.
 	Log string
 
+	// Send alert to OpsGenie.
+	// tick:ignore
+	UseOpsGenie bool
+
+	// VictorOps RoutingKey.
+	// tick:ignore
+	OpsGenieTeams string
```
nathanielc: To make it a list use

`OpsGenieTeams []string`

pipeline/alert.go
```diff
@@ -270,6 +283,71 @@ func (a *AlertNode) Flapping(low, high float64) Node {
...
+// Send alert to OpsGenie using the default teams, found in the configuration.
+// tick:property
+func (a *AlertNode) OpsGenie() *AlertNode {
+	a.UseOpsGenie = true
+	return a
+}
+
+// The OpsGenie Teams. If not set uses key specified in configuration.
+// tick:property
+func (a *AlertNode) OgTeams(teams string) *AlertNode {
```
nathanielc: Then here change the function to:

```
func (a *AlertNode) OgTeams(teams ...string) *AlertNode {
    a.OpsGenieTeams = teams
    return a
}
```

The `...` is Go for multiple args and `teams` will be a list of each of the args.

Then in TICKscript it will look like this:

```
.ogTeams('team1', 'team2', 'team,name,with,commas')
```
 Same for recipients.

services/opsgenie/service.go
```diff
@@ -0,0 +1,125 @@
...
+
+func (s *Service) Close() error {
+	return nil
+}
+
+func (s *Service) Global() bool {
+	return s.global
+}
+
+func (s *Service) Alert(teams, recipients, messageType, message, entityID string, t time.Time, details interface{}) error {
```
nathanielc: This will need to change as well:

```
func (s *Service) Alert(teams, recipients []string , messageType, message, entityID string, t time.Time, details interface{}) error {
```

nathanielc:
@ericiles Thanks for the great work!

Overall this looks great, I added some pointers on using a list of string instead of comma separated strings for splitting.

Not sure whether I like 
```
.alert()
   ...
   .opsGenie()
     .ogTeams('team1')
```

or 
```
.alert()
   ...
   .opsGenie()
     .teams('team1')
```

better. Thoughts?

Right now everything is flattened on the AlertNode, but this will not scale so I have been thinking of was to scope it better. Long term I would prefer the second option.

ericiles:
@nathanielc 

I agree that 
```
.alert()
   ...
   .opsGenie()
     .teams('team1')
```
would be the better way to go, but saw that doing so would possibly conflict with anything else that wanted to use teams or recipients, which is why I prefixed it with "og"

I am making the suggested changes for the teams and recipients. Also, I have a few things I hard coded in for our use, that I am going to go ahead and make configurable for the end user (the OverwritesQuietHours tag for Critical alerts, for example). I am going to make .ogTags('tag1', 'tag2') an option so that a user can specify whichever tags they wish.

Also, at the moment, on recovery, I am setting the note field to the message. This is fine if the user actually wants to send a note, instead of doing something else, such as closing the alert, but I think there should probably be a better way to allow the user to specify which parameters/values to send to OpsGenie along with their API call. But I am not exactly sure the best way to structure that portion. Possibly something like:
```
.alert()
   ...
   .opsGenie()
     .ogRecovery('{{ .Level}}: {{ .Name }}/{{ index .Tags "host" }} has normal cpu usage: {{ index .Fields "used" }}')
```
and then in the kapacitor.conf specifying the parameter that the above text would be sent as, since we already have the recovery_url in the config to specify the api url. Something like:
```
recovery_url = "https://api.opsgenie.com/v1/json/alert/note"
recovery_param = "note"
```
Any Suggestions?

nathanielc:
For now lets remove the `og` prefix from the calls. I know it could conflict but I plan to change that soon.

 I would like the templates to be global to all alert handlers if possible, and then each handler uses the appropriate template for its needs. This way it is clear what is available and what the purpose of each template is.

Right now there are two templates:

* ID -- a unique ID for the alert doesn't change when the alert state changes
* Message -- a short message that can change as the alert state changes

I plan to add another:

* Details -- a much longer message, possibly html etc, for displaying more information on the alert. 

For example the email alert handler just uses Message as the email subject and will use the Details as the body of the email. The email handler just ignores the ID while VictorOps/PagerDuty will probably ignore the Details template.

We could add a generic `RecoveryMessage` template that OpsGenie and others could use. I am not familiar with the OpsGenie workflow, why does the RecoveryMessage need to be different than the normal message? Wouldn't a generic `{{ .ID }} is {{ .Level }}` message work?

ericiles:
The message field would work fine for us, and is what we currently have it doing. However, other users could possibly wish to go with a different workflow for their alerting with OpsGenie. Using the API, on recovery, one could simply add a comment to the alert, acknowledge an alert, close an alert, add/remove tags to/from the alert, or simply ignore the recovery all together and do nothing. 

However, in the interest of keeping things uniform and simple, it is probably best that it just pass the message along as the 'note' parameter to whichever recovery_url they specify, as the note parameter is used for adding comments, closing (with a comment), and Acknowledging (with a comment). That would cover what a majority of users likely would want.

Also, I will go ahead and rename .ogTeams, .ogRecipients, etc. as requested.

nathanielc:
Ok, in that case lets go with the simple approach that solves the current use case and then if a different one shows up we can address it then. Again, thanks for this great PR.

ericiles:
The discussed changes have been completed. I ran this in our environment and triggered a few alerts, and all appears to be working as expected.

services/opsgenie/service.go
```diff
@@ -0,0 +1,119 @@
...
+			return err
+		}
+		ogData["description"] = string(b)
+	}
+
+	if len(teams) == 0 {
+		teams = s.teams
+	}
+
+	if len(teams) > 0 {
```
nathanielc: Should this be an error if no teams where specified?Or does OpsGenie have a sane default when no teams are specified? Same goes for recipients...

nathanielc:
@ericiles Looks great! Can you rebase/squash and then this should be ready for merging. Thanks.

services/opsgenie/service.go
ericiles: teams and recipients are optional per their API. The reason being, you can configure default teams/recipients at the API level, and I'd imagine worst case, no actual notifications go out, but the alert is simply created in the system.

nathanielc: ok, a quick rebase/squash and I'll merge this. Thanks!

PR #120, opened 2015-12-23
Allow for multiple alert handlers of the same type

### LINKED ISSUE #118: Update AlertNode so that configuring alert handlers can be correctly scoped, and easily added.
Currently Alert handlers are configured directly on the AlertNode and it is one large flat scope. It would be nice if each alert where configured in its own scope so that you do not have to worry about name conflicts etc. 

Goals:
* Make defining multiple Alert handlers in TICKscript easy and intuitive.
* Make writing a new alert handler a simple and isolated task.
* Be able to support(from a code organization perspective) at least dozens of different handlers with ease.  The performance profile of multiple handler should already be linear.

nathanielc:
Fixes #118 

With this change each alert handler is appropriately scoped so you do not have to worry about name conflicts. You can now add multiple handlers of the same type. 

For example:

```
stream
	.from().measurement('cpu')
	.groupBy('host')
	.window()
		.period(10s)
		.every(10s)
	.alert()
		.info(lambda: "count" > 6.0)
		.warn(lambda: "count" > 7.0)
		.crit(lambda: "count" > 8.0)
		.slack()
			.channel('#alerts')
		.slack()
			.channel('@jim')
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@
 ### Release Notes
 
 ### Features
+- [#118](https://github.com/influxdb/kapacitor/issues/118): Can now define multiple handlers of the same type on an AlertNode.
+- [#113](https://github.com/influxdb/kapacitor/issues/113): OpsGenie support thanks! @ericiles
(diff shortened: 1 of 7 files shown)


nathanielc:
Wait to merge this until after #113 so that we can rebase and integrate OpsGenie in the same way.

Issue #119, opened 2015-12-23
[Feature] - Alert to HipChat


JulienChampseix:
Hi team,

Will be great if we could push alert to an hipchat room.
Do you think it's possible ?

Thanks,
Julien.

nathanielc:
This should be easy to add. Do you want to take a crack at it? Just follow the code for Slack.

Contributing the handler is pretty straight forward:

1. Create a new service in the services package and implement methods for interacting with HipChat.
2. Add the HipChat service to the cmd/kapacitord/run package so the server can open/close and configure the service.
3. Update the AlertNode in pipeline so the TICKscripts can declare HipChat options
4. Update the AlertNode in kapacitor to register the handler when configured via the TICKscript.
5. Update the TaskMaster to have an interface for communicating with the HipChat service.

PR #125, opened 2015-12-28
Added HipChat service for alerting.

ericiles:
Implements #119 

This code has been tested and has successfully sent alerts to HipChat as of 12/28/2015.

Usage:
```
stream
	.from().measurement('cpu')
	.groupBy('host')
	.window()
		.period(10s)
		.every(10s)
	.alert()
		.info(lambda: "count" > 6.0)
		.warn(lambda: "count" > 7.0)
		.crit(lambda: "count" > 8.0)
		.hipChat()
			.room('1234567')
			.token('testtoken1234567')
```

Configuration:
```
[hipchat]
  # Configure HipChat.
  enabled = false
  # The HipChat API URL. Replace subdomain with your
  # HipChat subdomain.
  url = "https://subdomain.hipchat.com/v2/room"
  # Visit https://www.hipchat.com/docs/apiv2
  # for information on obtaining your room id and
  # authentication token.
  # Default room for messages
  room = ""
  # Default authentication token
  token = ""
  # If true then all alerts will be sent to HipChat
  # without explicitly marking them in the TICKscript.
  global = false
```

- [ ] CHANGELOG.md updated
- [x] Rebased/mergable
- [X] Tests pass
- [X] Sign [CLA](http://influxdb.com/community/cla.html) (if not already signed)

### DIFF
--- a/pipeline/alert.go
+++ b/pipeline/alert.go
@@ -19,7 +19,7 @@ const defaultMessageTmpl = "{{ .ID }} is {{ .Level }}"
 // See AlertNode.Info, AlertNode.Warn, and AlertNode.Crit below.
 //
 // Different event handlers can be configured for each AlertNode.
-// Some handlers like Email, Slack, OpsGenie, VictorOps and PagerDuty have a configuration
+// Some handlers like Email, HipChat, Slack, OpsGenie, VictorOps and PagerDuty have a configuration
 // option 'global' that indicates that all alerts implicitly use the handler.
 //
 // Available event handlers:
@@ -28,6 +28,7 @@ const defaultMessageTmpl = "{{ .ID }} is {{ .Level }}"
 //    * post -- HTTP POST data to a specified URL.
 //    * email -- Send and email with alert data.
 //    * exec -- Execute a command passing alert data over STDIN.
+//    * HipChat -- Post alert message to HipChat room.
 //    * Slack -- Post alert message to Slack channel.
 //	  * OpsGenie -- Send alert to OpsGenie.
 //    * VictorOps -- Send alert to VictorOps.
@@ -189,6 +190,10 @@ type AlertNode struct {
 	// Send alert to Slack.
 	// tick:ignore
 	SlackHandlers []*SlackHandler
+
+	// Send alert to HipChat.
+	// tick:ignore
+	HipChatHandlers []*HipChatHandler
 
 	// Send alert to OpsGenie
 	// tick:ignore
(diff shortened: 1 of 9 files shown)


services/hipchat/service.go
```diff
@@ -0,0 +1,93 @@
...
+
+	//Generate HipChat API Url including room and authentication token
+	if room == "" {
+		room = s.room
+	}
+	if token == "" {
+		token = s.token
+	}
+
+	url := s.url + "/" + room + "/notification?auth_token=" + token
```
nathanielc: Should we URL escape the `room` and `token` vars so the config doesn't have to do things like `Test%20Room`? See the answer to this SO question. It has a good example of how to safely craft the URL.  http://stackoverflow.com/questions/13820280/encode-decode-urls

pipeline/alert.go
```diff
@@ -482,6 +487,79 @@ type PagerDutyHandler struct {
...
+// If the 'hipchat' section in the configuration has the option: global = true
+// then all alerts are sent to HipChat without the need to explicitly state it
+// in the TICKscript.
+//
+// Example:
+//    [hipchat]
+//      enabled = true
+//      url = "https://orgname.hipchat.com/v2/room"
+//      room = "Test%20Room"
+//		token = "9hiWoDOZ9IbmHsOTeST123ABciWTIqXQVFDo63h9"
```
nathanielc: nit, could you fix the tabbing here so it aligns with the rest? thanks

pipeline/alert.go
```diff
@@ -482,6 +487,79 @@ type PagerDutyHandler struct {
...
+// To allow Kapacitor to post to HipChat,
+// go to the URL https://www.hipchat.com/docs/apiv2 for
+// information on how to get your room id and tokens.
+//
+// Example:
+//    [hipchat]
+//      enabled = true
+//      url = "https://orgname.hipchat.com/v2/room"
+//      room = "4189212"
+//		token = "9hiWoDOZ9IbmHsOTeST123ABciWTIqXQVFDo63h9"
```
nathanielc: same nit here.

nathanielc:
@ericiles This is great work! Just a few small changes to make the URL creation safer and easier for the config and this should be good to go.

How do you like the changes to having lists of handlers? Did that help clean it up a bit for you?

ericiles:
Having lists of handlers made things much nicer. Definitely an improvement. I have made the requested changes.

services/hipchat/service.go
ericiles: In a test I did, it appears that the http.Post automatically url encodes the url string. However, I didn't see this behavior documented, so to play it safe, I went ahead and implented the net/url method of parsing the URL in order to encode it, so that if the http.Post behavior changes in the future, it doesn't break this.

nathanielc: Great looks good.

PR #139, opened 2016-01-09
Alerta.io Integration

md14454:
Implements integration with [alerta.io](http://alerta.io)

The code has been tested with a local instance of alerta.

Usage:
```
stream
	.from().measurement('cpu')
	.groupBy('host')
	.window()
		.period(10s)
		.every(10s)
	.alert()
                .id('{{ index .Tags "host" }}')
		.message('kapacitor/{{ .Name }}/{{ index .Tags "host" }} is {{ .Level }}')
		.info(lambda: "count" > 6.0)
		.warn(lambda: "count" > 7.0)
		.crit(lambda: "count" > 8.0)
		.alerta()
			.token('testtoken1234567')
			.resource('serverA')
			.event('CPU Idle')
			.environment('production')
```

Configuration:
```
[alerta]
  # Configure Alerta.
  enabled = false
  # The Alerta URL.
  url = ""
  # Default authentication token.
  token = ""
  # Default environment.
  environment = ""
  # Default origin.
  origin = "kapacitor"
```

- [ ] CHANGELOG.md updated
- [ ] Rebased/mergable
- [x] Tests pass
- [x] Sign [CLA](http://influxdb.com/community/cla.html) (if not already signed)

### DIFF
--- a/alert.go
+++ b/alert.go
@@ -167,6 +167,11 @@ func newAlertNode(et *ExecutingTask, n *pipeline.AlertNode) (an *AlertNode, err 
 		n.IsStateChangesOnly = true
 	}
 
+	for _, alerta := range n.AlertaHandlers {
+		alerta := alerta
+		an.handlers = append(an.handlers, func(ad *AlertData) { an.handleAlerta(alerta, ad) })
+	}
+
 	for _, og := range n.OpsGenieHandlers {
 		og := og
 		an.handlers = append(an.handlers, func(ad *AlertData) { an.handleOpsGenie(og, ad) })
@@ -590,6 +595,52 @@ func (a *AlertNode) handleHipChat(hipchat *pipeline.HipChatHandler, ad *AlertDat
 	)
 	if err != nil {
 		a.logger.Println("E! failed to send alert data to HipChat:", err)
+		return
+	}
+}
+
+func (a *AlertNode) handleAlerta(alerta *pipeline.AlertaHandler, ad *AlertData) {
+	if a.et.tm.AlertaService == nil {
+		a.logger.Println("E! failed to send Alerta message. Alerta is not enabled")
+		return
+	}
+
+	var severity string
+	var status string
+
+	switch ad.Level {
+	case OKAlert:
+		severity = "ok"
+		status = "closed"
+	case InfoAlert:
+		severity = "informational"
+		status = "open"
+	case WarnAlert:
+		severity = "warning"
+		status = "open"
+	case CritAlert:
+		severity = "critical"
+		status = "open"
+	default:
+		severity = "unknown"
+		status = "unknown"
+	}
+
+	err := a.et.tm.AlertaService.Alert(
+		alerta.Token,
+		alerta.Resource,
+		alerta.Event,
+		alerta.Environment,
+		severity,
+		status,
+		alerta.Group,
+		alerta.Value,
+		ad.Message,
+		alerta.Origin,
+		ad.Data,
+	)
+	if err != nil {
+		a.logger.Println("E! failed to send alert data to Alerta:", err)
 		return
 	}
 }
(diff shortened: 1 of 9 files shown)


pipeline/alert.go
```diff
@@ -560,6 +565,84 @@ type HipChatHandler struct {
...
+	// Alerta event.
+	// This is a required field.
+        Event string
+
+	// Alerta environment.
+	// If empty uses the environment from the configuration.
+        Environment string
+
+	// Alerta status.
+        Status string
```
nathanielc: What is the use case for setting the status on the handler config? Seems like status is dynamic and doesn't make sense to configure here, but I am probably missing something.

nathanielc:
@md14454 Looks great. Just a few things

1. Can you run `go fmt` on the changes, there are several formatting errors.
2. Can you add another `alerta` handler in the test with different options? Showing that the defaults can be overridden and that multiple handlers can be configured. something like 
```
.alerta()
 			.token('testtoken1234567')
 			.resource('serverA')
                        .event('CPU Idle')
                        .origin('override')
```


Again this looks great, thanks for the PR!

pipeline/alert.go
md14454: Alerta.io has a number of different states, [see here](http://docs.alerta.io/en/latest/api/alert.html#status-table),  that it can parse and I didn't want to be prescriptive by automatically assigning the status. My intention was for the user to calculate a status string in the TICKscript and then use the variable as part of the Alerta status. Though I have realised that I don't run the string through the template engine. Are you able to advise on how to do this? (It may be useful for other fields also)

nathanielc: What status are you using internally?

Kapacitor is about generating alerts and doesn't (at least not yet) want to deal with managing alert workflows. This is why Kapacitor wants to integrate with many third party tools like Alerta that do that for you. So to me it seems logical to have the alert status be in the 'open' state when Kapacitor creates it and then in the 'closed' state once it resolves. The other states are about interacting with the alert which you cannot do with Kapacitor.

alert.go
```diff
@@ -594,6 +599,30 @@ func (a *AlertNode) handleHipChat(hipchat *pipeline.HipChatHandler, ad *AlertDat
...
+	if a.et.tm.AlertaService == nil {
+		a.logger.Println("E! failed to send Alerta message. Alerta is not enabled")
+		return
+	}
+	err := a.et.tm.AlertaService.Alert(
+		alerta.Token,
+		alerta.Resource,
+		alerta.Event,
+		alerta.Environment,
+		ad.Level.String(),
```
nathanielc: I looked at the Alerta Severities as well, since Alerta has many more you should use a switch statement to map the Kapacitor alert levels to the Alerta severity strings since they do not map up by default.

 See handleVictorOps for an example of what I mean.

md14454:
@nathanielc I've updated the alerta code to select severity and status based on the Kapacitor level, as per your request. The CI tests seem to fail on my local machine and they previously did on circleci both due to a timeout when pulling boltDB from github; not sure why. Running build.py --test seems to work locally however.

nathanielc:
@md14454 This looks good to go thanks. Can you `go fmt` the code and rebase? Then I'll merge. Thanks.

Don't worry about the tests, I have a few that are prone to error that I have been trying to chase down.

md14454:
@nathanielc ```go fmt ``` didn't change anything and ```git rebase master``` comes back saying that the branch is up to date. Hopefully that means we're good to go; or I've missed something.

nathanielc:
@md14454 Sorry I should be more specific.

You will need to run `go fmt ./...` in order for it to recursively format the sources.

And by rebase I meant rebase and squash all of your commits into a single commit. Here is an explanation of what I mean http://makandracards.com/makandra/527-squash-several-git-commits-into-a-single-commit Either method there works, I personally use *the hard way* because I find it easier and less prone to error.

md14454:
@nathanielc That should be done now. Apologies I haven't coded in golang or used a rebase workflow before.

nathanielc:
@md14454 no problem. Thanks again.

PR #149, opened 2016-01-19
Add deadman's switch

### LINKED ISSUE #137: Make creating an alert based on event count easy
Currently creating an alert on event count seems like a simple task, for example:

```
stream().from()...
   .window()
       .period(10s)
       .every(10s)
   .mapReduce(influxql.count('value'))
   .alert()
      .crit(lambda: "count" == 0)
```

**The above script will not work.** The reason is that once the event stream stops so does time and therefore the rest of the pipeline is not executed. As a result the alert logic is never evaluated and an alert is never triggered. 

Possible solutions:

* Use a separate task and the internal Kapacitor stats about throughput to achieve this. 
   For example:
 ```
stream()
   .from()
        .measurement('_kapacitor')
        .where(lambda: "task" == 'my_task' AND "parent" =='stream0')
   .alert()
      .crit(lambda: "collected" == 0)
```
 This will work but has the requirement that you use two tasks. One to receive the data and the other to check the throughput.

*  Add the ability to use the real clock time on a window node so that when real time passes it will emit a batch even if its empty. i.e.

```
stream().from()...
   .window()
       .period(10s)
       .every(10s)
       .realtime() //use realtime so even if the data stops the empty/partial window will still be emitted.
   .mapReduce(influxql.count('value'))
   .alert()
      .crit(lambda: "count" == 0)
```

Simple and optional. Using the realtime clock allows for races to occur when processing the data and so will generally not be recommended but for the specific use case of counting it could work. To be clear the races are not golang data races but rather races in how Kapacitor processes the data, i.e the datapoint arrives too late and so Kapacitor does not process the datapoint but has to drop it since the window was already emitted.

### PR DESCRIPTION
**Pull request by nathanielc**
Fixes #137 

A deadman's switch can now be added to any node within any task. This is accomplished via exposing the internal statistics per node to the TICKscript itself via the `stats` method.

The `stats` method emits the internal stats of a node at a given interval. The deaman's switch uses the `collected` stat to trigger an alert if it drops below a threshold.

The method `deadman` is available on all nodes and is a helper function to easily create a deadman's switch.

This:
```javascript
 var data = stream.from()...
    // Trigger critical alert if the throughput drops below 100 events per 10s and checked every 10s.
    data.deadman(100.0, 10s)
    //Do normal processing of data
    data....
```
 is equivalent to this:

```javascript
 var data = stream.from()...
    // Trigger critical alert if the throughput drops below 100 events per 1s and checked every 10s.
    data.stats(10s)
          .derivative('collected')
              .nonNegative()
              .unit(10s)
          .alert()
              .id('node \'stream0\' in task \'{{ .TaskName }}\'')
              .message('{{ .ID }} is dead: {{ index .Fields "collected" }} points/10s')
              .crit(lamdba: "collected" <= 100.0)
    //Do normal processing of data
    data....
```

In addition to the `deadman` method you can globally configure all stream tasks to have deadman's switches.

```
[deadman]
  global = true
  interval = "10s"
  threshold = 100.0
```

The `id` and `message` fields can also be configured globally.

EDIT: Changed examples to match final implementation

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,12 +11,12 @@ See [udf/agent/README.md](https://github.com/influxdata/kapacitor/blob/master/ud
 With the addition of UDFs it is now possible to run custom anomaly detection alogrithms suited to your needs.
 There are simple examples of how to use UDFs in [udf/agent/examples](https://github.com/influxdata/kapacitor/tree/master/udf/agent/examples/).
 
-
 The version has jumped significantly so that it is inline with other projects in the TICK stack.
 This way you can easily tell which versions of Telegraf, InfluxDB, Chronograf and Kapacitor work together.
 
 
 ### Features
+- [#137](https://github.com/influxdata/kapacitor/issues/137): Add deadman's switch. Can be setup via TICKscript and globally via configuration.
 - [#72](https://github.com/influxdata/kapacitor/issues/72): Add support for User Defined Functions (UDFs).
 - [#138](https://github.com/influxdata/kapacitor/issues/138): Change over to influxdata github org.
 - [#139](https://github.com/influxdata/kapacitor/issues/139): Alerta.io support thanks! @md14454
--- a/alert.go
+++ b/alert.go
@@ -392,6 +392,9 @@ type idInfo struct {
 	// Measurement name
 	Name string
 
+	// Task name
+	TaskName string
+
 	// Concatenation of all group-by tags of the form [key=value,]+.
 	// If not groupBy is performed equal to literal 'nil'
 	Group string
@@ -418,9 +421,10 @@ func (a *AlertNode) renderID(name string, group models.GroupID, tags models.Tags
 		g = "nil"
 	}
 	info := idInfo{
-		Name:  name,
-		Group: g,
-		Tags:  tags,
+		Name:     name,
+		TaskName: a.et.Task.Name,
+		Group:    g,
+		Tags:     tags,
 	}
 	var id bytes.Buffer
 	err := a.idTmpl.Execute(&id, info)
@@ -437,9 +441,10 @@ func (a *AlertNode) renderMessage(id, name string, group models.GroupID, tags mo
 	}
 	info := messageInfo{
 		idInfo: idInfo{
-			Name:  name,
-			Group: g,
-			Tags:  tags,
+			Name:     name,
+			TaskName: a.et.Task.Name,
+			Group:    g,
+			Tags:     tags,
 		},
 		ID:     id,
 		Fields: fields,
--- a/cmd/kapacitor/main.go
+++ b/cmd/kapacitor/main.go
@@ -616,6 +616,11 @@ func reloadUsage() {
 }
 
 func doReload(args []string) error {
+	if len(args) < 1 {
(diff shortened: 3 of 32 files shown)


nathanielc:
@gunnaraasen @rossmcdonald I know you both had some original input. How does the final result look?

rossmcdonald:
@nathanielc Just to make sure, if I wanted to alert if throughput drops below 1 event every 10s, would that be the following?

```
data.deadman(10s, 0.1)
```

I can see this notation being a little confusing, since it's not obvious that it samples every second. Without looking at any documentation, I would assume that this:

```
data.deadman(10s, 100.0)
```

Means "alert when you receive less than 100 events in 10 seconds".

nathanielc:
@rossmcdonald Correct. I was torn as well on which I liked.

With this one `data.deadman(10s, 0.1)` you can reason about it always in the units of seconds. So if you want to increase the frequency at which you check you only need change the interval. `data.deadman(1s, 0.1)`

For this one `data.deadman(10s, 100.0)` If you change the time you also have to remember to change the threshold or you just broke your deadman's switch.  `data.deadman(1s, 10.0)`

nathanielc:
To be clear, I can easily do it either way, I just decided that once you know its always events/second it becomes easier to use.

gunnaraasen:
I'll second @rossmcdonald's confusion on the arguments to the `deadman` function.

It would be nice if the rate interval in the `deadman` helper was configurable. This would make deadman checks over larger time periods, e.g. >15 min, easier and help add context for other people looking at the TICKscript after it's written. It's not immediately obvious that `.deadman(5m, 0.1111)` means check for a rate less than 10 events over the last 15 minutes every 5 minutes.

Also, what would be the recommendation for alerting on an absolute time since the last event? Say alerting only when a service hasn't sent any events in the last 10 minutes.

nathanielc:
@gunnaraasen  So do you want a third argument, like
```
.deaman(5m, 10.0, 15m)
```

or is 
```
.deadman(5m, 3.33) 
```
good enough?

nathanielc:
@gunnaraasen As for alerting on an absolute time since last event that is the same as a 0 threshold for that time period correct?

```
.deadman(10m, 0)
```

If you want both its doable(but not via the configuration)

```javascript
var data = stream.from()...

data.deadman(10m, 0.0)
data.deadman(1h, 5.0)// using Ross's notation read as less than 5 events for the hour

// Do normal data fprocessing
data....
```

As an aside I am leaning towards Ross's notation since otherwise its a pain to calculate.

rossmcdonald:
Regarding:

> With this one data.deadman(10s, 0.1) you can reason about it always in the units of seconds.

I like the idea, but I think in practice it will just end up becoming confusing since it's not explicitly obvious from the function name. It also requires a little bit of mental arithmetic that I think may be a bit too cumbersome.

gunnaraasen:
I like Ross' notation as well. Although switching the order of the arguments might be slightly more readable:

```
data.deadman(5.0, 1h) // Then reading left to right it would be 5 events / 1 hour
```

However, then it's unclear how often the rate checked? Does it default to checking once an hour?

Also, the 0 rate threshold works perfectly for checking the absolute time since the last event.

rossmcdonald:
I also think that switching the parameters would help improve readability.

PR #155, opened 2016-01-20
Change configuration package to influxdata/config

timraymond:
We are unifying the way that we handle configuration across the products into
the influxdata/config package. This provides the same API as the
BurntSushi/toml package used previously, but uses influxdata/toml under the
hood (which is a fork of naoina/toml). The underlying toml parser has been
changed because Telegraf uses specific features of this parser that cannot be
easily replicated with the BurntSushi parser.  Our fork of naoina/toml provides
support for maps and pointers within structs and toml documentation[1]
facilities, which the original does not, at present time (pull requests have
been submitted with no response with no response from the maintainer).

[1] This is accessible by adding a “doc” struct tag with a comment describing a
particular field of a config struct. When marshalling that struct as TOML, the
“doc” struct tag will be placed appropriately to
document that field in the resultant TOML output.


nathanielc:
@timraymond I am testing locally and this looks good. 

I will need to wait for InfluxDB core to merge their config first since mine config imports some of their config.

nathanielc:
Link to InfluxDB PR https://github.com/influxdata/influxdb/pull/5401

nathanielc:
@timraymond As per the issues found with the config package this PR is not viable in the near future. I am going to close the PR. We can open a new one once we have fixed influxdb/config

Issue #85, opened 2015-12-09
Support Sensu


sstarcher:
We currently use sensu for all of our alerting needs and routing.

JulienChampseix:
+1

nathanielc:
Could either of you provide and example TICKscript of what sending alert data to Sensu would look like? Don't worry about implementation just how you would expect it to look in the script. Thanks

sstarcher:
I'm not 100% sure what you are asking for so if I don't answer it correctly let me know.

Minimal data would look like the following.
```
{
   "name":"{{ .Env.NAME }}",
   "source":"kapacitor",
   "output":"{{ .Env.OUTPUT }}",
   "status": {{ .Env.STATUS }},
}
```

* Output can be any string
* Name would be a unique name for the alert - `/^[\w\.-]+$/`
* Status integer - 0=OK, 1=warn, 2=critical, 3+ unknown

https://sensuapp.org/docs/0.21/clients

Other configuration information required
* HOST
* PORT

TICKscript?
```
    stream...
         .alert()
             .sensu()
```

nathanielc:
@sstarcher That is what I was looking for thanks

Do you expect Kapacitor to act as a sensu client (heartbeats etc) or just send events over to sensu via the API and let it do the rest?

sstarcher:
I would say the easiest implementation would be to not send heartbeats.  My main goal would be to have all of my alerting outbound in one place.  Kapacitor would detect the problem and instead of configuring Kapacitor to send to slack/pagerduty etc.  We would instead using our existing sensu setup that all of our teams have configured routing and alerting rules already for.

jcmartins:
But with telegraf in all clients sending data to Influxdb and make alerts/checks in Kapacitor, you don't need use sensu anymore. ;-)

sstarcher:
@jcmartins I'll assume sarcasm ;)

As sensu has the capabilities to alert on much more than metrics.

jcmartins:
@sstarcher In kapacitor you can have alert like sensu(warning, critical, ok)  but I agree with you kapacitor and telegraf don't have mature and same features yet.
I'm using sensu+influxdb+grafana a lot with victorops on 400 ec2 aws instances and 50 RDS, but I'm doing some tests here and I'm surprised at kapacitor + Telegraf  can do.
I'll need to create/customize some plugins for Telegraf as I have done for Sensu.
Maybe I'll use it on my dev/uat environments in the version 0.4 February 2016?!

[...]

PR #160, opened 2016-01-23
Sensu integration for Kapacitor

sstarcher:
In relation to ticket #85

### DIFF
--- a/services/sensu/service.go
+++ b/services/sensu/service.go
@@ -1,0 +1,77 @@
+package sensu
+
+import (
+	"bytes"
+	"encoding/json"
+	"errors"
+	"github.com/influxdata/kapacitor"
+	"log"
+	"net/http"
+)
+
+type Service struct {
+	url    string
+	source string
+	logger *log.Logger
+}
+
+func NewService(c Config, l *log.Logger) *Service {
+	return &Service{
+		url:    c.URL,
+		source: c.Source,
+		logger: l,
+	}
+}
+
+func (s *Service) Open() error {
+	return nil
+}
+
+func (s *Service) Close() error {
+	return nil
+}
+
+func (s *Service) Alert(name, output string, level kapacitor.AlertLevel) error {
+	var status int
+	switch level {
+	case kapacitor.OKAlert:
+		status = 0
+	case kapacitor.InfoAlert:
+		status = 0
+	case kapacitor.WarnAlert:
+		status = 1
+	case kapacitor.CritAlert:
+		status = 2
+	default:
+		status = 3
+	}
+
+	postData := make(map[string]interface{})
+	postData["name"] = name
+	postData["source"] = s.source
+	postData["output"] = output
+	postData["status"] = status
+
+	var post bytes.Buffer
+	enc := json.NewEncoder(&post)
+	err := enc.Encode(postData)
+	if err != nil {
+		return err
+	}
+
+	resp, err := http.Post(s.url, "application/json", &post)
+	if err != nil {
+		return err
+	}
+	defer resp.Body.Close()
+	if resp.StatusCode != http.StatusOK {
+		type response struct {
+			Error string `json:"error"`
+		}
+		r := &response{Error: "failed to understand Sensu response"}
+		dec := json.NewDecoder(resp.Body)
+		dec.Decode(r)
+		return errors.New(r.Error)
+	}
+	return nil
+}
--- a/alert.go
+++ b/alert.go
@@ -144,6 +144,11 @@ func newAlertNode(et *ExecutingTask, n *pipeline.AlertNode, l *log.Logger) (an *
 		an.handlers = append(an.handlers, func(ad *AlertData) { an.handlePagerDuty(&pipeline.PagerDutyHandler{}, ad) })
 	}
 
+	for _, sensu := range n.SensuHandlers {
+		sensu := sensu
+		an.handlers = append(an.handlers, func(ad *AlertData) { an.handleSensu(sensu, ad) })
+	}
+
 	for _, slack := range n.SlackHandlers {
 		slack := slack
 		an.handlers = append(an.handlers, func(ad *AlertData) { an.handleSlack(slack, ad) })
@@ -568,6 +573,23 @@ func (a *AlertNode) handlePagerDuty(pd *pipeline.PagerDutyHandler, ad *AlertData
 	)
 	if err != nil {
 		a.logger.Println("E! failed to send alert data to PagerDuty:", err)
+		return
+	}
+}
+
+func (a *AlertNode) handleSensu(sensu *pipeline.SensuHandler, ad *AlertData) {
+	if a.et.tm.SensuService == nil {
+		a.logger.Println("E! failed to send Sensu message. Sensu is not enabled")
+		return
+	}
+
+	err := a.et.tm.SensuService.Alert(
+		ad.ID,
+		ad.Message,
+		ad.Level,
+	)
+	if err != nil {
+		a.logger.Println("E! failed to send alert data to Sensu:", err)
 		return
 	}
 }
--- a/cmd/kapacitord/run/config.go
+++ b/cmd/kapacitord/run/config.go
@@ -20,6 +20,7 @@ import (
 	"github.com/influxdata/kapacitor/services/pagerduty"
 	"github.com/influxdata/kapacitor/services/replay"
 	"github.com/influxdata/kapacitor/services/reporting"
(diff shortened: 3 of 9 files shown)


sstarcher:
Correct me if I'm wrong, but that failing CI build is unrelated to my changes and is related to upstream changes in github.com/influxdb/influxdb/meta

anlutro:
I don't think this is a good idea. This relies on the sensu client listening on a public IP as well as the sensu client port being open in whatever firewall/network settings are applied in the infrastructure - or running Kapacitor on each individual host. That's certainly not an option for me, and I would recommend anyone against doing any of those things.

sstarcher:
@anlutro This is actually a recommended setup for external sensu checks. https://sensuapp.org/docs/latest/clients#jit-clients

The normal setup is that you have a single sensu client or a few sensu clients that accept HTTP requests and forward those to the server.  

The Sensu client does not have to listen on a public IP.  You can certainly set sensu client up on another server and point Kapacitor to that server or you can run a sensu client on the same server as Kapacitor your choice.  I don't see how this is any less secure than communication to RabbitMQ.  Sensu specifically designed JIT checks a few versions back for this exact use case.

anlutro:
Aha, I guess I didn't read your PR thoroughly enough. Sorry.

I see you're setting `source = Kapacitor` for the example. My understanding is that `source` should be set to the hostname of the host that's triggering the alert. That way, the Kapacitor alert is grouped together with the other alerts related to that host.

Also, my understanding was that JIT checks are just that - checks, not events. Does that make any difference in this implementation?

Correct me if I'm wrong.

sstarcher:
@anlutro The source certainly could be an individual host or it could be an abstract external element.  If you ran Kapacitor on each of your servers than you would likely want the source to be the ID of that server similar to how you would set the sensu client name.

Or in the other situation you could run a single Kapacitor instance that is aggregating the metrics for all of your servers and alerting off of that data.

The third situation would be where you have a single Kapacitor instance and you want to segregate the sources under different individual sensu clients.  The code currently supports the first 2 cases, but not the third use case.

Sensu JIT checks are dynamic checks.  Essentially when you sent a event sensu will generate the relevant check data if it does not already exist.  So on sending the first HTTP Post sensu will generate the sensu check data for that HTTP Post on the fly.

anlutro:
> If you ran Kapacitor on each of your servers than you would likely want the source to be the ID of that server similar to how you would set the sensu client name.

I can't imagine anyone wanting to do this. It would mean either running InfluxDB on each host, or a tremendous amount of network activity as Kapacitor on each instance is streaming data from InfluxDB.

The third use case you're describing is really the only one that would be acceptable for our use case. Maybe it's as simple as adding a `source_from_tag` config key, which I would set to `host`, and then the `host` tag in the dataset Kapacitor is alerting on would be the source for the check?

nathanielc:
@sstarcher Thanks for the work and the explanation of how it can be used. Yes, the CI failure is unreleated. I'll get that fixed soon.

@anlutro As for running Kapacitor on each host. While this is a bit off topic I wanted to point out that you don't have to use InlfuxDB to send data to Kapacitor. A simple use case would be to have each host running Telegraf and Kapacitor. The Telegraf instance would send its data to both InfluxDB and to the local Kapacitor. With that setup your alert processing is distributed such that each node is checking its own health. Then you could still run a centralized Kapacitor but this time all you would care about are the alerts that span multiple hosts since all single node alerts have already be taken care off. While a little more work to setup and maintain is does provide a nice fault tolerant design. See #163 for details on the Kapacitor -> Telegraf connection, there are a few gotchas.

Overall this looks good. I'll probably have some more feedback soon.

services/sensu/service.go
```diff
@@ -0,0 +1,84 @@
...
+	return nil
+}
+
+func (s *Service) Close() error {
+	return nil
+}
+
+func (s *Service) Alert(name, output string, level kapacitor.AlertLevel) error {
+	var Url *url.URL
+	Url, err := url.Parse(s.url)
```
nathanielc: If the URL never changes, why don't we parse it once in the NewService call?

sstarcher:
@nathanielc sounds good I will likely add more functionality to the sensu integration like adding the ability to supply adhoc tags, but this should cover the most basic usecase.

nathanielc:
@sstarcher The CI build issue has been fixed. If you rebase the tests should pass.

anlutro:
@sstarcher I wasn't aware of the Telegraf → Kapacitor possibility, that could be an option I guess - though it does mean I lose access to influxql stuff, right? So I can't do calculations using mean, standard deviation and sigma?

nathanielc:
@anlutro With the Telegraf → Kapacitor you can send data from Telegraf to multiple outputs so you can still send it to InfluxDB. You could also have Kapacitor store it in InfluxDB if you like as well. You will not lose any functionality in either case.

@sstarcher This looks great thanks for putting it together.

nathanielc:
@sstarcher I just realized you haven't signed the CLA. Can you do so? http://influxdb.com/community/cla.html Thanks

sstarcher:
@nathanielc done

nathanielc:
Thanks

brettdh:
@sstarcher: this is great work, but one thing tripped me up: the default config uses port 3030, which is the default for the Sensu client external check input, not for the server API. However, the external check input socket just expects JSON, not an HTTP POST. Assuming you had this working against a real Sensu server, I imagine I just need to pick the right Sensu URL - but I haven't found it yet. It seems like it *should* go to the external check results API, but as I mentioned, that's not expecting HTTP POST.

Could you update the default config (and comments) to reflect what is going on there? I can't currently sign the CLA to submit this change myself (and I don't yet know the right answer anyway).

sstarcher:
@brettdh I don't believe you can post checks to the sensu server.  The sensu server just reads from the redis queue.  The current setup was to not pull in the reliance on the redis queue and it's implementation.  Which is why it's setup for port 3030 and HTTP Post to a sensu client.

One option would be to run both the sensu client and the sensu server on your server, but the server itself does not accept POST requests.

brettdh:
@sstarcher Right, but when I pointed it at my Sensu client, the Sensu client drops the incoming data. Looking at the log, it reports that the received data was invalid JSON, and it points to the HTTP header as the invalid data. Hence my supposition that it shouldn't be a HTTP POST to sensu-client:3030 (and uncertainty whether the URL is wrong or having it be HTTP is wrong). In addition, the Sensu docs regarding external check results show examples using raw JSON through a TCP socket, with no HTTP request.

nathanielc:
@brettdh This doc clearly states that the API expects HTTP POST request for JIT clients  https://sensuapp.org/docs/latest/api-clients.

sstarcher:
@brettdh after taking a better look it looks like 2 methods are supported
1. HTTP POST to Sensu API HOST:4567/clients
2. TCP data to Sensu Client port 3030

The current code sends an HTTP POST, but the default example shows the port of the sensu client.
It should function if you point it to
```
[sensu]
  enabled = true
  url = "http://sensu-api:4567/clients"
  source = "Kapacitor"  
```

So the current code if probably configured should be able to post JIT checks to the Sensu API via HTTP POST, but not to Sensu Clients via TCP.

brettdh:
@nathanielc I was referring to this one, which states that the Sensu client listening on port 3030 expects raw JSON: https://sensuapp.org/docs/latest/clients#example-external-check-result-input

@sstarcher Great, thanks for looking into that. I originally thought that the POST should be going to the Sensu API as well. I tried this configuration at the server earlier, but got a `400 Bad Request` error. I'm unsure whether I tried the `/clients` endpoint, but trying it now, this is what I still get. Sadly, the response body doesn't contain any useful error info.

I remain extra confused because the data that the Sensu output is sending in this PR appears to just be what the TCP input expects (name, output, source, status), wrapped in a POST, rather than what the `/clients` endpoint expects (name, address, subscriptions, environment).

brettdh:
I managed to get this working by:

  1. Using a simple http server to receive the POST and forward the JSON to the client; and
  1. Ensuring that the `.id()` defined in my tickscript doesn't have spaces or special chars.

I think this feature is great, but the implementation, documentation, and testing need some work, as evidenced by my experience. (Should I open a new issue for these things?) I'm currently having issues with getting my employer's approval to sign the CLA; otherwise I would love to help with this. Hopefully I can figure that out in the near-ish future.

nathanielc:
@brettdh Yes, please open an issue. Thanks for taking time to work through some of the issues. Based on what I am reading it shouldn't be too difficult to fix the issues.

Issue #169, opened 2016-01-25
Now with the addition of time functions in expressions we should support time math


nathanielc:
For example this should be a valid expression

`lambda: "time" - now() > 5s`

washingtoneg:
Hello. Is there any plan to prioritize this feature on the roadmap? Is there a workaround that can be utilized in the interim?

Issue #168, opened 2016-01-25
Add cron match time function


nathanielc:
In addition to the normal time functions (see #141) we should be able to do `lambda: cronMatch('* 8-17 * * * *', "time")` so greatly simplify and time based expressions one would write.

PR #170, opened 2016-01-25
Add time of day expressions

### LINKED ISSUE #141: Allow alert evaluation to only occur during certain times/days
It would be really useful to have a way to prevent alerts/expressions from being evaluated during weekends or scheduled maintenance windows. Non-24/7 services typically will go down when not being used so we'd like to not have to receive emails about it.

nathanielc:
Fixes #141

Now there are 6 functions for working with time in the lambda expressions:

* minute -- the minute within the hour [0,59]
* hour -- the hour within the day [0,59]
* weekday -- the weekday within the week [0,6] 0 is Sunday
* day -- the day within the month [1,31]
* month -- the month within the year [1,12]
* year -- the year.

Example using the deadman's switch:

```javascript
//Trigger deadman's switch only between 8am-5pm local time.
stream.deadman(0,10s,lambda: hour("time") >= 8 AND hour("time") <= 17)
```

Note this is based off the time of the data not the current system time.

Implementing a `cron` function was a little more work so I have created a new issue to track adding a cron time function.  #168

Adding these functions makes me think we should add the ability to perform time math, i.e `lamda: "time" - now() > 5s` I have created an issue add this feature as well. #169

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ See note on a breaking change in the HTTP API below. #163
 - [#72](https://github.com/influxdata/kapacitor/issues/72): Add support for User Defined Functions (UDFs).
 - [#139](https://github.com/influxdata/kapacitor/issues/139): Alerta.io support thanks! @md14454
 - [#85](https://github.com/influxdata/kapacitor/issues/85): Sensu support using JIT clients. Thanks @sstarcher!
+- [#141](https://github.com/influxdata/kapacitor/issues/141): Time of day expressions for silencing alerts.
 
 ### Bugfixes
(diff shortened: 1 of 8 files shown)


nathanielc:
@gargooza @lpalm How does this implementation look to you?

Here is another example without using the deadman's switch.

```javascript
stream
    .from()....
    .alert()
        .crit(lambda: "value" > 100 
                           AND hour("time") >= 8 
                           AND hour("time") <= 17 
                           AND weekday("time") >= 1 
                           AND weekday("time") <= 5)
```

Obviously the `cron` syntax would be much cleaner. It is coming.

lpalm:
This looks great, Nathaniel. Thanks. We'll test it ASAP. The cron syntax would make for nice syntactic sugar but this is definitely usable (and quite clear) as is.

PR #183, opened 2016-01-29
Specify RestartSec for systemd

anlutro:
Kapacitor may depend on an external resource (most likely influxdb) to be present when starting. If that service is unavailable, kapacitor will exit immediately. The default restart time for systemd services is
100ms. If a service restarts too often within a timespan, systemd will stop trying to restart it. Adding a higher RestartSec prevents this problem.

### DIFF
--- a/scripts/kapacitor.service
+++ b/scripts/kapacitor.service
@@ -13,6 +13,7 @@ EnvironmentFile=-/etc/default/kapacitor
 ExecStart=/usr/bin/kapacitord -config /etc/kapacitor/kapacitor.conf $KAPACITOR_OPTS
 KillMode=process
 Restart=on-failure
+RestartSec=5s
 
 [Install]
 WantedBy=multi-user.target


nathanielc:
@anlutro Help me understand, if Kapacitor has a dependency on InfluxDB and InfluxDB is not running systemd will restart Kapacitor but since it fails so quickly systemd will eventually stop restarting Kapacitor and leave it not running? Isn't that the expected behavior if InfluxDB is not running? If its a race condition between InfluxDB and Kapacitor startup than I suggest adding an explicit dependency instead of a longer timeout period.

anlutro:
You are correct, but influxdb may be running on another host, which means
defining it as a dependency in the kapacitor service isn't really an option.

nathanielc:
How do you synchronize starting InfluxDB and Kapacitor on different hosts to be within 5 sec? 
What if InfluxDB is recovering from a crash and is therefore replaying the WAL and takes more time? From my perspective increasing a timeout seems like the wrong solution, but maybe I am not seeing it yet.

anlutro:
In my eyes, the alternative would be to not restart kapacitor at all in
that case, but waiting some time between restarts seems like the most
practical solution to me.

nathanielc:
@anlutro v0.10.1 now has a retry mechanism for connecting to InfluxDB. Has that eliminated the need to add a systemd restart?

anlutro:
It probably would, for this use case. Do you know if in the case of something like a configuration file syntax error, would Kapacitor still try and restart itself?

nathanielc:
Kapacitor will not try and restart itself in any condition. It now will retry connecting to InfluxDB instead of exiting. In the case of a configuration file syntax error a restart is not much help. 

At this point I am closing this PR as the need for restart is not strong and global to all use cases, if an individual setup requires it, it is easy to add post install.

PR #191, opened 2016-02-01
Build Script Updates

rossmcdonald:
Porting changes from the InfluxDB and Telegraf build scripts to `build.py`. There are quite a few changes here, namely:

- Structure of script was changed to make it easier to port future changes. All project-specific variables and functions should go at the top of the file, while the bottom portion of the file should not change (unless it changes across all projects).
- `tar` package output has a `kapactor-<version>` base directory (instead of the root package filesystem)
- Modified the package output for 32-bit builds to use `i386` instead of `386` designation for easier integration with package managers (doesn't apply to Kapacitor at the moment)
- Moved `boto` import to upload function. Also enabled folder hierarchy when uploading packages to S3 ( can upload to `kapacitor/folder1`, etc).
- A few more CLI options were added (`--no-get`, `--debug`, `--goarm=`, `--pkgarch`)
- Added `--no-vet` to `circle.yml` since it was causing builds to break, and I believe it's unneeded for Kapacitor.


rossmcdonald:
@nathanielc I believe this is ready for your review. I've tried to push all project-specific code to the top of the file to make it easier to port features across the different projects in the future, but let me know if you think there may be any issues with that approach.

build.py
```diff
@@ -30,10 +31,13 @@
 DEFAULT_CONFIG = "etc/kapacitor/kapacitor.conf"
 PREINST_SCRIPT = None
 
+# Default AWS S3 bucket for uploads
+DEFAULT_BUCKET = "kapacitor"
```
nathanielc: I have been using the influxdb bucket. To my knowledge there is no kapacitor bucket.

rossmcdonald: I wasn't aware of that. Regardless, I've created a kapacitor bucket that we can use for releases or nightlies (or you can continue to use influxdb). I've been trying to spread out S3 uploads into separate buckets, since the influxdb bucket is becoming a bit unwieldy.

nathanielc: I like the idea of using a kapacitor specific bucket. I'll just make sure the next release we update the bucket path. Thanks

circle.yml
```diff
@@ -21,8 +21,8 @@ dependencies:
 
 test:
   override:
-    - ./build.sh --test --generate
-    - ./build.sh --test --generate --race
+    - ./build.sh --test --generate --no-vet
+    - ./build.sh --test --generate --race --no-vet
```
nathanielc: I do want to vet the code. If its not passing then we need to fix it.

nathanielc:
@rossmcdonald  I am trying this branch out locally. I am making a few small changes. I'll post them soon.

circle.yml
rossmcdonald: Ah, okay. I wasn't sure if there was a difference between the way InfluxDB and Kapacitor use that option. I remember seeing quite a few errors when I was testing it out, though.

nathanielc:
@rossmcdonald I have pushed up my changes.

I added a `--no-stash` option. Now `go_get` will first check for local changes and if `no-stash` was provided it will error out the script. This has a few nice benefits. 

* No longer possible to accidentally build a package off non committed code. This is important since we want to always know what code is in a published package.
* The build will fail if the generated code is out of date. Since generate is run first and if it changes anything then changes will stop the build. Again back to the first point we want to know exactly what code makes it into a published package.

Sorry for all the whitespace changes hopefully this change is beneficial to all projects and not too difficult to port later.

 I am still working on the go vet issue....

nathanielc:
@rossmcdonald For the go vet I am using different options to the command. So I turned the command into a project specific option so that each project can control what go vet does. Sound good?

rossmcdonald:
@nathanielc That sounds great. I was wondering why I was receiving so many errors when from the `vet` output, so that makes sense.

PR #204, opened 2016-02-03
add alert support for Details property

### LINKED ISSUE #75: Alert - Custom field
That will be good if inside the mail sent over alerting system we could include an content like an URL which could contain procedure.

By this way we could a mail template having:
- metrics impacted by the alert
- current threshold exceed
- URL procedure linked to this alert 

Just an idea when it's an support team receiving alert. By this way they will know which metrics impacted and whichaction required following the URL procedire included into the alert for resolve the problem ;-)

Thanks for your feedback.

Julien.

### LINKED ISSUE #196: Way of formatting email body?
Hi,

is there a way of formatting the alert's `email()` body? With a combination of `id()` and `message()` one can modify the email subject, can it be done for the body as well to get something other than the JSON body which is currently being sent?

Thank you

nathanielc:
Fixes #196 and Fixes #75 

Now the alert node has a `details` property that can be formatted HTML containing details of the alert. The details property is used as the body of email alerts.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ Improved UDFs, lots of bug fixes and improvements on the API. There was a breaki
 
 - [#176](https://github.com/influxdata/kapacitor/issues/176): Improved UDFs and groups. Now it is easy to deal with groups from the UDF process.
     NOTE: There is a breaking change in the BeginBatch protobuf message for this change.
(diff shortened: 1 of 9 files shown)


wonkaba:
Does this fix translate over to the official Dockerhub Kapacitor container?  I am using the details property but emails are still sending JSON

nathanielc:
@wonkaba Yes, this fix made it in before any official docker containers were distributed. Can you either post a question on the mailing list or submit a new issue if you have found a bug? Thanks

PR #211, opened 2016-02-04
Add test.sh for testing various envs

nathanielc:
This takes the work done for InfluxDB. Now Kapacitor can be tested against various envs.

There is an experimental arm env which uses qemu. If its too slow we will have to remove it but for now I would like to see if its possible.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,4 @@ kapacitor*.tar
 kapacitor*.zip
 *.pyc
 *.test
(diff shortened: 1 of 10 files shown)


nathanielc:
OK so golang and qemu on arm don't play nice :( https://groups.google.com/forum/#!topic/golang-nuts/MqKTX_XIOKE

I will remove the arm test for now.

PR #216, opened 2016-02-05
Minor build updates

rossmcdonald:
Fixes:

- Moved nightly versioning to be set after CLI parsing due to an error that would occur if the `--nightly` was specified before `--version`
- Moved `go get` to be run for every platform, as some projects have platform-specific includes (and I didn't know `go get` was affected by the `GOOS` env variable)

### DIFF
--- a/build.py
+++ b/build.py
@@ -685,9 +685,6 @@ def main():
         elif '--nightly' in arg:
             # Signifies that this is a nightly build.
             nightly = True
-            # In order to cleanly delineate nightly version, we are adding the epoch timestamp
-            # to the version so that version numbers are always greater than the previous nightly.
-            version = "{}.n{}".format(version, int(time.time()))
         elif '--update' in arg:
             # Signifies that dependencies should be updated.
             update = True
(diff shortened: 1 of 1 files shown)


nathanielc:
Is go get affected by GOOS? I know go get will compile the code it downloads but we are passing the `-d` flag so its only downloading the code not compiling it. Seems odd since go get doesn't download individual files but rather whole repos. How did you run into the issue?

rossmcdonald:
@nathanielc I ran into this issue when trying to cross-compile Telegraf for Windows. `go get` only retrieves the dependencies for the platform that it's building on (I'm guessing that [this](https://github.com/influxdata/telegraf/blob/0d391b66a3253d3287c9710dcdf5d49e42cbb0c6/plugins/inputs/win_perf_counters/win_perf_counters.go#L1) is the flag). The only documentation I can find related to this is at the _very_ bottom of the "Build Constraints" section [here](https://golang.org/pkg/go/build/). I don't know, though, we can leave it out if you think it will cause more harm than good.

And should the `go get` command be modified to leave out the `-d`? I pulled that from the original packaging script, so if it doesn't make sense we can change it.

nathanielc:
@rossmcdonald  We want the `-d` flag. By default go get compiles the downloaded packages. But since we are just about to explicitly compile the code we want go get to just download the code and nothing else.

As for GOOS effecting what is downloaded it seems really odd and I am not convinced yet that it does ;) Running tests locally to try it out...

nathanielc:
I am now convinced :)

I ran `go get -d -v g ./...` in a new GOPATH for the Telegraf repo and the package `github.com/lxn/win` was not downloaded.

I then ran `GOOS=windows go get -d -v g ./...` and then the package was downloaded. So like you said  `go get` will only inspect files for the current GOOS var and if dependent package is only listed in files for a different GOOS it is not downloaded. 

Merging...

rossmcdonald:
@nathanielc Haha, thanks. I love it when stuff like that is undocumented...

gunnaraasen:
Might be worth [filing an issue](https://github.com/golang/go/issues/new) that it's undocumented. :)

PR #248, opened 2016-02-19
Add performance metrics to tasks and node

### LINKED ISSUE #215: Create system so that users can measure the performance of their task
Beyond benchmarks of static tasks we need to make it easy for an end user to measure directly the performance of their own tasks so they can provision resources appropriately

nathanielc:
Fixes #215 

This PR exposes two metrics to the user about throughput and execution time:

1. The throughput in batches or points per second per taks is calculated depending on task type
2. The average execution time per node is calculated. 

These values are exposed via the Kapacitor `show` command and in the internal statistics Kapacitor emits. As a result it is possible to graph throughput and execution times or other stats a node exposes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,12 +4,18 @@
 
 ### Release Notes
 
+Kapacitor now exposes more internal metrics for determining the performance of a given task. 
+The internal stats now includes a measurement `node` that contains an averaged execution time for the node, tagged by the task, node, task type and kind of node (i.e. window vs union).
+These stats are also available in the DOT output of the Kapacitor show command.
+
+
 ### Features
 - [#236](https://github.com/influxdata/kapacitor/issues/236): Implement batched group by
 - [#231](https://github.com/influxdata/kapacitor/pull/231): Add ShiftNode so values can be shifted in time for joining/comparisons.
 - [#190](https://github.com/influxdata/kapacitor/issues/190): BREAKING: Deadman's switch now triggers off emitted counts and is grouped by to original grouping of the data.
     The breaking change is that the 'collected' stat is no longer output for `.stats` and has been replaced by `emitted`.
 - [#145](https://github.com/influxdata/kapacitor/issues/145): The InfluxDB Out Node now writes data to InfluxDB in buffers.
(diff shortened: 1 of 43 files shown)


yosiat:
Hi @nathanielc ,
I am just wondering why put the metrics values and key concatenated in the label attribute and don't create new attribute for each metric?
For example:

Instead of:
```
digraph testTaskName {
graph [label="Throughput: 0.00 points/s"];
stream0 [label="stream0 0"];
stream0 -> stream1 [label="0"];
stream1 [label="stream1 0"];
}
```

Do something like this:
```
digraph testTaskName {
graph [throughput="0.00"];
stream0 [processed="0"];
stream0 -> stream1 [processed="0"];
stream1 [processed="0"];
}
```

- The throughput will be a well known format - "points per second" 
- I have omitted the "stream0" / "stream1" since I don't know what they mean.

This way the DOT graph will be much easier to parse and show in a UI for example and it's valid syntax ([DOT Attributes](https://en.wikipedia.org/wiki/DOT_(graph_description_language)#Attributes))

nathanielc:
Both locally and via g.gravizo.com those attributes are ignored. They need to be `label` attributes. I am open to suggestions to make it better but it needs to remain valid/useful DOT syntax. 

Alt text

vs

Alt text

yosiat:
You want it to be valid for view and for parsing, so let's merge our solutions: both label and attributes.

nathanielc:
@yosiat I think you are right, but the labels will clutter it up too much. I think by default I will do it your way as that is more human readable which is the common use case. And then a flag can be specified to use labels instead so it renders correctly. Thoughts? 

I don't like the complexity of something that should be so simple :(

Maybe there is another graph language that is easier to read and still renders well? I'll take a look around.

nathanielc:
Looking through all of these https://gephi.org/users/supported-graph-formats/ I still like DOT the best as it reads well. For now I guess we can add the flag and revisit later if it becomes cumbersome to set the flag when its needed.

yosiat:
@nathanielc  I agree with this option of flags.

About graph formats - if you don't need the "renderabillity" you can use your own json format.

PR #250, opened 2016-02-19
Add buffering to InfluxDBOutNode

### LINKED ISSUE #145: Add buffering and fault tolerance to the InfluxDBOut Node.
The current implementation just writes points as it receives them to the InfluxDB. In high volume scenarios this will not be sufficient. The node should buffer the data points before sending them on to InfluxDB. This is appropriate since if you wanted any near real-time analytics of the data that is being buffered you already have them in Kapacitor and can do so. The small delay in being written to InfluxDB will be inconsequential. 

By fault tolerance the data points writes should be retried etc before erroring out.

nathanielc:
Fixes #145


nathanielc:
Merging changes in with #215

PR #261, opened 2016-02-26
Updates to build.py

rossmcdonald:
A few updates ported from the InfluxDB build script:
- Improved handling of ARM architectures, where `armhf`, `armel`, and `arm64` are broken out.
- Removed the `--goarm` and `--pkgarch` build flags.
- Removed buildTime from linker flag specification so that reproducible builds can be created.
- `go get` is not project-specific, since InfluxDB and Telegraf use `gdm` and Kapacitor uses Go vendoring
- Removed the per-platform `go get` call since dependencies are vendored (or pinned, as is the case with InfluxDB)

### DIFF
--- a/build.py
+++ b/build.py
@@ -797,7 +797,7 @@ def main():
             return 1
 
     if run_get:
-        go_get(branch, update=update)
+        go_get(branch, update=update, no_stash=no_stash)
 
     if test:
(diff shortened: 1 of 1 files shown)


nathanielc:
@rossmcdonald looks good, +1 on green

PR #272, opened 2016-03-02
Sensu fixes

nathanielc:
@brettdh I believe I have addressed all the issues. ATM it is not easy to use docker to create a sensu client for full integration testing. The unit tests have been updated to use a raw TCP socket and the code has been tested against a real Sensu client and is behaving correctly. 

Note in addition to the sensu changes a small change was added to the alert node to count triggered alerts.


brettdh:
Nice, thanks. I look forward to pulling this in.

PR #291, opened 2016-03-08
PagerDuty Enhancement

### LINKED ISSUE #288: PagerDuty Service Enhancement
Based on the [PagerDuty event documentation](https://developer.pagerduty.com/documentation/integration/events), I believe Kapacitor should be sending `trigger` events when an issue is opened, and `resolve` when an issue is marked resolved. From looking at [the current code](https://github.com/influxdata/kapacitor/blob/a5cf89b6ff2de970fa4a2fabc5faa23e0a08f224/services/pagerduty/service.go#L15), it appears only `trigger` alerts are sent. This means that a new `trigger` event will be opened for every state change in the alert. Also, it doesn't look like the `incident_key` value is stored or sent when the issue is opened or resolved.

Since PagerDuty events are tertiary (they are only ever `trigger`, `acknowledged`, or `resolved`), I that think that it makes sense to take a binary alerting approach when alerting through PagerDuty. Namely, I think that when an alert is set to either `WARN` or `CRITICAL` that a `trigger` alert is fired, and, when an alert is set to `OK`, a `resolve` event should fire (using the same `incident_key`). Multiple triggers to the same `incident_key` will only append themselves to the trigger log, so there should be no issue with executing multiple `WARN` or `CRITICAL` events.

@nathanielc What do you think? I'd be willing to take this on, as long as you don't have any objections.

rossmcdonald:
Improvements to the PagerDuty service integration:
- Adds the concept of alert levels that correspond to the PagerDuty `trigger` and `resolved` events
- Includes the alert ID in the PagerDuty requests, instead of opening a new incident with every alert

Fixes #288

### DIFF
--- a/alert.go
+++ b/alert.go
@@ -646,6 +646,7 @@ func (a *AlertNode) handlePagerDuty(pd *pipeline.PagerDutyHandler, ad *AlertData
 	err := a.et.tm.PagerDutyService.Alert(
 		ad.ID,
 		ad.Message,
+		ad.Level,
 		ad.Data,
 	)
 	if err != nil {
--- a/services/pagerduty/service.go
+++ b/services/pagerduty/service.go
@@ -12,8 +12,6 @@ import (
 	"github.com/influxdata/kapacitor"
 )
 
-const eventType = "trigger"
-
 type Service struct {
 	HTTPDService interface {
 		URL() string
@@ -45,11 +43,22 @@ func (s *Service) Global() bool {
 	return s.global
 }
 
-func (s *Service) Alert(incidentKey, desc string, details interface{}) error {
+func (s *Service) Alert(incidentKey, desc string, level kapacitor.AlertLevel, details interface{}) error {
+	var eventType string
+	switch level {
+	case kapacitor.WarnAlert, kapacitor.CritAlert:
+		eventType = "trigger"
+	case kapacitor.InfoAlert:
+		return fmt.Errorf("AlertLevel 'info' is currently ignored by the PagerDuty service")
+	default:
+		eventType = "resolve"
+	}
+
 	pData := make(map[string]string)
 	pData["service_key"] = s.serviceKey
 	pData["event_type"] = eventType
 	pData["description"] = desc
+	pData["incident_key"] = incidentKey
 	pData["client"] = kapacitor.Product
 	pData["client_url"] = s.HTTPDService.URL()
 	if details != nil {
--- a/task_master.go
+++ b/task_master.go
@@ -60,7 +60,7 @@ type TaskMaster struct {
 	}
 	PagerDutyService interface {
 		Global() bool
-		Alert(incidentKey, desc string, details interface{}) error
+		Alert(incidentKey, desc string, level AlertLevel, details interface{}) error
 	}
 	SlackService interface {
 		Global() bool


rossmcdonald:
@nathanielc I think this is ready to go. The current functionality has changed so that a `kapacitor.WarnAlert` or `kapacitor.CritAlert` will send a `trigger` event to the PagerDuty API (opening an alert). All other `kapacitor.AlertLevel` types will send a `resolve`, resolving the alert.

Testing this with the PagerDuty API, on a critical alert:

```
{
  "client": "kapacitor",
  "client_url": "http://[::]:9092",
  "description": "cpu/kapacitor1 is CRITICAL",
  "details": "{\"Series\":[{\"name\":\"cpu\",\"tags\":{\"cpu\":\"cpu0\",\"host\":\"kapacitor1\"},\"columns\":[\"time\",\"time_guest\",\"time_guest_nice\",\"time_idle\",\"time_iowait\",\"time_irq\",\"time_nice\",\"time_softirq\",\"time_steal\",\"time_system\",\"time_user\",\"usage_guest\",\"usage_guest_nice\",\"usage_idle\",\"usage_iowait\",\"usage_irq\",\"usage_nice\",\"usage_softirq\",\"usage_steal\",\"usage_system\",\"usage_user\"],\"values\":[[\"2016-03-09T20:57:30Z\",0,0,174491.01,11.72,41.94,5.68,107.41,0,341.67,167.67,0,0,21.72131147540617,0,1.2295081967237784,0,34.73360655744749,0,42.213114754184005,0.1024590163935572]]}],\"Err\":null}",
  "event_type": "trigger",
  "incident_key": "cpu/kapacitor1",
  "service_key": "<SERVICE KEY>"
}
```

And then on an okay:

```
{
  "client": "kapacitor",
  "client_url": "http://[::]:9092",
  "description": "cpu/kapacitor1 is OK",
  "details": "{\"Series\":[{\"name\":\"cpu\",\"tags\":{\"cpu\":\"cpu0\",\"host\":\"kapacitor1\"},\"columns\":[\"time\",\"time_guest\",\"time_guest_nice\",\"time_idle\",\"time_iowait\",\"time_irq\",\"time_nice\",\"time_softirq\",\"time_steal\",\"time_system\",\"time_user\",\"usage_guest\",\"usage_guest_nice\",\"usage_idle\",\"usage_iowait\",\"usage_irq\",\"usage_nice\",\"usage_softirq\",\"usage_steal\",\"usage_system\",\"usage_user\"],\"values\":[[\"2016-03-09T20:57:50Z\",0,0,174499.04,11.78,42.13,5.68,112.7,0,347.57,167.71,0,0,81.52284263953392,0.30456852791859523,0.406091370558127,0,8.527918781720883,0,9.238578680197262,0]]}],\"Err\":null}",
  "event_type": "resolve",
  "incident_key": "cpu/kapacitor1",
  "service_key": "<SERVICE KEY>"
}
```

The only downside I can see here is that an `.info` will also resolve the alert, but I would assume that people won't want to be alerted on `.info`.

nathanielc:
@rossmcdonald What if `info` alerts where just ignored by PagerDuty? Maybe log an error saying `infos` are ignore and then ignore it? That way infos are not side effecting and the concept of an informational alert doesn't work with PagerDuty since it will page someone. Thoughts?

nathanielc:
Overall looks great, thanks for digging in.

rossmcdonald:
@nathanielc I completely agree. Erring out will be much more straight-forward and ensure an alert is not accidentally resolved prematurely. 

How does this look?

```
[cpu_alert:alert2] 2016/03/09 22:32:40 E! failed to send alert data to PagerDuty: AlertLevel 'info' is currently ignored by the PagerDuty service
```

nathanielc:
:+1:

Issue #321, opened 2016-03-12
All tasks stats


yosiat:
Hi,

I am currently building UI for Kapacitor (that I hope to release in a month) and I am showing a list of all tasks.
I want to add to each task the number of points he processed and how much he outputed (points written or alerts triggered).

In the current api it will be slow, because I will need to fetch all tasks (using "/tasks") and fetch each task and parse it's dot file. and I want to do this every 10 seconds or something.

So I thoought about adding to "/tasks" something like this:
```javascript
{
	Name: "cpu_alert",
	Type: 0,
	DBRPs: [{
		db: "kapacitor_example",
		rp: "default"
	}],
	Enabled: true,
	Executing: true,
	ExecutionStats: {
		Inputs: {
			stream1: 234
		},
		Outputs: {
			alert1: 298,
			influxdbout: 12
		}
	}
}
```

Where the Inputs and Outputs are key value of how much they emitted.
I digged to the source code and found a way to do so, but I am not sure about this, and I want OK before I code it out.

1. I need node to export how much they emitted - and this can be calculated by the summing the "emitted" in Edge statsMap.
2. Add to ExecutingTask function that will return "ExecutionStats" struct that will:
  * Input nodes:
      1. Find them all by checking their type StreamNode/ BatchNode
      2. Get their value - by summing the edge outs emitted
  * Output nodes:
     1. Find the output nodes by the next predicate: node that have parents and doesn't have child nodes. 
     2. Expose function on "Node" (is this the right place) - that will return how much their outputed

@nathanielc  What do you think? am I in the right direction?

yosiat:
Moving discussion to the pull request https://github.com/influxdata/kapacitor/issues/323

PR #323, opened 2016-03-12
[WIP] Add stats to tasks

yosiat:
See issue - #321 

Currently I implemented the input nodes, I am waiting for feedback about this and a suggestion of how to implement the metrics fetching for the outputs.

### DIFF
--- a/node.go
+++ b/node.go
@@ -52,6 +52,10 @@ type Node interface {
 	nodeStatsByGroup() map[models.GroupID]nodeStats
 
 	collectedCount() int64
+
+	emittedCount() int64
+
+	stats() map[string]interface{}
 }
 
 //implementation of Node
@@ -239,11 +243,37 @@ func (n *node) edot(buf *bytes.Buffer, labels bool) {
 	}
 }
 
+// node collected count is the sum of emitted counts of parent edges
 func (n *node) collectedCount() (count int64) {
 	for _, in := range n.ins {
 		count += in.emittedCount()
 	}
 	return
+}
+
+// node emitted count is the sum of collected counts of children edges
+func (n *node) emittedCount() (count int64) {
+	for _, out := range n.outs {
+		count += out.collectedCount()
+	}
+	return
+}
+
+func (n *node) stats() map[string]interface{} {
+	stats := make(map[string]interface{})
+
+	n.statMap.Do(func(kv expvar.KeyValue) {
+		switch v := kv.Value.(type) {
+		case kexpvar.IntVar:
+			stats[kv.Key] = v.IntValue()
+		case kexpvar.FloatVar:
+			stats[kv.Key] = v.FloatValue()
+		default:
+			stats[kv.Key] = v.String()
+		}
+	})
+
+	return stats
 }
 
 // Statistics for a node
--- a/task.go
+++ b/task.go
@@ -301,6 +301,40 @@ func (et *ExecutingTask) registerOutput(name string, o Output) {
 	et.outputs[name] = o
 }
 
+type ExecutionStats struct {
+	TaskStats map[string]interface{}
+	NodeStats map[string]map[string]interface{}
+}
+
+func (et *ExecutingTask) ExecutionStats() (ExecutionStats, error) {
+	executionStats := ExecutionStats{
+		TaskStats: make(map[string]interface{}),
+		NodeStats: make(map[string]map[string]interface{}),
+	}
+
+	// Fill the task stats
+	executionStats.TaskStats["throughput"] = et.getThroughput()
+
+	// Fill the nodes stats
+	err := et.walk(func(node Node) error {
+		nodeStats := node.stats()
+
+		// Add collected and emitted
+		nodeStats["collected"] = node.collectedCount()
+		nodeStats["emitted"] = node.emittedCount()
+
+		executionStats.NodeStats[node.Name()] = nodeStats
+
+		return nil
+	})
+
+	if err != nil {
+		return executionStats, err
+	}
+
+	return executionStats, nil
+}
+
 // Return a graphviz .dot formatted byte array.
 // Label edges with relavant execution information.
 func (et *ExecutingTask) EDot(labels bool) []byte {
--- a/services/task_store/service.go
+++ b/services/task_store/service.go
@@ -49,6 +49,7 @@ type Service struct {
 		StartTask(t *kapacitor.Task) (*kapacitor.ExecutingTask, error)
 		StopTask(name string) error
 		IsExecuting(name string) bool
+		ExecutionStats(name string) (kapacitor.ExecutionStats, error)
 		ExecutingDot(name string, labels bool) string
 	}
 
@@ -684,11 +685,12 @@ func (ts *Service) Disable(name string) error {
 }
 
 type taskInfo struct {
-	Name      string
-	Type      kapacitor.TaskType
-	DBRPs     []kapacitor.DBRP
-	Enabled   bool
-	Executing bool
+	Name           string
+	Type           kapacitor.TaskType
+	DBRPs          []kapacitor.DBRP
+	Enabled        bool
+	Executing      bool
+	ExecutionStats kapacitor.ExecutionStats
 }
 
 func (ts *Service) IsEnabled(name string) (e bool) {
@@ -725,6 +727,15 @@ func (ts *Service) GetTaskInfo(tasks []string) ([]taskInfo, error) {
 				Enabled:   enabled,
 				Executing: ts.TaskMaster.IsExecuting(t.Name),
 			}
+
+			if info.Executing {
+				executionStats, err := ts.TaskMaster.ExecutionStats(t.Name)
+				if err != nil {
+					return fmt.Errorf("failed to fetch execution stats. name: %s, err: %s", t.Name, err)
+				}
+				info.ExecutionStats = executionStats
+			}
+
 			taskInfos = append(taskInfos, info)
 			return nil
 		}
--- a/task_master.go
+++ b/task_master.go
@@ -369,6 +369,17 @@ func (tm *TaskMaster) IsExecuting(name string) bool {
 	return executing
 }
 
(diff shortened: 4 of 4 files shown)


nathanielc:
@yosiat This is great work thanks!

A few administrative things: Can you sing the CLA? and add an entry to the CHANGELOG about this change? Thanks

I like the structure that you have added but I think it would be easy and valuable to expose more stats. 

First what stats are you after?

I see two paths we could go down:

1. Provide just summary stats about the task. Similar to the throughput stat that already exists/
2. Expose all stats from all nodes so that the DOT data does not need to be parsed.

We can even do both.. And I think we should. I'll comment directly in the code where I think we could update what you have already.

task.go
```diff
@@ -291,6 +311,40 @@ func (et *ExecutingTask) registerOutput(name string, o Output) {
 	et.outputs[name] = o
 }
 
+type ExecutionStats struct {
+	// Input nodes and their emitted points
+	// where input nodes - are batch/stream node
+	// for example: stream1 -> 12
+	Inputs map[string]int64
```
nathanielc: We should update the struct to have both summary stats for the task and a map of all stats for the all nodes in the task.

maybe a structure like 
```
type ExecutionStats struct {
    TaskStats map[string]interface{} // stats name -> value i.e throughput -> 4.3 points /sec
    NodeStats map[string]map[string]interface{} // node name -> stat name -> value i.e influxdb_out -> point_written -> 1000
}
```

Then use the statsMap that each node has to populate the node stats maps.

Thoughts?

yosiat:
@nathanielc Thanks!

I already signed the CLA (https://influxdb.com/community/cla.html) and I will add an entry to the change log once I will finish the output nodes as well.

About the structure of stats I will change it but before, I want to clarify the reason for this API - I want to expose general stats - "stats for glance" on all tasks, I don't want to extract all the data from the dot (for example: how much each node processed and how much time it took).

I think it should contain:

- Task stats
 - Throughput of the task - currently the throughput is all Kapacitor throughput I want to know how point s the whole pipeline gets (only those whom filtered by the measurement, db & rp)
 - Filtered points count - how much points we filtered
 - Total points we processed
- Input and Output stats - this the "glance" data of the nodes - I want to know how much points I processed and how much I outputed (alerts/written to influx) _(and this why I did direct mapping from node name to it's output value)_

But in the overall sense I agree this structure is more acceptable:

```go
type ExecutionStats struct {
    TaskStats map[string]interface{}
    NodeStats struct {
     Input map[string]map[string]interface{}
     Output map[string]map[string]interface{}
    }
}
```
I changed NodeStats to be struct that will contain only the Input and Output to note that "for all nodes stats - there is a dot graph, for general data (like input & output), this is the place.

About the values - why this is ``interface{}``` and not floats, I think those metrics should be raw value like "789" and not "789 points/s" - the resolution should be noted on the stat name for example "PointsPerSecond" => 789.

Aside the structure, I don't know how to implement the output nodes stats (as I wrote in the issue), currently I think to fetch all output nodes by predicate: node that have parents and doesn't have child nodes. but I don't know how to expose the output stats - 

1. should it be a function that will expose the "statsMap" or map[string]float ?
2. and where this function will be? on "Node" or add another interface for "OutputNodes" ?

yosiat:
@nathanielc any comment? do you want me to clarify my answer better? currently I am waiting to your response before proceeding on.

nathanielc:
@yosiat Sorry for the delay.

I like the plan, stats at a glance is a good approach. I am a little confused about the `Input`, `Output` stats. Are input stats just counts for how many points a node has received? If so why a map its just a single count per node? Same question for output stats.

> About the values - why this is interface{} and not floats, I think those metrics should be raw value like "789" and not "789 points/s" - the resolution should be noted on the stat name for example "PointsPerSecond" => 789.

Agreed, the interface is so you can use int64 or float64 but should always be one or the other.

yosiat:
@nathanielc it's ok :)
The "inputs" and "outputs" are plural because I want to handle edge cases where user might write something like this:
```javascript
var a = stream.from('a')
var b= stream.from('b')

a.join(b)
  // do some alerting logic
```

And output might be creating an alert and writing to influx

What do you think about the output nodes implementation? what do you prefer - specific function or exposing the statsMap?

yosiat:
Changed the ExecutionStats structure, waiting to @nathanielc to answer on how to implement the output nodes.

nathanielc:
@yosiat I think simpler is better here, since we are doing stats at a glance.

```go
// Stats about the execution of a task
type ExecutionStats struct {
    // Stats global to the task
    // throughput etc...
    TaskStats map[string]interface{}
    // Specific node stats,
    // Emitted, Collected counts plus anything in statsMap.
    NodeStats map[string]interface{}
}
```

NodeStats will be populated by whatever is in statsMap using statsMap.Do and the add in `collected` and `emitted` stats using the `node.collectedCount` and `node.emittedCount` stats. 

Populating TaskStats will be a bit more work since it doesn't already have a framework in place. We should probably add a statsMap to ExecutingTask too if it make sense. Otherwise we can populate TaskStats by aggregating other stats from nodes.

This should get all the stats we were looking for.

yosiat:
@nathanielc So you want ExecutionStats to contain all nodes? If so, what is the difference between this and a Dot?

nathanielc:
@yosiat Its exposed via HTTP in JSON so you do not have to parse DOT since its not a good method for conveying numerical data.

yosiat:
@nathanielc ok, I am doing this right now.
Just to make sure, you wrote:
```go
NodeStats map[string]interface{}
```

instead of:
```go
NodeStats map[string][string]interface{}
```

Was it by mistake?

nathanielc:
@yosiat Yes is should be `NodeStats map[string][string]interface{}`

yosiat:
@nathanielc I finished.

Here is the final result:
```javascript
{
	Name: "cpu_alert",
	Type: 0,
	DBRPs: [{
		db: "kapacitor_example",
		rp: "default"
	}],
	Enabled: true,
	Executing: true,
	ExecutionStats: {
		TaskStats: {
			throuput: 0
		},
		NodeStats: {
			alert2: {
				alerts_triggered: "0",
				avg_exec_time_ns: "0",
				collected: 0,
				emitted: 0
			},
			srcstream0: {
				avg_exec_time_ns: "0",
				collected: 0,
				emitted: 0
			},
			stream1: {
				avg_exec_time_ns: "0",
				collected: 0,
				emitted: 0
			}
		}
	}
}
```

Do you know how change statsMap representation to be numbers and not strings?

nathanielc:
>Do you know how change statsMap representation to be numbers and not strings? 
Yes , have a look at https://github.com/influxdata/kapacitor/blob/master/global_stats.go#L132
Kapacitor has created a bunch of special expvar types just for that purpose.

Also quick typo fix `throuput` -> `throughput`. Looks great.

yosiat:
@nathanielc done.
Do you want me to squash the commits?

nathanielc:
@yosiat yes, please.

yosiat:
@nathanielc done :+1:

node.go
```diff
@@ -246,6 +250,30 @@ func (n *node) collectedCount() (count int64) {
 	return
 }
 
+func (n *node) emittedCount() (count int64) {
+	for _, out := range n.outs {
+		count += out.emittedCount()
```
nathanielc: This needs to read `count += out.collectedCount()` see comment in node.collectedCount

task.go
```diff
@@ -301,6 +301,42 @@ func (et *ExecutingTask) registerOutput(name string, o Output) {
...
+
+	// Fill the nodes stats
+	err := et.walk(func(node Node) error {
+		name := node.Name()
+
+		// Add the node stats
+		executionStats.NodeStats[name] = node.stats()
+
+		// Add collected and emitted
+		nodeStat := executionStats.NodeStats[name]
```
nathanielc: Instead of indexing the map twice. And it reads cleaner so you know they are the same map.
```
nodeStat := node.stats()
executionStats.NodeStats[name]  = nodeStat
```

node.go
yosiat: Can you explain? I don't see any comment on node.collectedCount

task.go
yosiat: done

node.go
nathanielc: Hmm, to comment is gone, what it said was:

`node collected count is the sum of emitted counts of parent edges`

The opposite is true for node emitted count:

`node emitted count is the sum of collected counts of children edges`

Edges can buffer so we care about the actual points this node has collected or emitted.

yosiat: fixed :+1:

node.go
```diff
@@ -246,6 +250,30 @@ func (n *node) collectedCount() (count int64) {
 	return
 }
 
+func (n *node) emittedCount() (count int64) {
+	for _, out := range n.outs {
+		count += out.collectedCount()
```
nathanielc: Could you add those comments in so its not confusing for the next guy? Thanks

yosiat: Just to make sure:
`node collected count is the sum of emitted counts of parent edges` above collectedCount 
`node emitted count is the sum of collected counts of children edges` above emittedCount

?

nathanielc: Yep

yosiat: done :)

nathanielc:
@yosiat Thanks! merging...

yosiat:
@nathanielc I forgot to add a change log entry.

nathanielc:
I added no worries

On Fri, Mar 18, 2016 at 1:58 PM, Yosi Attias <notifications@github.com>
wrote:

> @nathanielc <https://github.com/nathanielc> I forgot to add a change log
> entry.
>
> —
> <https://github.com/influxdata/kapacitor/pull/323#issuecomment-198518730>
>



-- 
*Nathaniel Cook*
Kapacitor Lead
▼▴

*InfluxData.com <http://influxdata.com/>*

Twitter <http://www.twitter.com/nathanielvcook>

PR #347, opened 2016-03-17
Allow /root/go/src to be cached during docker build

jonseymour:
The docker build is rather punishing on 4G bandwidth consumption.

This change allow the user to cache the /root/go/src if they so choose, but
default to not using such a cache.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -7,14 +7,23 @@ set -e
 DIR=$(cd $(dirname ${BASH_SOURCE[0]}) && pwd)
 cd $DIR
 
-
 # Build new docker image
 docker build -f Dockerfile_build_ubuntu64 -t influxdata/kapacitor-builder $DIR
+if test "${KAPACITOR_USE_BUILD_CACHE}" = "true"; then
+	# create a container that owns the /root/go/src volume used as a cache of go dependency downloads
+	# ignore failures, since this usually means the container already exists.
+	docker run --entrypoint=/bin/true --name kapacitor-builder-cache influxdata/kapacitor-builder 2>/dev/null &&
+	VOLUME_OPTIONS=--volumes-from kapacitor-builder-cache || true
+else
+	VOLUME_OPTIONS=
+fi
+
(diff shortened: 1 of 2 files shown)


build.sh
```diff
@@ -7,14 +7,27 @@ set -e
...
+		set -- "${@:1:$(expr $x - 1)}" "${@:$(expr $x + 1):$#}"
+		break
+	fi
+done
 
 # Build new docker image
 docker build -f Dockerfile_build_ubuntu64 -t influxdata/kapacitor-builder $DIR
+if test -n "$KAPACITOR_USE_BUILD_CACHE"; then
+	docker run --entrypoint=/bin/true --name kapacitor-builder-cache influxdata/kapacitor-builder 2>/dev/null || true
+fi
```
nathanielc: What does this do?

build.sh
```diff
@@ -7,14 +7,27 @@ set -e
...
 
+x=0
+for arg in "$@"; do
+	let x=x+1
+	if test "$arg" = "--use-cache"; then
+		KAPACITOR_USE_BUILD_CACHE=defined
+		set -- "${@:1:$(expr $x - 1)}" "${@:$(expr $x + 1):$#}"
+		break
+	fi
+done
```
nathanielc: Instead of arg parsing I would rather that the KAPACITOR_USE_BUILD_CACHE be set externally to the script if one wants to use it. While odd, this is currently the way behavior is determined in the script.

Something like:

`KAPACITOR_USE_BUILD_CACHE=${KAPACITOR_USE_BUILD_CACHE-false}`

build.sh
jonseymour: This line creates a  container  - kapacitor-builder-cache - that owns the /root/go/src volume that acts as the cache of dependencies.

jonseymour:
@nathanielc - thanks for your review. Amended per your comments.

nathanielc:
@jonseymour Thanks!

PR #350, opened 2016-03-17
apply env to array elements

### LINKED ISSUE #348: applyEnvOverrides fails to override array elements
The following environment override:

```
KAPACITOR_INFLUXDB_URLS_0=http://influx:8086 kapacitord config
```

doesn't have the expected effect.

An attempt is made to ecursively apply overrides to array/slice elements, but the attempt fails because the recursive call returns early because the reflect.Value being operated on by the recursive call is not a struct.

jonseymour:
This fixes #348.

In particular:

```KAPACITOR_INFLUXDB_0_URLS_0=http://localhost:9000 kapacitord config```

should now produce a config which does the expected thing.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -103,6 +103,7 @@ For example:
 - [#266](https://github.com/influxdata/kapacitor/issues/266): Fixes error log for HipChat that is not an error.
 - [#333](https://github.com/influxdata/kapacitor/issues/333): Fixes hang when replaying with .stats node. Fixes issues with batch and stats.
 - [#340](https://github.com/influxdata/kapacitor/issues/340): BREAKING: Decouples global setting for alert handlers from the state changes only setting.
+- [#348](https://github.com/influxdata/kapacitor/issues/348): config.go: refactor to simplify structure and fix support for array elements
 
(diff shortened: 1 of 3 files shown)


nathanielc:
@jonseymour Thanks for the PR, I know I don't have any tests around config parsing, but could you add a test that demonstrates the new behavior? Don't worry about testing everything else just the changes. We will get to the other tests in time.

jonseymour:
Ok, I'll add tests. I'll have to abstract the call to os.GetEnv() to allow tests to mock this call, as required.

jonseymour:
@nathanielc I've extended a test and done a more complete reorganisation to simplify the structure of the applyEnvOverrides function.

jonseymour:
Actually, I missed an additional simplifcation that is now possible. I'll re-roll.

nathanielc:
@jonseymour Just to make sure, you made that last simplification correct? Also can you add a changelog entry? Otherwise this looks good to go. Thanks!

jonseymour:
Yes, simplifications complete. Changelog now updated.

offlinehacker:
Nice, this should remove sed magic from my Docker image, thanks!

PR #353, opened 2016-03-17
vendor InfluxDB core for the query engine changes

jonseymour:
Re-apply the initial vendoring change from the master branch onto v0.10 so that the v0.10.x branch can be compiled.


nathanielc:
@jonseymour Thanks for back porting this.

PR #364, opened 2016-03-21
Add alert logging + LogNode for easier debugging

### LINKED ISSUE #358: logging of important events within the system
While changing the log level from the default info to debug, it looks like the state changes/detections (crit, warn, ...) are not being logged. Tt makes sense to have some logging indicative of important events within the system.

nathanielc:
Fixes #358


yosiat:
@nathanielc maybe adding some lambda expression to the log can be helpful or threshold for filtering noise.

but how can you debug AlertNode if it dosen't exposes point?

nathanielc:
@yosiat If you want to log less add a WhereNode or a SampleNode  in front of the LogNode.

> but how can you debug AlertNode if it dosen't exposes point?

Ah yes, it should also log the alert data....

Issue #337, opened 2016-03-16
Add the ability to active all tick scripts with one command


panda87:
I see this feature very useful when using Kapacitor in docker containers, when the run command will active all TICK's at one single command, instead of follow one after one.

OrangeTux:
Maybe a solution to enable/disable TICK scripts using glob style pattern matching?

I often prefix the names of my TICK scripts like this:

```
frontend_memory
frontend_deadman
backend_mysql
```

If i want to enable all scripts I could use `kapacitor enable *`, but if I only want to enable scripts for frontend I could use `kapacitor enable frontend_*`.

nathanielc:
I like the idea of glob syntax. This would not be to hard to add, PRs welcome.

yosiat:
@nathanielc I can implement this.. Where do we want to happen?

- Client will send to the server - enable/frontend_* 
- Client will fetch all tasks and call enable per task?

nathanielc:
@yosiat I think having the server implement the glob logic is the right place. Thanks!

yosiat:
@nathanielc I will do this, I think we should enable this feature for disable and reload as well.

nathanielc:
Agreed lets make the change consistent.

yosiat:
@nathanielc This is harder than I thought.
I can write this code with lots of duplicated, but I decided to add "FindTasks(predicate func(string)bool) []rawTasks" so methods like:
- GetTasksInfo
- Enable
- Disable

will use them.. I will send PR once it will be finished.

yosiat:
@nathanielc I started to implement this and found it difficult, there two ways to implement this, and please tell me what you prefer:

Implement a function that will get a predicate with the task name and will return an array of rawTask, it will fetch the boltdb and will return all the matching tasks, we can use this in:
 - GetTasksInfo - to filter by the given array of tasks (?tasks=...)
 - Enable/Disable - the predicate will be global function
The downside of this will be - we are accessing twice the boltdb - one to fetch the list of tasks and then to enable them/disable them.

The other option will be less DRY, and will be to filter in each boltdb acess before enabling/disbeling.

nathanielc:
@yosiat The task_store has been implemented with simplicity in mind not performance. As a result the predicate function sounds better. Even with 1000 tasks the slow down should not even be noticeable and its a low throughput action.

yosiat:
@nathanielc agreed, I will send PR in the weekend. sorry for the delay!

[...]

PR #381, opened 2016-03-26
Enable/Disable by glob

yosiat:
Issue - #337 

- Add FindTasks method for fetching all rawTask[] by a predicate
- Enabling/Disabling by glob

For example:
`> kapacitor enable "window_*"` 
Will enable all tasks that starts with "window_"
The same for disable

**Notes:**
- I think this would be nice for delete as well
- The enable is not efficient as possible - I am accessing boltdb for each task, I can change it, but I think it will hurt the readability
- If at least one task is failed to enable/disable I am stopping and returning an error - are we ok about this behaviour?

### DIFF
--- a/cmd/kapacitor/main.go
+++ b/cmd/kapacitor/main.go
@@ -639,6 +639,16 @@ func enableUsage() {
 	var u = `Usage: kapacitor enable [task name...]
 
 	Enable and start a task running from the live data.
+
+For example:
+
+    You can enable by specific task name.
+
+    $ kapacitor enable cpu_alert
+
+		Or, you can enable by glob:
+
+    $ kapacitor enable *_alert
 `
 	fmt.Fprintln(os.Stderr, u)
 }
@@ -678,6 +688,16 @@ func disableUsage() {
 	var u = `Usage: kapacitor disable [task name...]
 
 	Disable and stop a task running.
+
+For example:
+
+    You can disable by specific task name.
+
+    $ kapacitor disable cpu_alert
+
+		Or, you can disable by glob:
+
+    $ kapacitor disable *_alert
 `
 	fmt.Fprintln(os.Stderr, u)
 }
@@ -717,6 +737,16 @@ func reloadUsage() {
 	var u = `Usage: kapacitor reload [task name...]
 
 	Disable then enable a running task.
+
+For example:
+
+    You can reload by specific task name.
+
+    $ kapacitor reload cpu_alert
+
+		Or, you can reload by glob:
+
+    $ kapacitor reload *_alert
 `
 	fmt.Fprintln(os.Stderr, u)
 }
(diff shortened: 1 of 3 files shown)


nathanielc:
> I think this would be nice for delete as well

+1

>The enable is not efficient as possible - I am accessing boltdb for each task, I can change it, but I think it will hurt the readability

Simple is better

> If at least one task is failed to enable/disable I am stopping and returning an error - are we ok about this behavior?

Yes

services/task_store/service.go
```diff
@@ -706,56 +762,95 @@ func (ts *Service) IsEnabled(name string) (e bool) {
...
+			return true, nil
+		}
+
+		// This code is not the optimal code ever, we could use map here
+		// but I am assuming the asked tasks count will be small
+		for _, askedTaskName := range tasks {
+			if askedTaskName == taskName {
+				return true, nil
+			}
+		}
```
nathanielc: If we are not doing any matching here why are we not looping over `LoadRaw`?

yosiat: Umm.. I agree, but the reason we need to handle the case where tasks is empty, and we want to fetch all tasks.

The looping LoadRaw is more simple & elegant.. do you have any way can do this and still support the empty tasks array?

nathanielc: I may be missing it but I do not see the change in behavior from the previous implementation to this one?

yosiat: Let's make sure we are on the same idea, you are suggesting that instead of using FindTasks and looking at the bucket we will do something like this:
```
for taskName := range tasks {
   raw:=ts.LoadRaw(taskName)
  // convert to TaskSummaryInfo, and IsEnabled, and ExecutionStatus
}
```
If you are suggesting doing this approach, we have the case where the tasks are empty and we want to fetch all tasks - "/tasks",  so we are using FindTasks to find the matching tasks.

nathanielc: I am not suggesting that, but rather looking at the original implementation it handled the case where tasks was empty via the `ForEach` boltdb bucket call. I see that method has changed but I do not see how its behavior has changed so I am confused as to why it was modified. I am probably just missing something....

yosiat: ok, you are suggesting to undo the change in this function? I changed this function to use the FindTasks, but after a little bit of thinking it makes things harder.

nathanielc: Yes, there is no need to use the FindTasks if it is not necessary.

yosiat:
@nathanielc are we ok on this PR or do you want me to implement delete by glob as well before we merge?

yosiat:
Added delete by glob

nathanielc:
@yosiat This looks good to go. Do you want to add a CHANGELOG entry real quick?

yosiat:
I will do in one hour, I forgot to update the usage for the relevant
commands in kapacitor cmd

yosiat:
@nathanielc Updated and added help usage..
please make sure it's ok, and I will squash the commits

cmd/kapacitor/main.go
```diff
@@ -874,7 +904,24 @@ func deleteUsage() {
...
+
+    $ kapacitor delete tasks *_alert 
+
+		You can delete recordings:
+
+    $ kapacitor delete recordings b0a2ba8a-aeeb-45ec-bef9-1a2939963586 
+
+		Or, you can enable by glob:
+
+    $ kapacitor enable *_alert
```
nathanielc: These lines look like there are extras?

yosiat: Oops.. I did copy/paste, Will fix this.

nathanielc:
@yosiat Looks good, I'll merge when its squashed

yosiat:
@nathanielc squashed

fatihcode:
@yosiat : does it mean `kapacitor enable "*"` and `kapacitor disable "*"` enabled and disable all the tasks?

yosiat:
@fatihcode yes

Issue #368, opened 2016-03-22
Alert node log files unreadable


gargooza:
I am running kapacitor as a service with user/group = kapacitor/kapacitor. Log files generated by alert nodes are unreadable to other users on the system since the log files have 0600 permissions and are owned by the kapacitor user.

Would it be possible to ease up on those restrictive log file permissions (https://github.com/influxdata/kapacitor/blob/7fdc137dc497d50c7ae2d6e2a481031a8b126511/alert.go#L616)? Maybe make an option to configure log file permissions for kapacitord in a `kapacitor.conf` file? I'd like to make them 0644 by default, but I'm sure that's a security concern for others and probably not something to be hardcoded.

nathanielc:
@gargooza A very reasonable request. This should be easy to add.

1. Add a field to the LogHandler struct int64  https://github.com/influxdata/kapacitor/blob/master/pipeline/alert.go#L432
2. Implement using the new field in the handleLog method  https://github.com/influxdata/kapacitor/blob/master/alert.go#L610

PRs welcome

yosiat:
@nathanielc do you want something like this:
```go
alert()
.log('/tmp/alert.log', 0644)
```

If so, does tick script support optional parameters?

I will implement this :)

nathanielc:
I was thinking something more along the lines of,

alert()
   .log('some path')
       .mode(644)

Keeps it consistent with the other handlers

nathanielc:
Closed with #404

PR #404, opened 2016-04-03
Adding mode to alert's LogHandler

yosiat:
Issue - #368

### DIFF
--- a/pipeline/alert.go
+++ b/pipeline/alert.go
@@ -16,6 +16,9 @@ const defaultMessageTmpl = "{{ .ID }} is {{ .Level }}"
 // Default template for constructing a details message.
 const defaultDetailsTmpl = "{{ json . }}"
 
+// Default log mode for file
+const defaultLogFileMode = 0600
+
 // An AlertNode can trigger an event of varying severity levels,
 // and pass the event to alert handlers. The criteria for triggering
 // an alert is specified via a [lambda expression](/kapacitor/v0.11/tick/expr/).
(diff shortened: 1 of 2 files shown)


pipeline/alert.go
```diff
@@ -428,11 +428,22 @@ type ExecHandler struct {
...
+//    stream
+//         |alert()
+//					.log('/tmp/alert')
+//					.mode(0644)
 // tick:property
 func (a *AlertNode) Log(filepath string) *LogHandler {
 	log := &LogHandler{
 		AlertNode: a,
 		FilePath:  filepath,
+		Mode:      0600,
```
nathanielc: We should define this as a constant `Default LogFileMode` or something.

nathanielc:
@yosiat Overall looks great, thanks for the PR.

yosiat:
@nathanielc Fixed :+1:

pipeline/alert.go
```diff
@@ -16,6 +16,9 @@ const defaultMessageTmpl = "{{ .ID }} is {{ .Level }}"
 // Default template for constructing a details message.
 const defaultDetailsTmpl = "{{ json . }}"
 
+// Default log mode for file
+const defaultLogFileMode = 0666
```
nathanielc: We still want a default of `0600` right?

yosiat: oops :(

eran-totango:
@yosiat Hey, this doesn't seem to work.
Kapcitor creates the alert log file with different permissions that what i tried to set.

lafrech:
I just used this and it works.

Here's the proper syntax, applied to an example from the docs:

```
dbrp "telegraf"."autogen"

stream
    // Select just the cpu measurement from our example database.
    |from()
        .measurement('cpu')
    |alert()
        .crit(lambda: int("usage_idle") < 100)
        // Whenever we get an alert write it to a file.
        .log('/tmp/alerts.log')
        .mode(0644)
```

PR #407, opened 2016-04-04
Buffer pooling in alert templates

yosiat:
Hi,

I added benchmarks for alert node and found a quick performance boost,
After memory profiling, I found that we are allocating new buffer for the templates of id,message and details, so I wrapped this in a buffer pool, and got a performance improvement:

```
benchmark                            old ns/op       new ns/op       delta
Benchmark_T10_P500_AlertTask-4       210263431       159024737       -24.37%
Benchmark_T10_P50000_AlertTask-4     18718435131     15589249575     -16.72%
Benchmark_T1000_P500_AlertTask-4     21833771619     20820186700     -4.64%

benchmark                            old allocs     new allocs     delta
Benchmark_T10_P500_AlertTask-4       490648         467352         -4.75%
Benchmark_T10_P50000_AlertTask-4     50044467       47560324       -4.96%
Benchmark_T1000_P500_AlertTask-4     48492207       46626729       -3.85%

benchmark                            old bytes      new bytes      delta
Benchmark_T10_P500_AlertTask-4       36155296       33162568       -8.28%
Benchmark_T10_P50000_AlertTask-4     3686308856     3347320896     -9.20%
Benchmark_T1000_P500_AlertTask-4     3604462496     3463221064     -3.92%
```
_Compared to master_

### DIFF
--- a/alert.go
+++ b/alert.go
@@ -11,6 +11,7 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"sync"
 	text "text/template"
 	"time"
 
@@ -100,6 +101,8 @@ type AlertNode struct {
 	idTmpl      *text.Template
 	messageTmpl *text.Template
 	detailsTmpl *html.Template
+
+	bufPool sync.Pool
 }
 
 // Create a new  AlertNode which caches the most recent item and exposes it over the HTTP API.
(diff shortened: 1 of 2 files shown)


yosiat:
by the way, I think we can improve further if we will move from text/json to https://github.com/pquerna/ffjson

nathanielc:
@yosiat Thanks this looks great. 

> by the way, I think we can improve further if we will move from text/json to https://github.com/pquerna/ffjson

I think we will hold off on this one for now since the number of triggered alerts should be orders of magnitude less than the number of points processed. Adding the extra dependency seems over kill at this point.

yosiat:
@nathanielc I do agree, but when I profiled the benchmarks I found out that text/json is called a lot, when I don't trigger the alert (I am not calling post(), log(), or something)

It might be in the id/message/details rendering, I will check that out.

PR #415, opened 2016-04-05
[Perf] Add support for int comparisons in StatefulExpression.EvalBool

yosiat:
Hi,

While adding the benchmarks to AlertNode (see #407 ), I found out the EvalPredicate is a tiny bottleneck, 
so I dug into this, and found out that StatefulExpression dosen't support int comparison so I added it and it improved the perf a bit ;) 

**StatefulExpression EvalBool benchmarks:**
```
benchmark                             old ns/op     new ns/op     delta
Benchmark_EvalBool_1ScopeItem-4       3069          3033          -1.17%
Benchmark_EvalBool_11_ScopeItem-4     3043          2708          -11.01%

benchmark                             old allocs     new allocs     delta
Benchmark_EvalBool_1ScopeItem-4       4              4              +0.00%
Benchmark_EvalBool_11_ScopeItem-4     4              4              +0.00%

benchmark                             old bytes     new bytes     delta
Benchmark_EvalBool_1ScopeItem-4       64            64            +0.00%
Benchmark_EvalBool_11_ScopeItem-4     64            64            +0.00%
```

**Alert benchmark:**
```
benchmark                            old ns/op       new ns/op       delta
Benchmark_T10_P500_AlertTask-4       145918779       143226340       -1.85%
Benchmark_T10_P50000_AlertTask-4     13739584526     13095020431     -4.69%
Benchmark_T1000_P500_AlertTask-4     15098593884     12948902710     -14.24%

benchmark                            old allocs     new allocs     delta
Benchmark_T10_P500_AlertTask-4       466191         466201         +0.00%
Benchmark_T10_P50000_AlertTask-4     47548950       47548763       -0.00%
Benchmark_T1000_P500_AlertTask-4     46101716       46098518       -0.01%

benchmark                            old bytes      new bytes      delta
Benchmark_T10_P500_AlertTask-4       32825881       32826965       +0.00%
Benchmark_T10_P50000_AlertTask-4     3344051504     3344020128     -0.00%
Benchmark_T1000_P500_AlertTask-4     3275554016     3274434464     -0.03%
```

_Those benchmarks are compared to the master (the edge buffer size is the default value)_

### DIFF
--- a/tick/eval_test.go
+++ b/tick/eval_test.go
@@ -266,3 +266,105 @@ var s2 = a.structB()
 		t.Fatal("expected error from Evaluate")
 	}
 }
+
+func Test_EvalBool(t *testing.T) {
+	n := &tick.BinaryNode{
+		Operator: tick.TokenGreater,
+		Left: &tick.ReferenceNode{
+			Reference: "value",
+		},
+		Right: &tick.NumberNode{
+			IsInt: true,
+			Int64: 20,
+		},
+	}
+
+	se := tick.NewStatefulExpr(n)
+
+	executionScope := tick.NewScope()
+	executionScope.Set("value", int64(40))
+	b, err := se.EvalBool(executionScope)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	t.Logf("Result: %v", b)
+}
+
+func Benchmark_EvalBool_1ScopeItem(b *testing.B) {
+	n := &tick.BinaryNode{
+		Operator: tick.TokenGreater,
+		Left: &tick.ReferenceNode{
+			Reference: "value",
+		},
+		Right: &tick.NumberNode{
+			IsInt: true,
+			Int64: 20,
+		},
+	}
+
+	se := tick.NewStatefulExpr(n)
+
+	b.ReportAllocs()
+	b.ResetTimer()
+
+	for i := 0; i < b.N; i++ {
+		// Don't time the creation of the tick scope
+		b.StopTimer()
+		executionScope := tick.NewScope()
+		// insert the value to the scope
+		executionScope.Set("value", int64(40))
+		b.StartTimer()
+
+		result, err := se.EvalBool(executionScope)
+		if err != nil {
+			b.Fatal(err)
+		}
+
+		if !result {
+			b.Fatalf("Got strange result: %v", result)
+		}
+	}
+
+}
+
+func Benchmark_EvalBool_11_ScopeItem(b *testing.B) {
+	n := &tick.BinaryNode{
+		Operator: tick.TokenGreater,
+		Left: &tick.ReferenceNode{
+			Reference: "value",
+		},
+		Right: &tick.NumberNode{
+			IsInt: true,
+			Int64: 20,
+		},
+	}
+
+	se := tick.NewStatefulExpr(n)
+
+	b.ReportAllocs()
+	b.ResetTimer()
+
+	for i := 0; i < b.N; i++ {
+		// Don't time the creation of the tick scope
+		b.StopTimer()
+		executionScope := tick.NewScope()
+		// insert the value to the scope
+		executionScope.Set("value", int64(40))
+		// insert some random values (this can be tags, fields)
+		for i := 0; i < 10; i++ {
+			executionScope.Set(string(i), int64(i))
+		}
+
+		b.StartTimer()
+		result, err := se.EvalBool(executionScope)
+		if err != nil {
+			b.Fatal(err)
+		}
+
+		if !result {
+			b.Fatalf("Got strange result: %v", result)
+		}
+	}
+
+}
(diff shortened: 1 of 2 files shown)


nathanielc:
@yosiat I am a little confused:

Int comparisons worked before but were always casted to floats if one of the operands was a float. Is the PR to add new functionality or show that by not casting to float and doing direct integer comparisons things are faster?

tick/eval_test.go
```diff
@@ -266,3 +266,109 @@ var s2 = a.structB()
...
+	scope := tick.NewScope()
+	a := &structLambda{}
+	scope.Set("a", a)
+
+	err := tick.Evaluate(script, scope)
+	if err != nil {
+		b.Fatal(err)
+	}
+
+	se := tick.NewStatefulExpr(a.Test)
```
nathanielc: Instead of parsing the script and grabing the tick.Node you can create it directly like this:

```
n := &BinaryNode{
    Operator: TokenGreater,
    Left: &ReferenceNode{
        Reference: "value",
   },
    Right: &NumberNode{
       IsInt: true,
       Int64: 20,
   }
}

se := tick.NewStatefulExpr(n)
```

yosiat:
@nathanielc It started by reading the code of evalBinary and found out there is cast to float64 even when both sides are int64, so I changed this for clarity, I found this very confusing.
And then I checked out the performance, and as we can see it improved.

The initial purpose was to improve performance.

tick/eval_test.go
yosiat: Umm.. I don't find this a big change, do you want me to change?

nathanielc: Yes, please make the change. I find it is easier to understand what is going on since I don't think someone new to the code would have any clue as to why `a.Test` is populated with the parsed lambda expression tick.Node. I had to take a second look to figure it out.

nathanielc:
> The initial purpose was to improve performance.

Ok, makes sense.

nathanielc:
@yosiat With the other work you are doing should we close out this PR since most of it will be replaced?

yosiat:
@nathanielc I think we should merge this PR, because we don't know when we will finish the "specialization" pr and this pr yields some nice perf improvement.

yosiat:
@nathanielc I am closing this PR, since I will do this improvement on the typed stack approach (I will submit for this today/tomorrow)

yosiat:
@nathanielc What is the status of this PR? I think we should merge it..

nathanielc:
I think its ready, but first lets merge #458 since its now has tests for this change. Then a rebase to see we are green and this should be good to go.

yosiat:
@nathanielc I squashed :+1: 

By the way, those are the results for this change on the new benchmarks that I am working on (https://github.com/yosiat/kapacitor/commit/50a94c906a93b984d2449e45646492acae3009d4):
```
name                                                                   old time/op    new time/op    delta
_EvalBool_OneOperator_NumberFloat64_NumberInt64-4                        1.85µs ± 0%    1.66µs ± 0%  -10.26%
_EvalBool_OneOperator_NumberInt64_NumberInt64-4                          1.64µs ± 0%    1.63µs ± 0%   -0.43%
_EvalBool_OneOperator_ReferenceNodeFloat64_NumberInt64-4                 2.00µs ± 0%    1.93µs ± 0%   -3.65%
_EvalBool_OneOperator_ReferenceNodeInt64_ReferenceNodeInt64-4            1.94µs ± 0%    2.02µs ± 0%   +4.34%
_EvalBool_OneOperatorWith11ScopeItem_ReferenceNodeInt64_NumberInt64-4    2.12µs ± 0%    2.20µs ± 0%   +3.78%

name                                                                   old alloc/op   new alloc/op   delta
_EvalBool_OneOperator_NumberFloat64_NumberInt64-4                         80.0B ± 0%     80.0B ± 0%   +0.00%
_EvalBool_OneOperator_NumberInt64_NumberInt64-4                           80.0B ± 0%     80.0B ± 0%   +0.00%
_EvalBool_OneOperator_ReferenceNodeFloat64_NumberInt64-4                  64.0B ± 0%     64.0B ± 0%   +0.00%
_EvalBool_OneOperator_ReferenceNodeInt64_ReferenceNodeInt64-4             64.0B ± 0%     64.0B ± 0%   +0.00%
_EvalBool_OneOperatorWith11ScopeItem_ReferenceNodeInt64_NumberInt64-4     64.0B ± 0%     64.0B ± 0%   +0.00%

name                                                                   old allocs/op  new allocs/op  delta
_EvalBool_OneOperator_NumberFloat64_NumberInt64-4                          5.00 ± 0%      5.00 ± 0%   +0.00%
_EvalBool_OneOperator_NumberInt64_NumberInt64-4                            5.00 ± 0%      5.00 ± 0%   +0.00%
_EvalBool_OneOperator_ReferenceNodeFloat64_NumberInt64-4                   4.00 ± 0%      4.00 ± 0%   +0.00%
_EvalBool_OneOperator_ReferenceNodeInt64_ReferenceNodeInt64-4              3.00 ± 0%      3.00 ± 0%   +0.00%
_EvalBool_OneOperatorWith11ScopeItem_ReferenceNodeInt64_NumberInt64-4      4.00 ± 0%      4.00 ± 0%   +0.00%                                                                                                         
```
_I compared using benchstat instead of benchcmp_

yosiat:
I am closing this pull request because I am going to submit another pull request that improve performance vastly.

After the another pull request will get merged we can decide if we want this pull request.

PR #422, opened 2016-04-06
StatefulExpression: Adding tests for BinaryNode comparison operator:

yosiat:
* Test cases
  * BinaryNode{ NumberNode, NumberNode }, BinaryNode{ ReferenceNode, NumberNode }
  * BinaryNode{ BoolNode, BoolNode }, BinaryNode{ ReferenceNode, BoolNode }
  * BinaryNode{ StringNode, StringNode }, BinaryNode{ ReferenceNode, StringNode }
  * BinaryNode{ StringNode, RegexNode }
  * Type mismatches (comparing bool to string, int to string, etc)
* Exporting TokenType so it can be used in tests.

Caveats:

1. I didn't tested the math operator since I tested only EvalBool (I can add tests for EvalNum, if it's required).
2. I didn't know how to make the EvalBool return a ReferenceNode - @nathanielc , is this possible?

### DIFF
--- a/tick/stateful_expr_test.go
+++ b/tick/stateful_expr_test.go
@@ -1,0 +1,599 @@
+package tick_test
+
+import (
+	"errors"
+	"regexp"
+	"strings"
+	"testing"
+
+	"github.com/influxdata/kapacitor/tick"
+)
+
+type keyStruct struct {
+	lhs interface{}
+	rhs interface{}
+	op  tick.TokenType
+}
+
+func TestStatefulExpression_EvalBool_BoolNode(t *testing.T) {
+	leftValues := []interface{}{true, false}
+
+	// Right values are the same as left, just add a mismatch case
+	rightValues := []interface{}{true, false, "NON_BOOL_VALUE"}
+	operators := []tick.TokenType{tick.TokenEqual, tick.TokenNotEqual, tick.TokenAnd, tick.TokenOr, tick.TokenLess}
+
+	createBoolNode := func(v interface{}) tick.Node {
+		if strValue, isString := v.(string); isString {
+			return &tick.StringNode{
+				Literal: strValue,
+			}
+		}
+
+		return &tick.BoolNode{
+			Bool: v.(bool),
+		}
+	}
+
+	runEvalBoolTests(t, createBoolNode, leftValues, rightValues, operators, map[keyStruct]bool{
+		// Left: True, Right: True
+		keyStruct{true, true, tick.TokenEqual}:    true,
+		keyStruct{true, true, tick.TokenNotEqual}: false,
+		keyStruct{true, true, tick.TokenAnd}:      true,
+		keyStruct{true, true, tick.TokenOr}:       true,
+
+		// Left: True, Right: False
+		keyStruct{true, false, tick.TokenEqual}:    false,
+		keyStruct{true, false, tick.TokenNotEqual}: true,
+		keyStruct{true, false, tick.TokenAnd}:      false,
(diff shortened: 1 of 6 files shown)


yosiat:
Oops, I forgot to add unknown operators, I will fix this ~

tick/stateful_expr_test.go
```diff
@@ -0,0 +1,398 @@
...
+				Float64: v.(float64),
+			}
+		}
+
+		// For the error case
+		if reflect.TypeOf(v).Kind() == reflect.String {
+			return &tick.StringNode{
+				Literal: v.(string),
+			}
+		}
```
nathanielc: Instead of using reflect can we just use a type switch?

```go
switch value:=v.(type){
 case int64:
    return &tick.NumberNode{
				IsInt: true,
				Int64: value,
			}
case float64:
    return &tick.NumberNode{
 				IsFloat: true,
 				Float64: value,
 			}
//etc...
}
```

nathanielc:
@yosiat I like how this is coming together :+1:

yosiat:
@nathanielc I finished, we are now at 69.2% coverage for tick and 55.2% coverage for StatefulExpression :)

I added:
  * Invalid comparing operators
  * evalBinary - referenced scope values doesn't exist

And before this is mergeable I want to cover the case where the result of EvalBinary can return ReferenceNode, is this possible?

yosiat:
This pull request and #425 are failing because of:
```
Downloading vet tool...
[DEBUG] go get golang.org/x/tools/cmd/vet


Executed command failed!
-- Command run was: go get golang.org/x/tools/cmd/vet
-- Failure was: package golang.org/x/tools/cmd/vet: cannot find package "golang.org/x/tools/cmd/vet" in any of:
	/usr/local/go/src/golang.org/x/tools/cmd/vet (from $GOROOT)
	/root/go/src/golang.org/x/tools/cmd/vet (from $GOPATH)
```

If I am not mistaken the "go tool vet" is baked in go and there is no need to "go get" it.

nathanielc:
> I didn't know how to make the EvalBool return a ReferenceNode - @nathanielc , is this possible?

No, EvalBool will always return true or false. 

I 'll look into the build failure, looks like something changed upstream....

yosiat:
EvalBool can return reference node, for: `"value"` where value is boolean, this can happen where someone have tag with boolean value and he wants to filter by it.

nathanielc:
Ok so which EvalBool are you talking about? I was thinking of this function  https://github.com/influxdata/kapacitor/blob/master/tick/stateful_expr.go#L34

yosiat:
@nathanielc so am I, and I am not mistaken, those scripts:
```
from('..').where(lambda: "yosi")
```

```
from('..').alert().id().crit(lambda: "yosi")
```

Creates stateful expression

nathanielc:
@yosiat any `lambda` expression creates a stateful-expression 

 EvalBool will look at the value of the reference node and return true or false. It will not return the reference node itself. To test such a case create a StefulExpr with the ReferenceNode directly:

```go
se := NewStatefulExpr(&ReferenceNode{...})
scope := ...
se.EvalBool(scope)
```

yosiat:
@nathanielc Ok, I asked the wrong question, I meant to ask how in EvalBool the stack can pop a ReferenceNode, but found the answer myself and fixed

nathanielc:
@yosiat Build issue fixed...

yosiat:
@nathanielc rebased, should be ok now

PR #428, opened 2016-04-07
Remove go get vet tool as its part of go by default, sorry Ross

nathanielc:


nathanielc:
@rossmcdonald FYI, since the go vet tool is now part of main I don't need to explicitly download it. It was removed from tools/master so it now fails if I try.

Issue #331, opened 2016-03-15
Panic with join + eval


nathanielc:
Exceprt of TICKscript
```
join(ssCpuRawSystem,ssCpuRawNice,ssCpuRawWait,ssCpuRawInt,ssCpuRawKernel,ssCpuRawSoftIRQ,ssCpuRawIdle) 
            .eval(lambda: "ssCpuRawUser.ssCpuRawUserDeriv" + "ssCpuRawSystem.ssCpuRawSystemDeriv") 
                         .as('divisor') 
                         .keep() 
                       .eval(lambda: "ssCpuRawUser.ssCpuRawUserDeriv" / "divisor" * float(100), lambda: "ssCpuRawSystem.ssCpuRawSystemDeriv" / "divisor" * float(100)) 
                         .as('user', 'system') 
```

all other tasks were disabled to ensure a cleaner stacktrace: 
```
2016/03/15 09:48:33 Using configuration at: /etc/kapacitor/kapacitor.conf 
panic: runtime error: index out of range 

goroutine 120 [running]: 
github.com/influxdata/kapacitor.(*group).collect(0xc8200144b0, 0x0, 0x7fe0f0b3dc40, 0xc8201e4500) 
        /root/go/src/github.com/influxdata/kapacitor/join.go:190 +0x778 
github.com/influxdata/kapacitor.(*group).run(0xc8200144b0) 
        /root/go/src/github.com/influxdata/kapacitor/join.go:173 +0xf8 
created by github.com/influxdata/kapacitor.(*JoinNode).getGroup 
        /root/go/src/github.com/influxdata/kapacitor/join.go:110 +0x333 

goroutine 1 [chan receive, 1 minutes]: 
main.(*Main).Run(0xc8203b3f10, 0xc82000a1c0, 0x2, 0x2, 0x0, 0x0) 
        /root/go/src/github.com/influxdata/kapacitor/cmd/kapacitord/main.go:96 +0xbf1 
main.main() 
        /root/go/src/github.com/influxdata/kapacitor/cmd/kapacitord/main.go:41 +0x363 

goroutine 17 [syscall, 1 minutes, locked to thread]: 
runtime.goexit() 
        /usr/local/go/src/runtime/asm_amd64.s:1721 +0x1 

goroutine 5 [syscall, 1 minutes]: 
os/signal.loop() 
        /usr/local/go/src/os/signal/signal_unix.go:22 +0x18 
created by os/signal.init.1 
[...]

PR #431, opened 2016-04-08
fix panic if as() is inconsistent with join()

jonseymour:
Previously, if the arguments passed to as() were not consistent with
the number of arms of a join() node, an index out of range panic
would occur at runtime.

Now, we fail during definition of the tick script if the number of
arguments to the as() method is inconsistent or if the as() method was omitted completely.

### DIFF
--- a/join.go
+++ b/join.go
@@ -3,7 +3,6 @@ package kapacitor
 import (
 	"fmt"
 	"log"
-	"strings"
 	"sync"
 	"time"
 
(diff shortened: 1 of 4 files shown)


jonseymour:
This should fix #331

jonseymour:
Having fixed this issue, it isn't clear to me why the join.as() properties method has the behaviour it does.

In particular, it is defined as adding a prefix to the names derived from each parent stream. So, consider this example:

```var metric1 = stream
        |from()
                .database('sampledb')
                .retentionPolicy('default')
                .measurement('sample')
                .groupBy('node')
        |window()
                .align()
                .period(5m)
                .every(5m)
        |shift(-5m)
        |mean('metric1').as('metric1')


var metric2 = stream
        |from()
                .database('sampledb')
                .retentionPolicy('default')
                .measurement('sample')
                .groupBy('node')
        |window()
                .align()
                .period(5m)
                .every(5m)
        |shift(-5m)
        |mean('metric2').as('metric2')

metric1
        |join(metric2)
           .as('m1', 'm2')
        |log()
```

The resulting stream has fields called 'm1.metric1' and 'm2.metric2'. Furthermore, I can't use a common prefix for both arms of the join - so I can't get 'j.metric1' and 'j.metric2'.

While I understand the need to assign unique names to each of the output fields, my expectation would be the names specified to the as() method are the actual names of the outputs and not prefixes as currently implemented.

jonseymour:
Actually, I understand the reason now -for cases where there is a join between two nodes which have multiple fields. I am going to raise another issue about a suggestion about a better way to handle this,

jonseymour:
This PR contains an implementation of #443. I will re-roll this PR if #443 is accepted first.

nathanielc:
@jonseymour I merged #443 this looks good to go after a rebase.

nathanielc:
Closing in favor of #509

PR #443, opened 2016-04-09
Add a validate phase that allows more validation at definition time

jonseymour:
With this change, more validation can be pushed from runtime to definition time.

See the current implementation of #431 for an example of how it is used.

### DIFF
--- a/pipeline/node.go
+++ b/pipeline/node.go
@@ -61,6 +61,9 @@ type Node interface {
 	// The type of output the node provides.
 	Provides() EdgeType
 
+	// Check that the definition of the node is consistent
+	validate() error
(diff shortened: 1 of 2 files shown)


nathanielc:
@jonseymour I like this change. The only part that I don't like is that now that all specific `Validate` implementation have to mark them selves as `tick:ignore` or they will show up in the public API docs, and be callable from TICKscript. Want do you think about making this validate method private? It would still exist and provide utility but isn't necessary outside the package since its called on `CreatePipeline` for you.

jonseymour:
I'll have to rebase again tomorrow - dead tired now.

Issue #435, opened 2016-04-08
[feature request] support a paste operator in join.as() syntax


jonseymour:
As it currently stands, the field names in output stream if a Kapacitor join node are assigned prefixes of the form \<aliasN\>. where \<aliasN> is the alias assigned to the Nth input stream by the as() property method of the join node.

So, for example:

```
var metric1 = stream
        |from()
                .database('sampledb')
                .retentionPolicy('default')
                .measurement('sample')
                .groupBy('node')
        |window()
                .align()
                .period(5m)
                .every(5m)
        |shift(-5m)
        |mean('metric1').as('metric1')


var metric2 = stream
        |from()
                .database('sampledb')
                .retentionPolicy('default')
                .measurement('sample')
                .groupBy('node')
        |window()
                .align()
                .period(5m)
                .every(5m)
        |shift(-5m)
        |mean('metric2').as('metric2')

metric1
        |join(metric2)
           .as('m1', 'm2')
        |log()
```

will result in a stream with fields called 'm1.metric1' and 'm2.metric2'

It would be nice if the mandatory period in the resulting name could be avoided. 

The suggestion to avoid this is to support use of a trailing paste operator in the arguments to as()

So:

```
metric1
        |join(metric2)
           .as('s_#', 't_#')
```

would result in a stream with fields called s_metric1 and t_metric2 and

and:

```
metric1
        |join(metric2)
           .as('#', '#')
```

would result in a stream with fields called: metric1 and metric2.

'#' is chosen to mimic similar behaviour in the C pre-processor.

Note: A specification of as('foo##', 'foo##') would result in names called 'foo#metric1', 'foo#metric2'.

A specification of as('foo#bar', 'foo#bar') would result in names called foo#bar.metric1 and foo#bar.metric2.

jonseymour:
An assertNotNill('metric1', 'metric4', ... etc) node might be useful for scripts that need to assert they are written with an assumption that the specified fields never contain nil

nathanielc:
@jonseymour  I like the suggestion but the names of the joining nodes are not always available. If they were I would make 'as' optional. 

Example:

```javascript
var metric1 = stream
        |from()
                .database('sampledb')
                .retentionPolicy('default')
                .measurement('sample')
                .groupBy('node')
        |window()
                .align()
                .period(5m)
                .every(5m)
        |shift(-5m)
        |mean('metric1').as('metric1')

var metric2 = stream
        |from()
                .database('sampledb')
[...]

PR #444, opened 2016-04-09
Add support for a paste operator in join.as() prefix specifications.

jonseymour:
Per the discussion in #435, the join.as() method now:

- is optional, if the field names of all input streams are unique
- supports the use of a trailing '#' paste operator in prefix specifications.
- supports duplicate prefixes, provided that the resulting field names are unique

An output column derived from field k of input stream i is now named as follows:

```
if the ith prefix is empty or not specified:
    k
if the ith prefix contains a trailing '#'
   prefix[i][0:len(len(prefix[i])-1]+k
otherwise
  prefix[i]+"."+k
```
An error is detected at runtime if the output of a join node constructed using these rules would result in points with duplicate field names.

All existing, valid Tick scripts should operate per the current behaviour. Some Tick scripts that were invalid are now valid.

### DIFF
--- a/join.go
+++ b/join.go
@@ -404,6 +404,8 @@ type joinset struct {
 	tolerance time.Duration
 	values    []models.PointInterface
 
+	claims map[string]int
+
 	expected int
 	size     int
 	finished int
@@ -434,6 +436,7 @@ func newJoinset(
 		time:      time,
 		tolerance: tolerance,
 		logger:    l,
+		claims:    make(map[string]int),
 	}
 }
 
@@ -580,8 +583,33 @@ func (js *joinset) JoinIntoBatch() (models.Batch, bool) {
 	return newBatch, true
 }
 
+// outName returns the fully qualified name for field k, in stream joined stream i.
+//
+// If the specified prefix is empty, the output name is k.
+//
+// If the specified prefix contains a trailing #, the output name is the prefix
+// without the trailing # concatenated with k
+//
+// Otherwise, the output name is prefix+"."+k
+//
+// A runtime check is performed to guarantee that all output points have unique field
+// names.
 func (js *joinset) outName(i int, k string) string {
-	return js.prefixes[i] + "." + k
+	prefix := js.prefixes[i]
+	if len(prefix) == 0 {
+		prefix = ""
+	} else if prefix[len(prefix)-1] == '#' {
+		prefix = prefix[0 : len(prefix)-1]
+	} else {
+		prefix = prefix + "."
+	}
+	n := prefix + k
+	if claim, ok := js.claims[n]; ok && claim != i {
+		panic(fmt.Errorf("field %s of input %d conflicts with input %d", k, i, claim))
+	} else if !ok {
+		js.claims[n] = i
+	}
+	return n
 }
 
 type durationVar struct {
--- a/join.go
+++ b/join.go
@@ -467,11 +467,11 @@ func (js *joinset) JoinIntoPoint() (models.Point, bool) {
 			switch js.fill {
 			case influxql.NullFill:
 				for k := range js.First().PointFields() {
-					fields[js.prefixes[i]+"."+k] = nil
+					fields[js.outName(i, k)] = nil
 				}
 			case influxql.NumberFill:
 				for k := range js.First().PointFields() {
-					fields[js.prefixes[i]+"."+k] = js.fillValue
+					fields[js.outName(i, k)] = js.fillValue
 				}
 			default:
 				// inner join no valid point possible
@@ -479,7 +479,7 @@ func (js *joinset) JoinIntoPoint() (models.Point, bool) {
 			}
 		} else {
 			for k, v := range p.PointFields() {
-				fields[js.prefixes[i]+"."+k] = v
+				fields[js.outName(i, k)] = v
 			}
 		}
 	}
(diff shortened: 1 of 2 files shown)


jonseymour:
Note this PR currently includes implementations of #443 and #331, assuming that these PRs will be reviewed and merged in that order. I will re-roll this PR once the other PRs are merged or otherwise resolved.

jonseymour:
The tests [here](https://github.com/wildducktheories/influxdata-testcase/tree/master/tests/issue-435) demonstrate the use of the new as() implementation.

All of these are now valid constructions:

```
metric1
	|join(metric2)
	|log()

metric1
	|join(metric2).as('out1', 'out2')
	|log()

metric1
	|join(metric2).as('out1')
	|log()

metric1
	|join(metric2).as('#')
	|log()

metric1
	|join(metric2).as('#', '#')
	|log()

metric2
	|join(metric2a).as('i1_#', 'i2_#')
	|log()
```

The following is an example of a join that causes a runtime error because of duplicate names in the output stream.

```
var metric2 = stream
	|from()
		.database('sampledb')
		.retentionPolicy('default')
		.measurement('sample')
		.groupBy('node')
	|window()
		.align()
		.period(5m)
		.every(5m)
	|shift(-5m)
	|mean('metric2').as('metric2')

var metric2a = stream
	|from()
		.database('sampledb')
		.retentionPolicy('default')
		.measurement('sample')
		.groupBy('node')
	|window()
		.align()
		.period(5m)
		.every(5m)
	|shift(-5m)
	|mean('metric2').as('metric2')

metric2
	|join(metric2a)
	|log()
```

jonseymour:
I wonder if there is really a good reason to prevent "." appearing in a specified prefix?

nathanielc:
> I wonder if there is really a good reason to prevent "." appearing in a specified prefix?

prefix1: 'abc'
prefix2: 'ab'

field1 'd'
field2 'cd'

resulting fields using `.` delimiter  'abc.d' and 'ab.cd'  but without a delimiter you get a conflict on 'abcd'.

Any delimiter is needed to resolve issues where one of the prefixes is a prefix of the other. As long as you disallow the delimiter in the prefix conflicts are impossible. 

This case is obviously not likely but I'd rather just prevent it.

jonseymour:
@nathanielc I've rebased this change onto master. Ready for your review.

join.go
```diff
@@ -580,6 +591,35 @@ func (js *joinset) JoinIntoBatch() (models.Batch, bool) {
...
+	if len(prefix) == 0 {
+		prefix = ""
+	} else if prefix[len(prefix)-1] == '#' {
+		prefix = prefix[0 : len(prefix)-1]
+	} else {
+		prefix = prefix + "."
+	}
+	n := prefix + k
+	if claim, ok := js.claims[n]; ok && claim != i {
+		panic(fmt.Errorf("field %s of input %d conflicts with input %d", k, i, claim))
```
nathanielc: Could we remove the panic here and return an error instead?

The calling functions `JoinInto*` can then log the error and `return models.Point/Batch{},false` the join will be skipped.

pipeline/join.go
```diff
@@ -90,11 +91,23 @@ func newJoinNode(e EdgeType, parents []Node) *JoinNode {
...
+// If a prefix is not specified or is empty, then field names from parent nodes
+// will be copied without attaching a prefix.
+//
+// If a prefix contains a trailing '#' character, then the prefix
+// upto, but not including the trailing '#' character is prepended to the field name
+// from the parent node without adding the '.' character after the prefix.
+//
+// It is the callers responsibility to ensure that when these rules are applied
+// the collection of output field names does not contain any duplicate names. Failure
+// to ensure this will result in a runtime error.
```
nathanielc: Should the entire task be stopped because of a name conflict? I know that in the `EvalNode` for example if invalid data arrives an error is logged and the task continues to process. I see two extremes

1. A single point can stop an entire task
2. A single point can be dropped form the data set.

Either extreme could be detrimental based on the use case but I think #1 is the worse of the two since is means possibly lots of data is dropped, while the task is off.

nathanielc:
@jonseymour I have been thinking about the user experience here and would like some feedback:

First, to be clear the entire purpose of the `#` char is to remove the `.` from the prefix name? And as a result the onus is now on the user to choose unique prefixes? 

To keep it simple I see two ways of going about the `.as` property

1. Leave it how it is currently, which prevents the user from creating name conflicts, but gives less control over the names.
2. Give user full control and explicitly check for name conflicts and handle them.

There may be some middle ground but its not clear to me what that would be. 

I think if the ability to have full control over the field name is something we want then we go with #2. Based on your experience and others I believe we do need the control.

Also by adding runtime checking for name conflicts we can remove the requirement to have an `as` property. The default will be to assume that the field names are unique, if that assumption fails an error occurs and is handled appropriately.

Thoughts?

pipeline/join.go
jonseymour: I believe that the task should be stopped since data it is likely that if a set of points fails the uniqueness test then all subsequent sets of points will also fail the test in which case joined data will be silently dropped from the output stream.

join.go
jonseymour: I made use of panic because it complicates every caller of outName which must now check for an error I'll push another commit which uses error.

jonseymour:
Yes, I think we do need to give the user more control and one of the implications of this change is that .as() becomes optional unless there is a conflict to resolve.

nathanielc:
Closing in favor of a simpler solution #698 

The `.as` property is still required but that can be changed later as well.

Issue #433, opened 2016-04-08
improving comprehensibility of kapacitor documentation


jonseymour:
As they say: there are 2 hard problems in computer science: cache invalidation, naming things and off by one errors.

The Kapacitor documentation currently documents 4 node types:

- Stream
- Batch
- SourceStream
- SourceBatch

Unfortunately, the correspondence between the documentation and keywords used in Tick scripts is as follows:

- Stream <=> from
- Batch <=> query
- SourceStream <=> stream
- SourceBatch <=> batch

It would be helpful if the documentation could be renamed so that the correspondence was instead:

- From <=> from
- Query <=> query
- Stream <=> stream
- Batch <=> batch

For completeness, it might also be worth refactoring the source code so that the source code names matched the documentation names.

I think this would make the documentation much easier to comprehend.

jonseymour:
Closed in favour of https://github.com/influxdata/docs.influxdata.com/issues/384

jonseymour:
Reopened since this is auto-generated documentation.

PR #445, opened 2016-04-09
Refactor type names to improve Kapacitor documentation.

jonseymour:
As discussed in #433 the type names used in the online Kapacitor documentation are unhelpful since the correspondence between Tick script keywords and type names is not as clear as it could be.

Since this documentation is auto-generated we really need to fix the type names in the source code.

Since this would be tedious and error prone if done manually, the refactoring is done automatically
with a bash script which:

- applies one refactoring at a time with grep and sed
- compiles the resulting code to check for compile errors
- commits the result to git
- finally, removes itself from the code base.

Once this is ready to be merged, the first commit in the series should be rebased onto the current origin/master and refactor-doc.sh executed which will produce a refactoring that exactly matches the current origin/master.

The other commits in this series are provided so that the refactorings the script is doing can be reviewed.

/cc @nathanielc


pipeline/batch.go
```diff
@@ -129,8 +129,8 @@ type BatchNode struct {
 	Cluster string
 }
 
-func newBatchNode() *BatchNode {
-	b := &BatchNode{
+func newQueryNode() *QueryNode {
+	b := &QueryNode{
 		chainnode: newBasicChainNode("batch", BatchEdge, BatchEdge),
```
nathanielc: Change "batch" to "query"

pipeline/stream.go
```diff
@@ -104,8 +104,8 @@ type StreamNode struct {
 	Truncate time.Duration
 }
 
-func newStreamNode() *StreamNode {
-	return &StreamNode{
+func newFromNode() *FromNode {
+	return &FromNode{
 		chainnode: newBasicChainNode("stream", StreamEdge, StreamEdge),
```
nathanielc: Change "stream" to "from"

nathanielc:
@jonseymour Thanks! The original names have drifted in meaning, I really like these new names simple and clear.

There are a few spots that should be changed to reflect this name change.

* https://github.com/influxdata/kapacitor/blob/master/task_master.go#L509 change srcstream0 to stream0. 
* Two other places in the diff that I commented.

pipeline/batch.go
```diff
@@ -21,12 +21,12 @@ import (
...
+type BatchNode struct {
 	node
 }
 
-func newSourceBatchNode() *SourceBatchNode {
-	return &SourceBatchNode{
+func newBatchNode() *BatchNode {
+	return &BatchNode{
 		node: node{
 			desc:     "srcbatch",
```
nathanielc: This should also change to "batch"

pipeline/stream.go
```diff
@@ -6,17 +6,17 @@ import (
...
+type StreamNode struct {
 	node
 }
 
-func newSourceStreamNode() *SourceStreamNode {
-	return &SourceStreamNode{
+func newStreamNode() *StreamNode {
+	return &StreamNode{
 		node: node{
 			desc:     "srcstream",
```
nathanielc: This should change to "stream"

nathanielc:
@jonseymour I would like to get this change in before too much changes and makes this PR invalid. Do you think you will have time to make a few small changes and rebase over the next few days?

nathanielc:
closing in favor of #476

jonseymour:
@nathanielc apologies for not getting back to you earlier about this change - I have deeply absorbed in some other tasks that reduced my time for working on these changes. Thanks for making the changes in #476

nathanielc:
@jonseymour No problem, Kapacitor is cutting a release and I wanted these good changes to make it in, so I finished up. Thanks for getting it started!

PR #447, opened 2016-04-09
[WIP] add support for recording tags to kapacitor

jonseymour:
This PR is a work in progress. It is being published now to allow feedback to be gathered as I am working on it. Expect many re-rolls and rebasing until it becomes feature complete.

- [x] sketch proposed changes to kapacitor command
- [x] extend types in cli package to support transport of tag related data
- [x] add routes to server to support tags
- [ ] provide Replay service with a TagStore service
- [ ] implement tag routes
- [ ] implement a TagStore backed by Bolt DB

**Open questions**
- should the creation of tags that look exactly like a recording id be restricted? Leaning towards yes.
- should re-use of an existing tag for a different recording be prevented?  Leaning towards no.

### DIFF
--- a/client/v1/client.go
+++ b/client/v1/client.go
@@ -129,6 +129,13 @@ type Recording struct {
 	Size    int64
 	Created time.Time
 	Error   string
+	Tags    []string
+}
+
+// Tag is a symbolic name for a recording identifier.
+type Tag struct {
+	Tag string
+	ID  string
 }
 
 // Set of recordings sorted by created date.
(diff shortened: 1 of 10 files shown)


nathanielc:
>should the creation of tags that look exactly like a recording id be restricted? Leaning towards yes.

I agree, lets restrict them.

>should re-use of an existing tag for a different recording be prevented? Leaning towards no.

No, seems like a reasonable update action to change which recording a tag points to.

nathanielc:
@jonseymour With version 0.13 its now possible to supply your own names for recordings etc. As such tags are not as important. They could still be useful but since the storage format has significantly changed since this PR was started, a fresh start will be needed.

Issue #448, opened 2016-04-09
failure running TestServer_UDFStreamAgents on OSX


jonseymour:
I received this when running the Kapacitor tests on OSX. Presumably this is because my local python installation is missing the protobuf dependency. 

It would be helpful if the required dependencies could be documented in the comments for the test method.

Probably ideal, would be to run the test in a dockerised container to void the need to install anything other than docker on the client.

```
--- FAIL: TestServer_UDFStreamAgents (32.37s)
	server_test.go:917: open service *udf.Service: failed to load process info for "movingAvg": write error: write |1: broken pipe
2016/04/10 08:40:02 Server listening on /var/folders/q2/_r9_5f1112n7y3704hg3cl2m0000gn/T/kapacitor_server_test911469067/echo.go.sock
2016/04/10 08:40:03 Starting agent for connection 0
2016/04/10 08:40:03 Agent for connection 0 finished
2016/04/10 08:40:03 Starting agent for connection 1
2016/04/10 08:40:03 Agent for connection 1 finished
Traceback (most recent call last):
  File "/Users/jonseymour/ninja/go/src/github.com/influxdata/kapacitor/udf/agent/examples/echo/echo.py", line 3, in <module>
    from agent import Agent, Handler, Server
  File "/Users/jonseymour/ninja/go/src/github.com/influxdata/kapacitor/udf/agent/py/agent.py", line 7, in <module>
    import udf_pb2
  File "/Users/jonseymour/ninja/go/src/github.com/influxdata/kapacitor/udf/agent/py/udf_pb2.py", line 6, in <module>
    from google.protobuf.internal import enum_type_wrapper
```

jonseymour:
I cribbed the dependencies from the Dockerfile:

```
# Install protobuf3
ENV PROTO_VERSION 3.0.0-beta-2
# Download and compile protoc
RUN wget https://github.com/google/protobuf/archive/v${PROTO_VERSION}.tar.gz && \
    tar xf v${PROTO_VERSION}.tar.gz && \
    rm -f v${PROTO_VERSION}.tar.gz && \
    cd protobuf-${PROTO_VERSION} && \
    ./autogen.sh && \
    ./configure --prefix=/usr && \
    make -j $(nproc) && \
    make check && \
    make install

# Install Python Protobuf3
RUN cd protobuf-${PROTO_VERSION}/python && \
    python setup.py install;


```

jonseymour:
Closed via #449

PR #449, opened 2016-04-10
Factor out the protobuf dependency into a shell script.

jonseymour:
The kapacitor unit tests don't run locally unless the protobuf
dependencies are installed locally.

Since the required glue is somewhat non-trivial, factor the glue
out from the Dockerfiles and change the docker files to use the same
glue.

Add a comment to the test to indicate how to fix the missing dependencies
issue.

### DIFF
--- a/cmd/kapacitord/run/server_test.go
+++ b/cmd/kapacitord/run/server_test.go
@@ -838,6 +838,8 @@ func TestServer_RecordReplayBatch(t *testing.T) {
 	}
 }
 
+// If this test fails due to missing python dependencies, run 'INSTALL_PREFIX=/usr/local ./install-deps.sh' from the root directory of the
+// kapacitor project.
 func TestServer_UDFStreamAgents(t *testing.T) {
 	dir, err := os.Getwd()
 	if err != nil {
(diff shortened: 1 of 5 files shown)


jonseymour:
This addresses #448.

cmd/kapacitord/run/server_test.go
```diff
@@ -838,6 +838,8 @@ func TestServer_RecordReplayBatch(t *testing.T) {
 	}
 }
 
+// If this test fails due to missing python dependencies, run 'INSTALL_PREFIX=/usr/local ./install-deps.sh' from the root directory of the
+// kapacitor project.
```
nathanielc: Same applies to the TestServer_UDFBatchAgent below

install-deps.sh
```diff
@@ -0,0 +1,31 @@
+#!/bin/bash
+
+# this script installs the python dependencies required to
```
nathanielc: It installs both the `protoc` compiler and the python plugin and libs.

nathanielc:
@jonseymour Thanks for the simplification...

nathanielc:
@jonseymour I'll merge this and update the comment, Thanks for the PR.

PR #453, opened 2016-04-11
Fix record query as stream with group by to save data in time order

### LINKED ISSUE #423: groupBy node doesn't have expected behaviour
I defined this tick script:

```
var aggregate = stream
	|from()
		.database('sampledb')
		.retentionPolicy('default')
		.measurement('sample')
	|window()
		.align()
		.period(5m)
		.every(5m)
	|groupBy('node')
	|mean('metric1').as('metric1')
	|influxDBOut()
		.database('sampledb')
		.measurement('actual')
```

My expectation is that when this script ran, it would produce the equivalent of:

```
select mean(metric1) as metric1 
into actual 
from sample 
where time >= '2016-03-08 05:00:00' and time < '2016-03-08 08:00:00' 
group by time(5m), *
```

However, what actually happened is that only one of the "node" groups was written into "actual" - the other groups were discarded.

Is my expectation of what the groupBy node is meant to do incorrect or is this a bug?

I have a Vagrantfile project that includes test data and reproduces the problem available here:    https://github.com/wildducktheories/influxdata-testcase

I was using kapacitor 0.12 and influxdb 0.11

nathanielc:
Fixes #423


nathanielc:
@jonseymour When you get a change can you confirm that this change fixes your test case? Thanks

PR #461, opened 2016-04-13
Make Alerta event property configurable

satterly:
Make the Alerta event property configurable to ensure [de-duplication](http://docs.alerta.io/en/latest/server.html#de-duplication) and [event correlation](http://docs.alerta.io/en/latest/server.html#simple-correlation) work as expected.

- [x] CHANGELOG.md updated
- [x] Rebased/mergable
- [x] Tests pass
- [x] Sign [CLA](http://influxdb.com/community/cla.html) (if not already signed)

### DIFF
--- a/alert.go
+++ b/alert.go
@@ -232,6 +232,10 @@ func newAlertNode(et *ExecutingTask, n *pipeline.AlertNode, l *log.Logger) (an *
 		if err != nil {
 			return nil, err
 		}
+		evtmpl, err := text.New("event").Parse(alerta.Event)
+		if err != nil {
+			return nil, err
+		}
 		etmpl, err := text.New("environment").Parse(alerta.Environment)
 		if err != nil {
 			return nil, err
@@ -247,6 +251,7 @@ func newAlertNode(et *ExecutingTask, n *pipeline.AlertNode, l *log.Logger) (an *
 		ai := alertaHandler{
 			AlertaHandler:   alerta,
 			resourceTmpl:    rtmpl,
+			eventTmpl:       evtmpl,
 			environmentTmpl: etmpl,
 			groupTmpl:       gtmpl,
 			valueTmpl:       vtmpl,
@@ -873,6 +878,7 @@ type alertaHandler struct {
 	*pipeline.AlertaHandler
 
 	resourceTmpl    *text.Template
+	eventTmpl       *text.Template
 	environmentTmpl *text.Template
 	valueTmpl       *text.Template
 	groupTmpl       *text.Template
@@ -907,6 +913,14 @@ func (a *AlertNode) handleAlerta(alerta alertaHandler, ad *AlertData) {
 	resource := buf.String()
 	buf.Reset()
 
+	err = alerta.eventTmpl.Execute(&buf, ad.info)
+	if err != nil {
+		a.logger.Printf("E! failed to evaluate Alerta Event template %s", alerta.Event)
+		return
+	}
+	event := buf.String()
+	buf.Reset()
+
 	err = alerta.environmentTmpl.Execute(&buf, ad.info)
 	if err != nil {
 		a.logger.Printf("E! failed to evaluate Alerta Environment template %s", alerta.Environment)
@@ -938,7 +952,7 @@ func (a *AlertNode) handleAlerta(alerta alertaHandler, ad *AlertData) {
 	err = a.et.tm.AlertaService.Alert(
 		alerta.Token,
 		resource,
-		ad.ID,
+		event,
 		environment,
 		severity,
 		group,
--- a/alert.go
+++ b/alert.go
@@ -913,7 +913,15 @@ func (a *AlertNode) handleAlerta(alerta alertaHandler, ad *AlertData) {
 	resource := buf.String()
 	buf.Reset()
 
-	err = alerta.eventTmpl.Execute(&buf, ad.info)
+	type eventData struct {
+		idInfo
+		ID string
+	}
+	data := eventData{
+		idInfo: ad.info.messageInfo.idInfo,
+		ID:     ad.ID,
+	}
+	err = alerta.eventTmpl.Execute(&buf, data)
 	if err != nil {
 		a.logger.Printf("E! failed to evaluate Alerta Event template %s", alerta.Event)
 		return
--- a/integrations/streamer_test.go
+++ b/integrations/streamer_test.go
@@ -2707,6 +2707,9 @@ func TestStream_AlertAlerta(t *testing.T) {
 			if exp := "cpu"; pd.Resource != exp {
 				t.Errorf("unexpected resource got %s exp %s", pd.Resource, exp)
 			}
+			if exp := "serverA"; pd.Event != exp {
+				t.Errorf("unexpected event got %s exp %s", pd.Event, exp)
+			}
 			if exp := "production"; pd.Environment != exp {
 				t.Errorf("unexpected environment got %s exp %s", pd.Environment, exp)
 			}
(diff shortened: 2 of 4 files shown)


nathanielc:
So I am a little confused. Kapacitor uses its ID field to correlate different alerts. Whats the value in allowing the `event` field to be defined separately from the ID? 

As I understand the Alerta docs if I send these requests to Alerta:

```
event=kapacitor/host/example.com,resource=ping,severity=warning
event=kapacitor/host/example.com,resource=ping,severity=warning
event=kapacitor/host/example.com,resource=ping,severity=critical
event=kapacitor/host/example.com,resource=ping,severity=ok
```

The second warnings is de-duped and dropped.
Then the next critical request correlates with the existing warning and updates it to a critical.
Finally the last request clears the alert with OK.

If either event of resource were different than they would be separate alerts entirely.

Correct?

satterly:
You're understanding of how Alerta de-duplicates and correlates is spot on and yes, the integration between Kapacitor and Alerta could be made to work this way. However, it would be quite limiting.

From your example above, it would make much more sense for the `resource` to be "kapacitor/host/example.com" and the `event` to be "ping" but that isn't possible because `event` is hardcoded to the Kapacitor ID.

Think of the combination of `event-resource` as the equivalent of the Kapitor ID so for the integration with Alerta to be fully realised you need to split the Kapacitor ID somewhere and assign one part to `event` and the other part to `resource`. That it is possible to assign values to the `resource` field but not possible to assign values to the `event` field seems wrong because they are equally as important to Alerta.

It probably makes sense to default the `event` to Kapacitor ID instead of `TaskName` though.

nathanielc:
@satterly Currently the `event` is the `ID` property and `resource` is configured independently.   So both are configurable like so:

```
|alert()
    .id('custom event {{ index .Tags "key"}}')
    .resource('custom resource {{ index .Tags "someotherkey"}}')
    .alerta()
```

Would that not work? Is it just a mater of better documentation?

satterly:
Wouldn't that mean that alert generation in Kapacitor wouldn't work properly because someone wanting to integrate with Alerta would only want to put event name in the Kap ID so it would end up only generating one alert per alert definition for many different resources.

That is, the Kap ID needs to make reference to a resource doesn't it?

nathanielc:
@satterly You are right. It would break Kapacitor's concept of an ID if they were also using another alerting system in addition to Alerta. Probably not a common use case but it would definitely be an obscure/confusing behavior.

OK, so lets make this change as it make sense now. Just a few things...

1. Like you said default event to the ID 
2. Provide only the `idInfo` struct when executing the `event` template so the user cannot create an `event` with data that changes depending on the state of the alert.

Thanks!

satterly:
@nathanielc will do. And thanks for the feedback. The discussion was very useful in clarifying the approach.

.idea/compiler.xml
```diff
@@ -0,0 +1,22 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
nathanielc: Can you remove these .idea files?

pipeline/alert.go
```diff
@@ -768,6 +770,11 @@ type AlertaHandler struct {
 	// Default: {{ .Name }}
 	Resource string
 
+	// Alerta event.
+	// Can be a template and has access to the same data as the AlertNode.Details property.
```
nathanielc: Can you update this comment since it uses on the Info struct and not the Details?

integrations/streamer_test.go
```diff
@@ -2534,7 +2534,10 @@ func TestStream_AlertAlerta(t *testing.T) {
 			if exp := "cpu"; pd.Resource != exp {
 				t.Errorf("unexpected resource got %s exp %s", pd.Resource, exp)
 			}
-			if exp := "production"; pd.Environment != exp {
+			if exp := "TestStream_Alert"; pd.Event != exp {
```
nathanielc: based on your event template `event: {{ .TaskName }}` the expected value should be `exp := "event: TestStream_Alert"`

nathanielc:
Here is the build failure output:

````
--- FAIL: TestStream_AlertAlerta (0.01s)
	streamer_test.go:3536: digraph TestStream_Alert {
		stream0 -> from1;
		from1 -> window2;
		window2 -> count3;
		count3 -> alert4;
		}
	streamer_test.go:2532: unexpected url got /alert?api-key=anothertesttoken exp /alert?api-key=testtoken1234567
	streamer_test.go:2535: unexpected resource got serverA exp cpu
	streamer_test.go:2538: unexpected event got event: TestStream_Alert exp TestStream_Alert
	streamer_test.go:2541: unexpected environment got Development exp Production
	streamer_test.go:2544: unexpected group got serverA exp host=serverA,
	streamer_test.go:2547: unexpected value got 10 exp 
	streamer_test.go:2550: unexpected service got [serviceA serviceB] exp [cpu]
	streamer_test.go:2553: unexpected origin got override exp Kapacitor
	streamer_test.go:2582: unexpected event got event: TestStream_Alert exp TestStream_Alert
	streamer_test.go:2635: unexpected requestCount got 1 exp 2
```

Have you been able to get the tests to run locally? You can use this comand to run only the Alerta tests.
```sh
go test -run TestStream_AlertAlerta -v ./integrations
```

satterly:
Thanks for the suggestions. Will give running just the Alerta tests locally
a go.

--Nick.

On Fri, Apr 22, 2016 at 4:04 PM, Nathaniel Cook <notifications@github.com>
wrote:

> Here is the build failure output:
>
> --- FAIL: TestStream_AlertAlerta (0.01s)
>     streamer_test.go:3536: digraph TestStream_Alert {
>         stream0 -> from1;
>         from1 -> window2;
>         window2 -> count3;
>         count3 -> alert4;
>         }
>     streamer_test.go:2532: unexpected url got /alert?api-key=anothertesttoken exp /alert?api-key=testtoken1234567
>     streamer_test.go:2535: unexpected resource got serverA exp cpu
>     streamer_test.go:2538: unexpected event got event: TestStream_Alert exp TestStream_Alert
>     streamer_test.go:2541: unexpected environment got Development exp Production
>     streamer_test.go:2544: unexpected group got serverA exp host=serverA,
>     streamer_test.go:2547: unexpected value got 10 exp
>     streamer_test.go:2550: unexpected service got [serviceA serviceB] exp [cpu]
>     streamer_test.go:2553: unexpected origin got override exp Kapacitor
>     streamer_test.go:2582: unexpected event got event: TestStream_Alert exp TestStream_Alert
>     streamer_test.go:2635: unexpected requestCount got 1 exp 2
>
> Have you been able to get the tests to run locally? You can use this
> comand to run only the Alerta tests.
>
> go test -run TestStream_AlertAlerta -v ./integrations
>
> —
> <https://github.com/influxdata/kapacitor/pull/461#issuecomment-213465513>
>
> ------------------------------
> This e-mail and all attachments are confidential and may also be
> privileged. If you are not the named recipient, please notify the sender
> and delete the e-mail and all attachments immediately. Do not disclose the
> contents to another person. You may not use the information for any
> purpose, or store, or copy, it in any way.  Guardian News & Media Limited
> is not liable for any computer viruses or other material transmitted with
> or as part of this e-mail. You should employ virus checking software.
>
> Guardian News & Media Limited is a member of Guardian Media Group plc. Registered
> Office: PO Box 68164, Kings Place, 90 York Way, London, N1P 2AP.  Registered
> in England Number 908396
>
>
>


-- 
gpg: using PGP trust model
pub   4096R/1EE38BD9 2013-01-06 [expires: 2018-01-06]
      Key fingerprint = 3EE9 550D D9D8 DB65 58C2  B58D CE78 EC6C 1EE3 8BD9
uid                  Nicholas Satterly (Debian Key) <nfsatterly@gmail.com>
sub   4096R/23804EE9 2013-01-06 [expires: 2018-01-06]

.idea/compiler.xml
satterly: Submitted a PR to exclude IntelliJ files #488

satterly:
When I run the tests I get ...
```
E! failed to evaluate Alerta Event template {{ .ID }}
```
This would be expected because we only provide the `idInfo` struct to the event template. However the `idInfo` struct doesn't seem to have access to `.ID` and confusingly we made it the default.

alert.go
```diff
@@ -907,6 +913,14 @@ func (a *AlertNode) handleAlerta(alerta alertaHandler, ad *AlertData) {
 	resource := buf.String()
 	buf.Reset()
 
+	err = alerta.eventTmpl.Execute(&buf, ad.info)
```
nathanielc: Right here you can create a new type like:

```go
type eventData struct {
      idInfo
      ID string
}
data := eventData{
     idInfo: ad.info.messageInfo.idInfo,
     ID: ad.ID,
}
err = alerta.eventTmpl.Execute(&buf, data)
```

This way on the ID Info data can be used in addition to the ID itself.

satterly: Thanks. Updated.

nathanielc:
@satterly Looks great, Thanks!

PR #462, opened 2016-04-14
Fixing misspells

yosiat:
Hi,

I found out the "Go Report Card" and found out some misspells (some that I wrote :( )

Here is the report card:
https://goreportcard.com/report/github.com/influxdata/kapacitor

And I used this tool for fixing the misspells:
https://github.com/client9/misspell


nathanielc:
Thanks!

PR #464, opened 2016-04-14
[WIP] Build script updates

rossmcdonald:
Major improvements to `build.py`:
- Now compatible with Python 3
- Now uses Python argparse for cleaner handling of arguments
- Now uses Python logging
- Added function documentation
- Removed a few unneeded functions
- Updated nightly logic to increment minor version
- Added support for building from specific branch or commit
- Added a --release flag, default package output will now contain the commit hash in the version number
- Static builds are now listed as an architecture (and can be built with `--arch=all`)

### DIFF
--- a/build.py
+++ b/build.py
@@ -4,13 +4,13 @@ import sys
 import os
 import subprocess
 import time
-import datetime
+from datetime import datetime
 import shutil
 import tempfile
 import hashlib
 import re
-
-debug = False
+import logging
+import argparse
 
 ################
 #### Kapacitor Variables
@@ -45,9 +45,9 @@ VENDOR = "InfluxData"
 DESCRIPTION = "Time series data processing engine"
 
 # SCRIPT START
-go_vet_command = ["go", "tool", "vet", "-composites=false"]
+go_vet_command = "go tool vet -composites=false"
 prereqs = [ 'git', 'go' ]
-optional_prereqs = [ 'fpm', 'rpmbuild' ]
+optional_prereqs = [ 'fpm', 'rpmbuild', 'gpg' ]
 
 fpm_common_args = "-f -s dir --log error \
  --vendor {} \
@@ -85,7 +85,7 @@ targets = {
 
 supported_builds = {
     'darwin': [ "amd64", "i386" ],
-    'linux': [ "amd64", "i386", "armhf", "arm64", "armel" ],
+    'linux': [ "amd64", "i386", "armhf", "arm64", "armel", "static_i386", "static_amd64" ],
     'windows': [ "amd64", "i386" ]
 }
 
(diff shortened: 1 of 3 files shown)


build.py
```diff
@@ -100,37 +100,49 @@
...
+def print_banner():
+    logging.info("""
+                               _             
+  /\\ /\\__ _ _ __   __ _  ___(_) |_ ___  _ __ 
+ / //_/ _` | '_ \\ / _` |/ __| | __/ _ \\| '__|
+/ __ \\ (_| | |_) | (_| | (__| | || (_) | |   
+\\/  \\/\\__,_| .__/ \\__,_|\\___|_|\\__\\___/|_|   
+           |_|                               
+  Build Script
+""")
```
nathanielc: Can we make this the same as https://github.com/influxdata/kapacitor/blob/master/cmd/kapacitord/run/command.go#L19

build.py
```diff
@@ -319,220 +358,196 @@ def upload_packages(packages, bucket_name=None, nightly=False):
...
                 n = k.set_contents_from_filename(p, replace=True)
             else:
                 n = k.set_contents_from_filename(p, replace=False)
             k.make_public()
         else:
-            print "!! Not uploading package {}, as it already exists.".format(p)
-    print ""
-    return 0
-
-def go_list(vendor=False, relative=False):
```
nathanielc: Why are we removing `go_list`?

rossmcdonald: Sorry, that was a casualty of the initial merge of the changes. It's fixed now, though.

build.py
rossmcdonald: Yep! Changed.

nathanielc:
@rossmcdonald Overall this looks great! I am going to play around with it locally for a bit.

nathanielc:
@rossmcdonald  What is the motivation for?

> default package output will now contain the commit hash in the version number

nathanielc:
@rossmcdonald Is it because the `--release` flag removes the commit hash and so builds not marked as release are easily identifiable?

nathanielc:
So here are some example versions generated from the builds.

non-release version:  0.12.0-a10b80f-0
release version:  0.12.0-1
nightly version: 0.13.0~n201604151656

If my understanding is correct than the version order should be:

non-release < release < nightly

This works for rpm but the Debian version comparison fails for non-release < release:

```
$ dpkg --compare-versions 0.12.0-a10b80f-0 lt 0.12.0-1 && echo true || echo false
false
```

But if we switch to using `~` like we do  for the  nightlies is works:

```
$ dpkg --compare-versions 0.12.0~a10b80f-0 lt 0.12.0-1 && echo true || echo false
true
```

Also when is a non-release version used as opposed to the nightly builds? I am not quite sure I understand the purpose of the git sha in the version yet...

rossmcdonald:
> Is it because the --release flag removes the commit hash and so builds not marked as release are easily identifiable?

Yes, that's part of it. The idea is that we are going to start uploading all Circle builds to S3 so that developers and testing team can have easy access to packaged builds. I wanted to add a unique identifier for non-release build artifacts so that we didn't have any collisions when uploading and are easy to spot. I also eventually want to automate the entire release process, so this is the first step in that direction.

Regarding the version comparisons, I assumed that the non-release builds will only be used for testing and development purposes (and maybe installed on a one-off basis to verify hotfixes, etc). I think that changing it to fit the nightly naming convention (with a `~`) is a great idea, though. That way anyone will always be able to upgrade from non-release to release to nightly without any issues.

I'm also open to any thoughts or ideas you may have (or if you think the whole idea should be scrapped).

nathanielc:
We will have an archive of all builds, sounds good. +1

 Since CircleCI manages the releases for Kapacitor can you add the `--release` flag to the circle.yml file for the deployment section? Thanks.

 Also when you are ready to start testing automation pieces let me know. Kapacitor is fully automated already but just uses CircleCI so we can't do things like sign the packges etc. Should be simple to migrate to a more trusted system.

PR #467, opened 2016-04-15
Can now require that all points in alert batch match criteria

### LINKED ISSUE #281: How to check last n continue value
want to check cpu usage value for last 3 data sample ponits,  when all of these data points match condition, then tirgger alarm. how to write it ?

var data = stream.from().measurement('cpu')
var data = data.where(lambda: hour("time") >= 10 AND hour("time") <= 17)

data
    .groupBy('host', 'cpu')
    //.window()
    //    .period(20s)
    //    .every(10s)
    .eval(lambda: 100.0 - "usage_idle").as('used')
    .alert()
        .id('{{ .Name}}/{{ index .Tags "host" }}/{{ index .Tags "cpu"}}')
        .message('this is msg')
        .details('this is detail')       
        .warn(lambda: "used" > 80.0)
        //.crit(lambda: "used" > 90.0)
        .log('/tmp/alarms.log')
        //.stateChangesOnly()var data = stream.from().measurement('cpu')
var data = data.where(lambda: hour("time") >= 10 AND hour("time") <= 17)
//var data = data.where(lambda: "cpu" == 'cpu1')
//var data_jq = stream.from().measurement('cpu').where(lambda: "dc" == 'jq')
//var data_njq = stream.from().measurement('cpu').where(lambda: "dc" != 'jq')

nathanielc:
Fixes #281 

Note the `any` method was not added as that is the default behavior.

This is now possible:

```javascript

|alert()
    .all()
    .warn(lamnda: "value" > 5)
    .crit(lamnda: "value" > 10)
```

Given a batch of data with:

```
value 7
value 12
value 4
```

No alert would be triggered

Given a batch of data with:

```
value 7
value 12
value 6
```

A `WARNING` will be triggered

Given a batch of data with:

```
value 13
value 12
value 15
```

A `CRITICAL` will be triggered

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,6 +47,8 @@ For example, let's say we want to store all data that triggered an alert in Infl
 - [#417](https://github.com/influxdata/kapacitor/issues/417): UDFs can be connected over a Unix socket. This enables UDFs from across Docker containers.
 - [#451](https://github.com/influxdata/kapacitor/issues/451): StreamNode supports `|groupBy` and `|where` methods.
 - [#93](https://github.com/influxdata/kapacitor/issues/93): AlertNode now outputs data to child nodes. The output data can have either a tag or field indicating the alert level.
+- [#281](https://github.com/influxdata/kapacitor/issues/281): AlertNode now has an `.all()` property that specifies that all points in a batch must match the criteria in order to trigger an alert.
+
 
 ### Bugfixes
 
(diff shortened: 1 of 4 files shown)


yosiat:
@nathanielc this sounds very interesting.. but just to make sure If have 3 points in batch and only is critical while others are ok - an alert will be dispatched? or all points must be critical?

nathanielc:
@yosiat If you use the `.all()` property on the alert then all must be critical in order to trigger an alert. So in your example no alert will trigger. If you want an alert to trigger don't use `.all()`

I'll update the comments with an example.

yosiat:
@nathanielc If so this is great :+1: 
I have requirement that asks to dispatch the alert only after X times when critical severity.

xaon2017:
but the value when the alarm is restored is the value when the alarm is triggered.

PR #470, opened 2016-04-16
Adding benchmark for stateful expression

yosiat:
Current results

```
PASS
Benchmark_EvalBool_OneOperator_NumberFloat64_NumberFloat64-4                      	 3000000	       543 ns/op	      72 B/op	       5 allocs/op
Benchmark_EvalBool_OneOperator_NumberFloat64_NumberInt64-4                        	 3000000	       554 ns/op	      72 B/op	       5 allocs/op
Benchmark_EvalBool_OneOperator_NumberInt64_NumberInt64-4                          	 3000000	       551 ns/op	      72 B/op	       5 allocs/op
Benchmark_EvalBool_OneOperator_ReferenceNodeFloat64_NumberFloat64-4               	 3000000	       530 ns/op	      64 B/op	       4 allocs/op
Benchmark_EvalBool_OneOperator_ReferenceNodeFloat64_NumberInt64-4                 	 3000000	       522 ns/op	      64 B/op	       4 allocs/op
Benchmark_EvalBool_OneOperator_ReferenceNodeFloat64_ReferenceNodeFloat64-4        	 3000000	       508 ns/op	      49 B/op	       3 allocs/op
Benchmark_EvalBool_OneOperatorWith11ScopeItem_ReferenceNodeFloat64_NumberFloat64-4	 3000000	       534 ns/op	      64 B/op	       4 allocs/op
Benchmark_EvalBool_OneOperatorValueChanges_ReferenceNodeFloat64_NumberFloat64-4   	  500000	      3029 ns/op	      64 B/op	       4 allocs/op
Benchmark_EvalBool_OneOperator_ReferenceNodeInt64_ReferenceNodeInt64-4            	 3000000	       496 ns/op	      49 B/op	       3 allocs/op
Benchmark_EvalBool_OneOperatorWith11ScopeItem_ReferenceNodeInt64_NumberInt64-4    	 3000000	       544 ns/op	      64 B/op	       4 allocs/op
Benchmark_EvalBool_OneOperatorValueChanges_ReferenceNodeInt64_NumberInt64-4       	  500000	      3023 ns/op	      64 B/op	       4 allocs/op
ok  	github.com/influxdata/kapacitor/tick	59.607s
```

### DIFF
--- a/tick/stateful_expr_benchmarks_test.go
+++ b/tick/stateful_expr_benchmarks_test.go
@@ -1,0 +1,292 @@
+package tick_test
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/influxdata/kapacitor/tick"
+)
+
+/*
+	Benchmarks for evaluating stateful expression
+
+	We will test by few basic scenario:
+		*) One operator
+		*) Two operators - with AND and OR relationship
+
+	And will test with reference node and without to check the
+	isolation performance overhead of reference node
+
+    Test name format:
+        Benchmark_{Evaluation type: EvalBool or EvalNum}_{Type: OneOperator, TwoOperator}_{LeftNode}_{RightNode}
+*/
+
+func Benchmark_EvalBool_OneOperator_NumberFloat64_NumberFloat64(b *testing.B) {
+
+	emptyScope := tick.NewScope()
+	benchmarkEvalBool(b, emptyScope, &tick.BinaryNode{
+		Operator: tick.TokenGreater,
+		Left: &tick.NumberNode{
+			IsFloat: true,
+			Float64: float64(20),
+		},
+		Right: &tick.NumberNode{
+			IsFloat: true,
+			Float64: float64(10),
+		},
+	})
+}
+
+func Benchmark_EvalBool_OneOperator_NumberFloat64_NumberInt64(b *testing.B) {
+
+	emptyScope := tick.NewScope()
+	benchmarkEvalBool(b, emptyScope, &tick.BinaryNode{
+		Operator: tick.TokenGreater,
+		Left: &tick.NumberNode{
+			IsFloat: true,
+			Float64: float64(20),
+		},
+		Right: &tick.NumberNode{
+			IsInt: true,
+			Int64: int64(10),
+		},
+	})
+}
+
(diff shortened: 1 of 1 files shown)


tick/eval_benchmarks_test.go
```diff
@@ -0,0 +1,293 @@
+package tick_test
```
nathanielc: Can we name the file `stateful_expr_benchmarks_test.go` since it only applies the the stateful expression  evaluation?

yosiat:
@nathanielc done!

tick/stateful_expr_benchmarks_test.go
```diff
@@ -0,0 +1,293 @@
...
+
+func benchmarkEvalBool(b *testing.B, scope *tick.Scope, node tick.Node) {
+	b.ReportAllocs()
+	b.ResetTimer()
+
+	se := tick.NewStatefulExpr(node)
+	var result bool
+
+	for i := 0; i < b.N; i++ {
+		result, err := se.EvalBool(scope)
```
nathanielc: This `result` is hiding the func level `result` var because of the `:=` operator. Declare a `var err error` outside the loop as well so you can use the `=` operator.

yosiat: To be honest the reason for this was to escape compiler optimisation, but the next line after evaluating is checking result and err values. so I am pretty sure we can remove this "confusing" var result bool and evalBoolResult. @nathanielc what do you think?

nathanielc: I think we still want to avoid the compiler optimization issue.

```
var err error
for i := 0; i < b.N; i++ {
        evalBoolResult, err = se.EvalBool(scope)
        if err != nil {
              t.Fatal(err)
        }
}
```

that should do it, just remove the result var and use `evalBoolResult` directly. Then check the error and terminate the test if there is one since it shouldn't happen.

nathanielc: We don't actually care about the value of `evalBoolResult` we just need the compiler to think we do.

yosiat: Ok, I will fix this in a minute 👍

nathanielc:
@yosiat  Thanks! I'll merge on green.

nathanielc:
Merged, anyway... the only failed test was #474 which I just fixed in #487

PR #477, opened 2016-04-19
Add elapsed function

### LINKED ISSUE #384: TICK script - how to find elapsed time between two measurement points
I'm having difficulties to create a TICK script, that can find elapsed time between two measurement points. My Influx Measurement contains state points with values "on" and "off" (always in this order). I'd like to trigger event once "off" state arrives and calculates elapsed time between "on" and "off" state.

nathanielc:
Fixes #384

Add an `elapsed` function from the InfluxQL functions. Allows for the elapsed time between subsequent points to be measured.


seb-koch:
Could you please show a brief example? I am struggling to understand how the function expects its input.

PR #484, opened 2016-04-21
Proposed HTTP API Changes

nathanielc:
Up until this point the Kapacitor HTTP API has been "implemented" rather than "designed". This PR presents an API "design" in the form of documentation, in preparation for locking down the API to a more supportable long lived API.

Currently the existing API has some inconsistencies and odd usage patterns. This doc attempts to unify and simplify using the API. It makes the API more RESTful and consistent with itself.

For the most part only small changes where made to the existing API. For example when query parameters vs JSON structures vs path elements are used is clear and consistent. Parameter names and top level keys in JSON structures all use lower case names. 

This represents  a breaking change to the API. It is intended that once we settle on this API design that it will not need any further changes before it becomes a stable locked down 1.0 version.

This PR only documents the proposed behavior of the API but does not implement any of the changes. Once accepted another PR will finalize the implementation.

Some notable usage changes from the existing workflow.

* Task pattern matching is only supported on the `list task` requests. To enable/disable multiple tasks matching a pattern one must first list matching tasks and then make the appropriate request for each returned task. This simplifies using the API since each task has a unique path and only modified via that path. If this is not going to work for some's use case please comment and we can work on a consistent solution, but I defaulted for simplicity in this case.
* No more plural path names like `/tasks` or `/recordings` since the lack of a child path indicates you want all to interact with all entities.

### DIFF
--- a/client/API.md
+++ b/client/API.md
@@ -388,15 +388,17 @@ A recording ID is returned to later identify the recording.
 
 ##### Stream
 
-| Parameter | Purpose                                                             |
-| --------- | -------                                                             |
-| task      | ID of a task, used to only record data for the DBRPs of the task. |
-| stop      | Record stream data until stop date.                                 |
+| Parameter | Purpose                                                                    |
+| --------- | -------                                                                    |
+| id        | Unique identifier for the recording. If empty a random one will be chosen. |
+| task      | ID of a task, used to only record data for the DBRPs of the task.          |
+| stop      | Record stream data until stop date.                                        |
 
 ##### Batch
 
 | Parameter | Purpose                                                                                                          |
 | --------- | -------                                                                                                          |
+| id        | Unique identifier for the recording. If empty a random one will be chosen.                                       |
 | task      | ID of a task, records the results of the queries defined in the task.                                            |
 | start     | Earliest date for which data will be recorded. RFC3339Nano formatted.                                            |
 | stop      | Latest date for which data will be recorded. If not specified uses the current time. RFC3339Nano formatted data. |
@@ -404,11 +406,12 @@ A recording ID is returned to later identify the recording.
 
 ##### Query
 
-| Parameter | Purpose                                                                   |
-| --------- | -------                                                                   |
-| type      | Type of recording, `stream` or `batch`.                                   |
-| query     | Query to execute.                                                         |
-| cluster   | Name of a configured InfluxDB cluster. If empty uses the default cluster. |
+| Parameter | Purpose                                                                    |
+| --------- | -------                                                                    |
+| id        | Unique identifier for the recording. If empty a random one will be chosen. |
+| type      | Type of recording, `stream` or `batch`.                                    |
+| query     | Query to execute.                                                          |
+| cluster   | Name of a configured InfluxDB cluster. If empty uses the default cluster.  |
 
 >NOTE: A recording itself is typed as either a stream or batch recording and can only be replayed to a task of a corresponding type.
 Therefore when you record the result of a raw query you must specify the type recording you wish to create.
@@ -451,6 +454,17 @@ Create a recording using the `query` method specifying a `batch` type.
 ```
 POST /kapacitor/v1/recording/query
 {
+    "query" : "SELECT mean(usage_idle) FROM cpu WHERE time > now() - 1h GROUP BY time(10m)",
+    "type" : "batch"
+}
+```
+
+Create a recording with a custom ID.
+
+```
+POST /kapacitor/v1/recording/query
+{
+    "id" : "MY_RECORDING_ID",
     "query" : "SELECT mean(usage_idle) FROM cpu WHERE time > now() - 1h GROUP BY time(10m)",
     "type" : "batch"
 }
@@ -633,6 +647,18 @@ POST /kapacitor/v1/replays/
     "recording" : "RECORDING_ID",
     "clock" : "real",
     "recording-time" : true,
+}
+```
+
+
+Replay a recording using a custom ID.
+
+```
+POST /kapacitor/v1/replays/
+{
+    "id" : "MY_REPLAY_ID",
+    "task" : "TASK_ID",
+    "recording" : "RECORDING_ID"
 }
 ```
 
--- a/client/API.md
+++ b/client/API.md
@@ -5,14 +5,15 @@ With the API you can control which tasks are executing, query status of tasks an
 
 Each section below defines the available API endpoints and there inputs and outputs.
 
+All requests are versioned and namespaced using the root path `/kapacitor/v1/`.
+
 ### Response Codes
 
 All requests can return these response codes:
 
 | HTTP Response Code | Meaning                                                                                                                                                             |
 | ------------------ | -------                                                                                                                                                             |
-| 200                | The request was a success the content is dependent on the request.                                                                                                  |
-| 204                | The request was a success and no content was returned.                                                                                                              |
+| 2xx                | The request was a success the content is dependent on the request.                                                                                                  |
 | 4xx                | Invalid request, refer to error for what it wrong with the request. Repeating the request will continue to return the same error.                                   |
 | 5xx                | The server was unable to process the request, refer to the error for a reason. Repeating the request may result in a success if the server issue has been resolved. |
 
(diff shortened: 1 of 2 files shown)


client/API.md
```diff
@@ -0,0 +1,536 @@
+# Kapacitor API Reference Documentation
+
+Kapacitor provides an HTTP API on port 9092 by default.
+With the API you can control which tasks are executing, query status of tasks and manage recordings etc.
+
+Each section below defines the available API endpoints and there inputs and outputs.
```
nathanielc: This document needs a TOC.

nathanielc:
@yosiat @goller @arussellsaw How do you like this proposed change?

arussellsaw:
:+1: will this also be hosted on the influxdata kapacitor docs page?

nathanielc:
@arussellsaw Yes

client/API.md
```diff
@@ -0,0 +1,536 @@
...
+
+## Tasks
+
+A task represents work for Kapacitor to perform.
+A task is defined by its name, type, TICKscript, and list of database retention policy pairs it is allowed to access.
+
+### Define Task
+
+To define a task POST to the `/task/TASK_NAME` endpoint.
+If no task exists it with that name it will be created.
```
goller: remove "it"

client/API.md
```diff
@@ -0,0 +1,536 @@
...
+    "type" : "stream",
+    "dbrps": [{"db": "DATABASE_NAME", "rp" : "RP_NAME"}],
+    "script": "stream\n    |from()\n        .measurement('cpu')\n"
+}
+```
+
+Modify only the dbrps of the task.
+
+```
+POST /task/TASK_NAME
```
goller: Modifying an existing REST resource should be PATH instead of POST.  Many times it is important to not create an entry more than once.

client/API.md
```diff
@@ -0,0 +1,536 @@
...
+| 204  | Success |
+
+
+### Get Task
+
+To get information about a task make a GET request to the `/task/TASK_NAME` endpoint.
+
+| Query Parameter | Default | Purpose                                                                                                                                                            |
+| --------------- | ------- | -------                                                                                                                                                            |
+| dot-labels      | false   | Return the DOT formatted string using only label attributes. This is useful if you intend to render a diagram from the DOT content. Otherwise it is less readable. |
```
goller: Instead of bools I think there are other options that will help for the future:

1. ?fields=dot-labels&fields=stat  There could also be defaults
2. ?view=all or ?view=default or ?view=foo where a few is a named list of fields.

client/API.md
```diff
@@ -0,0 +1,536 @@
...
+
+| Code | Meaning             |
+| ---- | -------             |
+| 200  | Success             |
+| 404  | Task does not exist |
+
+
+### Enable Task
+
+To enable a task and start it executing make a POST request to the `/task/TASK_NAME/enable` endpoint.
```
goller: I think this should be a PATCH to `/TASK_NAME`

```http
PATCH /TASK_NAME HTTP/1.1
Content-Type: application/json

{"enabled": "true"}
```

This way you are modifying the resource "implying" an RPC

client/API.md
```diff
@@ -0,0 +1,536 @@
...
+| Query Parameter | Purpose                                                                                                                                           |
+| --------------- | -------                                                                                                                                           |
+| pattern         | Filter results based on the pattern. Uses standard shell glob matching, see [this](https://golang.org/pkg/path/filepath/#Match) for more details. |
+
+#### Example
+
+Get all tasks.
+
+```
+GET /task
```
goller: This should have pagination when there are many tasks.  So,  I think it should have RFC5988 links:

```http
GET /task HTTP/1.1

Link: <https://kapa.com/task?offset=20&limit=10>; rel="next",<https://kapa.com/task?offset=10&limit=10>; rel="self",<https://kapa.com/task?offset=0&limit=10>; rel="prev",
```

In addition to RFC5988 header links, some REST interfaces also put a `link` object in the response:

```json
"links" : [
  {"rel": "next", "href": "https://kapa.com/task?offset=20&limit=10"},
  {"rel": "prev", "href": "https://kapa.com/task?offset=0&limit=10"},
  {"rel": "self", "href": "https://kapa.com/task?offset=10&limit=10"},
]
```
... I find these extra link sections a bit too verbose, but, it can be nice for moving through pages in browser.

goller: I think this resource should be `/tasks` as the resource is a many resource.

client/API.md
```diff
@@ -0,0 +1,536 @@
...
+    ]
+}
+```
+
+#### Response
+
+| Code | Meaning                |
+| ---- | -------                |
+| 200  | Success                |
+| 404  | No tasks match pattern |
```
goller: I think 404 should be for resources that truly don't exist.  I think this should return an empty list of tasks.

```json
{
    "tasks" : []
}
```

... this makes client code way easier... rather than if statements checking status codes you simply iterate over the response.

client/API.md
```diff
@@ -0,0 +1,536 @@
...
+There are three methods for recording data with Kapacitor:
+
+| Method | Description                                        |
+| ------ | -----------                                        |
+| stream | Record the incoming stream of data.                |
+| batch  | Record the results of the queries in a batch task. |
+| query  | Record the result of an explicit query.            |
+
+All record POST requests must specify the parameter `method` with one of the above options.
+Each different `method` has it's own parameters.
```
goller: Because each method has its own parameters I think that's an indication that there should be three different resources:

- `/recording/stream`
- `/recording/batch`
- `/recording/query`

client/API.md
```diff
@@ -0,0 +1,536 @@
...
+
+The request returns once the recording is started and does not wait for it to finish.
+A recording ID is returned to later identify the recording.
+
+##### Stream
+
+| Parameter | Purpose                                                             |
+| --------- | -------                                                             |
+| name      | Name of a task, used to only record data for the DBRPs of the task. |
+| duration  | Duration string indicating how long to record the stream.           |
```
goller: Is this a golang duration string?  It may be better to go with a rfc3339 date to indicate when to stop.  rfc3339 has much deeper support with other languages.

client/API.md
```diff
@@ -0,0 +1,536 @@
...
+```
+
+
+#### Response
+
+All recordings are assigned an ID which is returned in this format.
+
+```
+{
+    "id" : "e24db07d-1646-4bb3-a445-828f5049bea0"
```
goller: I think this should respond with a Location of the nearly created resource:

```http
HTTP 201 Created
Location: http://kapa.com/recording/e24db07d-1646-4bb3-a445-828f5049bea0
```

and/or a link in the response:

```json
{
    "id" : "e24db07d-1646-4bb3-a445-828f5049bea0",
    "link": {"rel": "self", "href": "/recording/e24db07d-1646-4bb3-a445-828f5049bea0"}
}
```

client/API.md
```diff
@@ -0,0 +1,536 @@
...
+}
+```
+
+| Code | Meaning                             |
+| ---- | -------                             |
+| 200  | Success, the recording has started. |
```
goller: I think the status code should be 201 or 202.

client/API.md
```diff
@@ -0,0 +1,536 @@
...
+
+### Wait for Recording
+
+In order to determine when a recording has finished you must request it by making a GET request to `/recording/RECORDING_ID`.
+A GET request for a recording will block until the recording is complete.
```
goller: Blocking can be really problematic.  It should just do the normal 202 response when not complete and 200 when complete.

client/API.md
```diff
@@ -0,0 +1,536 @@
+# Kapacitor API Reference Documentation
```
goller: This api should be versioned:

`/v1/kapa/tasks` etc

This will ease the transition from your existing api to something else.

Additionally, when one tries to sew many APIs routes into a single domain it is nice if they are namespaced.

client/API.md
```diff
@@ -0,0 +1,536 @@
...
+{
+    "name" : "TASK_NAME",
+    "clock" : "real",
+    "recording-time" : true,
+}
+```
+
+#### Response
+
+The request blocks until the replay is complete.
```
goller: I think this should 202 until finished, then, 204

client/API.md
yosiat: @goller I think this should be an rpc style, sending a PATCH that only supports changing the enablement status is confusing, users might think you can change the script/dbrps/etc via this api.

client/API.md
yosiat: I totally agree with @goller, and I think it would be nice to add recording progress status, something like:
```
> GET /recording/e24db07d-1646-4bb3-a445-828f5049bea0

{
   "status": "running", // will be "running" or "done"
   "pct": 0.8 // how much recorded , for example the recording duration is 100m and it's done only 80m
}
```

yosiat:
@nathanielc this looks great! added in-line comment about the recordings.

By the way - "/debug/vars" will stays? or it's removed?

nathanielc:
> By the way - "/debug/vars" will stays? or it's removed?

It will stay.

client/API.md
goller: @yosiat  In many ways, I think it would be nice to change the script, dbrps, etc via a PATCH.

client/API.md
nathanielc: To be clear, POST to create  PATCH to modify? Works for me.

client/API.md
nathanielc: @goller More clarification, as per above you recommend using `PATCH` to modify an existing script instead of POST which makes sense. Here you are recommending dropping the `enable` part of the path and using `PATCH` again. 

This would mean enabling/disabling and modifying a task are all PATCH method calls to the same path. 

Does that means something like this is valid:

```
PATCH /task/TASK_NAME
{
    "script": "....",
    "enable": true
}
```

Which would modify the script and then enable the task in the same request?

client/API.md
nathanielc: Won't it be confusing to POST to `/recording/stream` and then GET `/recording/RECORDING_ID`?

client/API.md
nathanielc: OK, so a client will have to constantly poll to know when a recording is complete?

Seems like that's not too much different from a retry loop if your connection does have issues while blocking. And you make fewer requests and have a faster response time.

Although one disadvantage is around monitoring API performance since a blocking request is difficult to distinguish from a slow request.

client/API.md
nathanielc: So I forgot to mention this. The API also has the `/write` endpoint that behaves just like the InfluxDB `/write` endpoint. Since the InfluxDB API is not versioned most clients expect to hit `/write` and call it a day. I originally has the API versioned but found that most clients couldn't talk to Kapacitor as a result since they didn't allow the URL path to be changed, so I removed it. 

But since its only the `/write` path that has the issue maybe we version the whole API and then add an alias from `/write` to `/v1/write` for these older clients?

>Additionally, when one tries to sew many APIs routes into a single domain it is nice if they are namespaced.

Wouldn't that be the job of the sewer (aka load balancer)? Seem's like if your are doing this you have to use something to proxy the requests and whatever that is, it would  be trivial to add a path rewrite step.

nathanielc:
@goller @yosiat @arussellsaw I have updated the API doc per the many good suggestions.

In general the changes are:

1. Use POST for new creations, Use PATCH for modification. Only tasks can be modified.
2. Use `id` for everything so that words like `name` are not implicitly associated with `tasks` etc. IDs for everything can be client specified or generated randomly server side.
3. The three entities are `tasks`, `recordings` and `replays`. Each has its basic POST,GET, GET (list) endpoints.
4. All list requests can be paginated.
5. Both replay, and recordings will return progress information and never block.


This represent a lot of work but I like the consistency and flexibility it is going to provide. And most of the work is duplicated across entities so it should be too bad if done right (fingers crossed).

client/API.md
```diff
@@ -0,0 +1,808 @@
...
+
+| Query Parameter | Purpose                               |
+| --------------- | -------                               |
+| db              | Database name for the writes.         |
+| rp              | Retention policy name for the writes. |
+
+>NOTE: Kapacitor scopes all points by their database and retention policy. This means you MUST specify the `rp` for writes or Kapacitor will not know which retention policy to use.
+
+#### Example
+
```
yosiat: It will be nice to add reference to influxdb docs where they specify the format - http://docs.influxdata.com/influxdb/v0.12/write_protocols/line/

This endpoint should have precision, am I correct?

client/API.md
```diff
@@ -0,0 +1,808 @@
...
+In order to determine when a recording has finished you must make a GET request to the returned link typically something like `/kapacitor/v1/recordings/RECORDING_ID`.
+
+A recording has these read only properties.
+
+| Property | Description                                                                               |
+| -------- | -----------                                                                               |
+| id       | A unique identifier for the recording.                                                    |
+| type     | One of `stream` or `batch`. A recording cannot be replayed to a task of a different type. |
+| size     | Size of the recording on disk in bytes.                                                   |
+| date     | Date the recording finished.                                                              |
```
yosiat: If the recording isn't finished, the time will be estimated time?

yosiat:
I don't know if this is related to this pull request, but it will be nice if Kapacitor will expose (or have in the repo) a swagger file, then you can generate clients for all languages from go (https://goswagger.io/) to c#

nathanielc:
@yosiat  @goller  I have added a swagger file for the API. It probably has a few typos at this points since I haven't tested it but the general structure is all there.

nathanielc:
Merging this for now, implementation coming that makes a few small changes.

PR #488, opened 2016-04-22
Ignore IntelliJ IDEA files

satterly:


nathanielc:
@satterly This kind of ignore belongs in a personal gitignore file, not for the project. http://stackoverflow.com/questions/5724455/can-i-make-a-user-specific-gitignore-file

satterly:
Didn't know you could do that. Thanks.

PR #491, opened 2016-04-24
Compiled stateful expression

yosiat:
Hi,

This is a one big pull request with the next bottom line changes:

- Performance of evaluating stateful expression signifactly improved
- Added 11 unit tests for stateful expression and coverage got up from 16.2% to 18.2%
- **All tests are passing** - I changed all the usages of tick.NewStatefulExper to use the new one - and all integrations tests passed.
- There a behaviour changes - priority to errors have changes, etc - but in my opinion they are not big
- DurationNode is not supported
- Currently, I didn't replaced the stateful expression with the new one.

# Implementation
_Those are explanations of the core algorithm, if there are more questions/clarifications requested, I will update this._

## Basic explanation
The overall idea: Instead of using stack-based AST interpreter compiled the expressions to specialized functions.
For example: given this expression ``"value" > 8.0``, let's assume two assumptions:
- "value" is float64
- 8.0 is float64

The specializer will take this expression and will evantually run float64 > float64 all the time, instead of doing for every evaluation:
- Type checking and guessing: checking the type of ref node and the right node type
- Run through the whole AST tree

## Deeper explanation
First, let's set up simple terminology:
* Dynamic Node - node that is value changes on runtime like FunctionNode and ReferenceNode
* Constant Node - node that is value is constant for the whole lifetime of the tick script
* Evaluation Function - evaluation function is the function that accepts three arguments: scope, left and right node (this is a simplified version)

When we get a BinaryNode we determine if it's dynamic or constant - let's examine the dynamic case.

If this is dynamic node in the constructor (NewStatefulExpr) we set the evaluation function to be "dynamic evaluation function" otherwise
we fetch the matching evaluation function based on the nodes types and their operator.

The dynamic evaluation function is doing the next instructions (_this were the "specialization" happens_):
* Read the values of the left and right node (for example: for a reference node we will access the scope and read the value)
* Find a matching evaluation function based on the types we got and save it (in field in the StatefulExpression struct)
* call EvalBool

**The real meat is in EvalBool/EvalNum:**
1. If the evaluation function is null it means that we have some error:
 1. Type mismatch: int > string
 2. Not a comparison/math operator: int - int
 3. Invalid operator for type: bool > bool
2. We have evaluation function and evaluate her - the evaluation function returns bool and error
3. We examine the error if it's our special error (ErrTypeGuardFailed) that indicates we ran the wrong comparison function - this can happen on type changes - for example: "value" started as int64 and eventually changed to float6
[...]

PR #496, opened 2016-04-29
Scope reusing & smaller scopes

yosiat:
**This pull request, is experiment, If you like the idea, we can improve the readability and the quality of the code**

For each expression we are creating "scope pool", which is object pool of scopes - with some extra magic.
By doing quick analysis on the node AST I know which tags and fields he requires. so we put only the required ones.
For example: "value" > 10, I fill only "value" from field or tag.
```
name                     old time/op    new time/op    delta
_T10_P500_AlertTask-4       133ms ± 4%     123ms ± 4%   -7.58%  (p=0.008 n=5+5)
_T10_P50000_AlertTask-4     13.4s ± 8%     12.3s ± 7%     ~     (p=0.056 n=5+5)
_T1000_P500_AlertTask-4     13.5s ± 4%     12.1s ± 3%  -10.46%  (p=0.008 n=5+5)

name                     old alloc/op   new alloc/op   delta
_T10_P500_AlertTask-4      32.2MB ± 0%    26.0MB ± 0%  -19.32%  (p=0.008 n=5+5)
_T10_P50000_AlertTask-4    3.26GB ± 0%    2.62GB ± 0%  -19.71%  (p=0.008 n=5+5)
_T1000_P500_AlertTask-4    3.21GB ± 0%    2.61GB ± 0%  -18.56%  (p=0.008 n=5+5)

name                     old allocs/op  new allocs/op  delta
_T10_P500_AlertTask-4        408k ± 0%      335k ± 0%  -17.85%  (p=0.008 n=5+5)
_T10_P50000_AlertTask-4     41.5M ± 0%     34.1M ± 0%  -17.98%  (p=0.008 n=5+5)
_T1000_P500_AlertTask-4     40.2M ± 0%     33.1M ± 0%  -17.61%  (p=0.008 n=5+5)
```

I thought about this idea while researching the performance of alerts, but before that I wanted to implement "compiled stateful expression" ( #491 ).
If we combine this pull request with #491 , we will have great performance and low memory usage while evaluating predicates.

### DIFF
--- a/expr.go
+++ b/expr.go
@@ -10,11 +10,14 @@ import (
 )
 
 // EvalPredicate - Evaluate a given expression as a boolean predicate against a set of fields and tags
-func EvalPredicate(se stateful.Expression, now time.Time, fields models.Fields, tags models.Tags) (bool, error) {
-	vars, err := mergeFieldsAndTags(now, fields, tags)
+func EvalPredicate(se stateful.Expression, scopePool stateful.ScopePool, now time.Time, fields models.Fields, tags models.Tags) (bool, error) {
+	vars := scopePool.Get()
+	defer scopePool.Put(vars)
+	err := fillScope(vars, scopePool.ReferenceVariables(), now, fields, tags)
 	if err != nil {
 		return false, err
 	}
+
 	b, err := se.EvalBool(vars)
 	if err != nil {
 		return false, err
@@ -22,17 +25,30 @@ func EvalPredicate(se stateful.Expression, now time.Time, fields models.Fields, 
 	return b, nil
 }
 
-func mergeFieldsAndTags(now time.Time, fields models.Fields, tags models.Tags) (*tick.Scope, error) {
-	scope := tick.NewScope()
-	scope.Set("time", now.Local())
-	for k, v := range fields {
-		if _, ok := tags[k]; ok {
-			return nil, fmt.Errorf("cannot have field and tags with same name %q", k)
+// fillScope - given a scope and reference variables, we fill the exact variables from the now, fields and tags.
+func fillScope(vars *tick.Scope, referenceVariables []string, now time.Time, fields models.Fields, tags models.Tags) error {
+	for _, refVariableName := range referenceVariables {
+		if refVariableName == "time" {
+			vars.Set("time", now.Local())
+			continue
 		}
-		scope.Set(k, v)
+
+		// Support the error with tags/fields collison
+		var fieldValue interface{}
+		var isFieldExists bool
+
+		if fieldValue, isFieldExists = fields[refVariableName]; isFieldExists {
+			vars.Set(refVariableName, fieldValue)
+		}
+
+		if tagValue, ok := tags[refVariableName]; ok {
+			if isFieldExists {
+				return fmt.Errorf("cannot have field and tags with same name %q", refVariableName)
+			}
+
+			vars.Set(refVariableName, tagValue)
+		}
 	}
-	for k, v := range tags {
-		scope.Set(k, v)
-	}
-	return scope, nil
+
+	return nil
 }
--- a/alert.go
+++ b/alert.go
@@ -98,6 +98,7 @@ type AlertNode struct {
 	endpoint    string
 	handlers    []AlertHandler
 	levels      []stateful.Expression
+	scopePools  []stateful.ScopePool
 	states      map[models.GroupID]*alertState
 	idTmpl      *text.Template
 	messageTmpl *text.Template
@@ -274,13 +275,16 @@ func newAlertNode(et *ExecutingTask, n *pipeline.AlertNode, l *log.Logger) (an *
 
 	// Parse level expressions
 	an.levels = make([]stateful.Expression, CritAlert+1)
+	an.scopePools = make([]stateful.ScopePool, CritAlert+1)
 
 	if n.Info != nil {
 		statefulExpression, expressionCompileError := stateful.NewExpression(n.Info)
 		if expressionCompileError != nil {
 			return nil, fmt.Errorf("Failed to compile stateful expression for info: %s", expressionCompileError)
 		}
+
 		an.levels[InfoAlert] = statefulExpression
+		an.scopePools[InfoAlert] = stateful.NewScopePool(stateful.FindReferenceVariables(n.Info))
 	}
 
 	if n.Warn != nil {
@@ -289,6 +293,7 @@ func newAlertNode(et *ExecutingTask, n *pipeline.AlertNode, l *log.Logger) (an *
 			return nil, fmt.Errorf("Failed to compile stateful expression for warn: %s", expressionCompileError)
 		}
 		an.levels[WarnAlert] = statefulExpression
+		an.scopePools[WarnAlert] = stateful.NewScopePool(stateful.FindReferenceVariables(n.Warn))
 	}
 
 	if n.Crit != nil {
@@ -297,6 +302,7 @@ func newAlertNode(et *ExecutingTask, n *pipeline.AlertNode, l *log.Logger) (an *
 			return nil, fmt.Errorf("Failed to compile stateful expression for crit: %s", expressionCompileError)
 		}
 		an.levels[CritAlert] = statefulExpression
+		an.scopePools[CritAlert] = stateful.NewScopePool(stateful.FindReferenceVariables(n.Crit))
 	}
 
 	// Setup states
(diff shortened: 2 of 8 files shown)


nathanielc:
@yosiat Thanks! I was hoping this would also get done. At a quick glance it looks good to go once #491 gets merged.

yosiat:
@nathanielc I think we need to define how it will be used.
I currently stuck in StatefulExpression but it might be more clean and ok to be it in it's own struct that StatefulExpression will own?

nathanielc:
>I currently stuck in StatefulExpression but it might be more clean and ok to be it in it's own struct that StatefulExpression will own?

I think having it on its own would be good.

yosiat:
@nathanielc I will fix that tomorrow and this PR will be ready to merge; I am going to sleep 💤

nathanielc:
@yosiat yes, please get some rest :sleeping:

yosiat:
@nathanielc I finished this PR 😄

- Refactored the code and the tests
- Rebased with the new expressions

**Benchmarks (compared to master with new stateful expressions)**
_I updated the benchmarks in the pull request as well_
Ran with count=5
```
name                     old time/op    new time/op    delta
_T10_P500_AlertTask-4       133ms ± 4%     123ms ± 4%   -7.58%  (p=0.008 n=5+5)
_T10_P50000_AlertTask-4     13.4s ± 8%     12.3s ± 7%     ~     (p=0.056 n=5+5)
_T1000_P500_AlertTask-4     13.5s ± 4%     12.1s ± 3%  -10.46%  (p=0.008 n=5+5)

name                     old alloc/op   new alloc/op   delta
_T10_P500_AlertTask-4      32.2MB ± 0%    26.0MB ± 0%  -19.32%  (p=0.008 n=5+5)
_T10_P50000_AlertTask-4    3.26GB ± 0%    2.62GB ± 0%  -19.71%  (p=0.008 n=5+5)
_T1000_P500_AlertTask-4    3.21GB ± 0%    2.61GB ± 0%  -18.56%  (p=0.008 n=5+5)

name                     old allocs/op  new allocs/op  delta
_T10_P500_AlertTask-4        408k ± 0%      335k ± 0%  -17.85%  (p=0.008 n=5+5)
_T10_P50000_AlertTask-4     41.5M ± 0%     34.1M ± 0%  -17.98%  (p=0.008 n=5+5)
_T1000_P500_AlertTask-4     40.2M ± 0%     33.1M ± 0%  -17.61%  (p=0.008 n=5+5)
```

I think we can do the same for eval.go using some pooling their, what do you think? do we have good test coverage for what's happening there?

nathanielc:
@yosiat Looks good. Can we also update the `eval.go` usage to use the scope pool? Basically any place that use `mergeFieldsAndTags` should be updated and then I think we can remove the `mergeFieldsAndTags` function.

yosiat:
@nathanielc It's a bit problematic there because we have multiple expressions, we can create at startup "combined reference variables" set and use it, what do you think?

nathanielc:
@yosiat I don't believe the scope is shared across the expressions, so they should all be able to operate independently.

yosiat:
@nathanielc Looking at the code, looks like they use the same "vars", and after evaluation you are updating the scope..

It looks like the second expression is using the result of the first expression, etc.

nathanielc:
@yosiat Yes, the `scope` cascades across expressions but the `state` is independent, and the set of referenced values is constant as well.

You should be able to use `ScopePool.Get` and check if the next expression even wants the result of the previous expression.  In other words create a scope for each expression and then only cascade needed values. The currently implementation is lazy and just reuses the same scope.

yosiat:
umpf, so why not simply on upstart get the reference variables used in all scopes and merge them to one set? so the initial scope will be the merged? in term of performance your approach will win, but this approach will be much simpler.

nathanielc:
Go with the simpler approach :) The biggest win is to remove allocations for each evaluation of the expression. Other than that lets keep it simple.

yosiat:
What do you think about creating "MergedScopePool" which will accept list of nodes and and create scope pool for them?

nathanielc:
I would rather not create another type. 

What if the `NewScope` method took a list of names `[]string` instead of a node. Then you could easily create ScopePools for difference use cases. You would need to expose a way to get the referencedVariables still. I think adding a method `ReferencedVariables` to the `StatefulExpression` would work as it has a reference to the node already. Then remove the `ScopePool` method from the `StatefulExpression`, in other words let the caller use the pool but don't require it. Then in `eval.go` on upstart you would call `ReferencedVariable` for each expression and merge them and create a single ScopePool for all the expressions. Does that makes sense? Basically instead of pushing the merge logic down into the `stateful` package leave it in the `eval.go` code since its the code that needs it.

yosiat:
@nathanielc liked your idea, I will start to play with it and update ~

yosiat:
@nathanielc I will make "findReferenceVariables" a function on top of Node and will put under "tick/reflect.go"

yosiat:
@nathanielc after thinking a bit, if the StatefulExpression won't own the scope pool, it means that:

- alert: will need to contain a symmetric array for holding the scope pools
- EvalPredicate: will need to accept the scope pool

Are we ok with those changes?

nathanielc:
@yosiat +1

yosiat:
@nathanielc This is getting some what tedious, In where, alert and eval we have to "maintain" symmetric array for holding the scope pools. I think it might hear readability.

nathanielc:
@yosiat Then maybe you can supply the ScopePool to the StatefulExpression and it does the `maintaining`? But it still allows for the use case of the pool to be different.

yosiat:
@nathanielc I thought about it, but then it will be too much weird, because they will some dependency.

In order to create ScopePool you will need reference variables, which you get by calling "ReferenceVariables" on StatefulExpression.

nathanielc:
@yosiat OK, you are more in the details at this point. Do what makes sense...

yosiat:
@nathanielc What do you think? I haven't integrated it in eval.go yet..

nathanielc:
@yosiat 

>In order to create ScopePool you will need reference variables, which you get by calling "ReferenceVariables" on StatefulExpression.

This is essentially the only difference between alert/where and eval, is how the scope is created. It is used in the same manner. So either have a little duplicate code in alert and where or add helper functions to the `stateful` package. Something like `CreateScopePoolForExpressions` and it takes a list of expressions. Then eval passes in a list and where/alert pass in a list of one.

yosiat:
@nathanielc do we want "CreateScopePoolForExpressions" to replace "NewScope" or to be helper method in stateful/scope_pool.go? (I think the latter)

nathanielc:
@yosiat  The later, if something wants to create a scope in yet another different way it is still possible if we leave NewScope as is. Also that is why I called it `Create...` because there is implied logic happening not simply instantiating a new object.

yosiat:
@nathanielc I started implementing the `CreateScopePoolForExpressions` in eval.go, and it looks like we have another problem.
We need `ReferenceVariables` for `fillScopePool (in expr.go)` and in eval.go we will need to have another list for it.

I think instead of `CreateScopePoolForExpressions` we will create: `FindReferenceVariablesForNodes` and remove it from Expression.

So the usage will be something like:
```go
var node tick.Node // let's say we have node
refVariables := stateful.FindReferenceVariablesForNodes(node)
stateful.NewScopePool(refVariables)
```

and in eval.go, it will something like this:
```go
refVariables := stateful.FindReferenceVariablesForNodes(n.Expressions...)
stateful.NewScopePool(refVariables)
```

yosiat:
@nathanielc It's finished, squashed & updated benchmarks.

expr.go
```diff
@@ -10,29 +10,44 @@ import (
...
-	scope := tick.NewScope()
-	scope.Set("time", now.Local())
-	for k, v := range fields {
-		if _, ok := tags[k]; ok {
-			return nil, fmt.Errorf("cannot have field and tags with same name %q", k)
+// fillScope - given a scope and reference variables, we fill the exact variables from the now, fields and tags.
+func fillScope(vars *tick.Scope, referenceVariables []string, now time.Time, fields models.Fields, tags models.Tags) error {
+	for _, refVariableName := range referenceVariables {
+		if refVariableName == "time" {
+			vars.Set("time", now.Local())
```
nathanielc: Should we `break` after this line? That way we don't accidentally set a field value named `time`.

yosiat: Agree, but not `break`, `continue`

nathanielc: Ah yes :)

nathanielc:
@yosiat :+1:  I'll merge on green. Since nothing user facing changes no need to write a CHANGELOG.

Issue #501, opened 2016-05-02
Allow bools to be converted to ints in expressions


nathanielc:
Add support to the int function https://github.com/influxdata/kapacitor/blob/master/tick/functions.go#L253

for converting a boolean to a 1 or 0.

Relevant discussion:
https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/influxdb/5ptiuF54DJI/QdcfSlWjAgAJ

Issue #495, opened 2016-04-28
[0.12.0] distinct() and count() functions do not work on string fields


cnelissen:
I have a TickScript like:

```
stream
    |from()
        .database('mydb')
        .measurement('a')
    |distinct('stringField')
    |count('stringField')
```

and I am seeing errors in the logs like:

```
[edge:test|stream1->distinct2] 2016/04/28 20:08:08 I! aborting c: 22 e: 1
[test:distinct2] 2016/04/28 20:08:08 E! invalid field type: string
[edge:test|stream->srcstream0] 2016/04/28 20:08:08 D! closing c: 16208 e: 16208
[edge:test|srcstream0->stream1] 2016/04/28 20:08:08 D! closing c: 16208 e: 16208
[edge:test|stream1->distinct2] 2016/04/28 20:08:08 D! closing c: 22 e: 1
[task_master] 2016/04/28 20:08:08 E! Stopped task: test invalid field type: string
[task_store] 2016/04/28 20:08:08 E! task test finished with error: invalid field type: string
```

I posted a message on the mailing list and I got a reply from @nathanielc that said:

"Kapacitor does not support string fields yet. It won't be hard to add, it just hasn't been done yet."

This is a showstopper for me since my queries depend on getting an accurate count based on a string field. I didn't see an open (or closed) issue about this but it seems like it is a known issue...

Any ETA on a fix for this?

nathanielc:
@cnelissen Thanks for filing an issue. The 0.13 milestone is coming to a close but I'll make this a priority for 0.14.

cnelissen:
What is the timeframe for v0.14? Do you have any ideas for a workaround? I am pretty much dead in the water without being able to query string fields...

nathanielc:
@cnelissen For a work around you could write a UDF to map strings to an int64s. You could use a hashing mechanism of if you know the static set via a direct mapping. Then Kapacitor will work well with the int64s. Then you could use another UDF to reverse the mapping back to strings.

See https://github.com/influxdata/kapacitor/tree/master/udf/agent/examples for example UDFs in Go and Python.

cnelissen:
@nathanielc Thanks for the ideas! I think I might have a solution in place.

nathanielc:
This was fixed in #524

PR #524, opened 2016-05-06
Support float,integer,string and boolean

### LINKED ISSUE #500: Fully support string,int,bool,float data types.
Kapacitor supports int and float operations currently. Add support for string and bool operations as well. This will enable full interaction with InfluxDB data sets since these are the same native types.  This also means that the `eval` expressions should support returning and manipulating these types as well. Recent work #491 has enabled extending the expressions in this way. 

Related issue and discussions:

#495 
https://groups.google.com/forum/?pli=1#!topic/influxdb/TY_2IuuIxs8
https://groups.google.com/forum/?pli=1#!topic/influxdb/5ptiuF54DJI

nathanielc:
Fixes #500 #495 #501 

This was way easier than expected. I also did a check for all uses of `float64` to make sure a field type wasn't assumed to be a `float64`.

TODO:

- [x] Enable EvalNode to return any of these types.


yosiat:
@nathanielc What do you mean by "Enable EvalNode to return any of these types.", we are supporting "EvalFloat", "EvalString" and etc.

nathanielc:
Yes 👍. I am waiting till #496 is merged though so we don't conflict.

On Sat, May 7, 2016, 1:51 AM Yosi Attias <notifications@github.com> wrote:

> @nathanielc <https://github.com/nathanielc> What do you mean by "Enable
> EvalNode to return any of these types.", we are supporting "EvalFloat",
> "EvalString" and etc.
>
> —
> <https://github.com/influxdata/kapacitor/pull/524#issuecomment-217614556>
>
-- 
Nathaniel Cook
Kapacitor Lead
https://influxdata.com

yosiat:
@nathanielc ok, I thought there is something in stateful that doesn't support those types.

PR #547, opened 2016-05-14
Buffer pooling alert post request body buffer

yosiat:
Added buffer pooling for post handler in alert.

I tried to add benchmark for this, by using `httptest.NewServer` and got: `accept: too many open files`, I tried to increase the maximum file descriptors to 4096 and didn't helped.

### DIFF
--- a/alert.go
+++ b/alert.go
@@ -779,13 +779,17 @@ func (a *AlertNode) renderMessageAndDetails(id, name string, t time.Time, group 
 // Alert handlers
 
 func (a *AlertNode) handlePost(post *pipeline.PostHandler, ad *AlertData) {
-	b, err := json.Marshal(ad)
+	bodyBuffer := a.bufPool.Get().(*bytes.Buffer)
+	defer a.bufPool.Put(bodyBuffer)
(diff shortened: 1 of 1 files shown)


nathanielc:
@yosiat Looks good thanks.

yosiat:
@nathanielc can you think of a way we can benchmark the post handler of alert?

nathanielc:
Using `httptest.NewServer` should work, `accept: too many open files` should only occur if the requests were being made massively parallel or they were not being closed after use. Could you share the code you wrote initially? 

Also I merged because it was a simple enough change that I didn't think it merited its own benchmark (it does have a test). But if you want to add the benchmark I am happy to help.

yosiat:
@nathanielc The code created httptest.NewServer and doing nothing in the handler function, and I passed the url of the httptest to the post handler, nothing special here.

I think I got the error (might be because of #572) or because we have 1000 tasks, which means 1000 requests..

nathanielc:
@yosiat  I think you are right. Worth trying again once its merged.

Issue #82, opened 2015-12-07
Multiple services for PagerDuty alert


oldmantaiter:
Some users might want alerts to take different escalation paths (eg in Pagerduty to multiple rotations), so adding support to reference "named" alert handlers of the same type would be useful.

I was thinking I could attempt to tackle this when I get some time, or at least open a discussion on the best way to accomplish this (while staying in the style guidelines of the project).

Very cool project by the way, really like the DSL and how it is structured.

EDIT: Not sure if this already exists, haven't had a whole bunch of time to go through the code. So if it does, please ignore this request, I was basing this off configuration examples.

nathanielc:
Do you want to provide some examples of how the DSL would look for 'named' alert handlers? I like the idea, if we can decide on a DSL structure we like I can point you in the right direction for implementing a solution. Thanks

oldmantaiter:
@nathanielc It would be along the lines of how the slack channel is configured. So it looks to be something that is just pagerduty specific at this point.

Based on VictorOps and how that is specified, it could either be done that way (adding another function to be chained in the DSL) or in the config.

As for examples:

Main config
```
[pagerduty.<name>] # Defaults to 'default'
  enabled = true
  service-key = "xxxxxxx"
```

DSL

```
.alert()
  .pagerDuty("<name>")
```

This would allow a bit of readability when configuring which schedule to escalate to. As I said above, we could also just have a chained function like `routingKey` for VictorOps and pass in the service key for Pagerduty in there, but that is less readable due to the random nature of the service key.

Let me know your thoughts. I have no problem with either way, but potentially adding the ability for "named" outputs that could be referenced (not only for Pagerduty) might be of use?

nathanielc:
I like the design. The configuration will probably change to something like:

```
[pagerduty]
  enabled = true
  [pagerduty.schedules]
    schedule_name = "service_key"
    another_schedule  = "different_service_key"
```

Which would match with a config structure like:
```
type Config struct {
	// Whether PagerDuty integration is enabled.
	Enabled bool `toml:"enabled"`
	// The PagerDuty API URL, should not need to be changed.
	URL string `toml:"url"`
	// Whether every alert should automatically go to PagerDuty
	Global bool `toml:"global"`
	// Schedules name -> service key
	Schedules map[string]string
}
```
    
Is `schedule` the right name? I am not very familiar with PagerDuty so I could be using the wrong name here. 

[...]

PR #554, opened 2016-05-17
Add optional .serviceKey('XXX') argument to pagerduty alerts.

savagegus:
Resolves issue #82, should work like VictorOps integration, if you do not specify a service key, the default service key from the configuration file is used. If you do include a service key, it is used instead of the default.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Release Notes
 
 ### Features
+- [#82](https://github.com/influxdata/kapacitor/issues/82): Multiple services for PagerDuty alert
 
 
 ### Bugfixes
--- a/alert.go
+++ b/alert.go
@@ -878,6 +878,7 @@ func (a *AlertNode) handlePagerDuty(pd *pipeline.PagerDutyHandler, ad *AlertData
 		return
 	}
 	err := a.et.tm.PagerDutyService.Alert(
+		pd.ServiceKey,
 		ad.ID,
 		ad.Message,
 		ad.Level,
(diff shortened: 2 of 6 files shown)


nathanielc:
@savagegus Thanks for the PR! Could you update the test [here](https://github.com/influxdata/kapacitor/blob/master/integrations/streamer_test.go#L3045 ) so that the second call to `pagerDuty` has a custom service key?

savagegus:
Of course, I followed the pattern from the slack().channel() tests.

nathanielc:
@savagegus Looks good. And I saw you signed the CLA already thanks. I'll merge on green.

savagegus:
Thanks so much, this was painless.

zygis:
Not working
```
kapacitor define task_name -type batch -tick tickScript.tick -dbrp someDB
invalid TICKscript: line 14 char 3: no method or property "serviceKey" on *pipeline.PagerDutyHandler
```
kapacitor version: 0.13.1 (tried v1.0.0-beta2 too)

tick script:
```
batch
    |query('''
        SELECT xx
        FROM xxx
        WHERE xxx
    ''')
        .period(2m)
        .every(30s)
    |alert()
        .id('xxx')
        .message('xxx')
        .crit(lambda: "count" < 2)
        .pagerDuty()
        .serviceKey("xxx")
```
What syntax should I use?

nathanielc:
@zygis You will need >= 1.0.0-beta1 and change the double quotes to single quotes. 

```
batch
    |query('''
        SELECT xx
        FROM xxx
        WHERE xxx
    ''')
        .period(2m)
        .every(30s)
    |alert()
        .id('xxx')
        .message('xxx')
        .crit(lambda: "count" < 2)
        .pagerDuty()
        .serviceKey('xxx')
```

It should work at that point.

PR #562, opened 2016-05-18
Support HTTP based subscriptions

nathanielc:
This PR depends on https://github.com/influxdata/influxdb/pull/6655

TODO:
- [x] CHANGELOG
- [x] Do not hardcode port 9092

### DIFF
--- a/etc/kapacitor/kapacitor.conf
+++ b/etc/kapacitor/kapacitor.conf
@@ -83,6 +83,22 @@ data_dir = "/var/lib/kapacitor"
   # This is insecure.
   insecure-skip-verify = false
 
+  # Maximum time to try and connect to InfluxDB during startup
+  startup-timeout = "5m"
+
+  # Turn off all subscriptions
+  disable-subscriptions = false
+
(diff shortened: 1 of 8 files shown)


allen13:
Looking forward to this feature!

etc/kapacitor/kapacitor.conf
```diff
@@ -83,6 +83,12 @@ data_dir = "/var/lib/kapacitor"
   # This is insecure.
   insecure-skip-verify = false
 
+  # Host part of a bind addres for UDP listeners.
+  # For example if a UDP listener is using port 1234
+  # and `udp-bind = "hostname_or_ip"`,
+  # then the UDP port will be bound to `hostname_or_ip:1234`
+  # The default empty value will bind to all addresses.
+  udp-bind = ""
```
nathanielc: @rossmcdonald Since I was in this code today, I added this option so that the UDP bind address is configurable separate from the `hostname` config option. Now the default is to bind to all addresses, which should work well.

PR #567, opened 2016-05-19
Fix issue with nested binary operators counting as constant typed

### LINKED ISSUE #561: Getting "mismatched type to binary operator" for int64/int64 or float64/float/64 in 0.13
Working a script for checking memory usage. We pull total and used (in bytes) from InfluxDB, so I need to generate the percent myself. In 0.12, I was able to do something akin to this, an example of usage being over 50%:

```
.info(lambda: float("used") / float("total") > 0.50)
```

Unfortunatley, since 0.13, damn near any combination of the above (using float or int) results in an error like this:

```
E! error evaluating expression: mismatched type to binary operator. got int64 > int64. see bool(), int(), float()
```

Still learning this, so my gut say it's something silly I'm missing, but I can't seem to figure it out.

And FWIW, I've managed to get similar errors for float64 > float64, int64 / int64 and int64 * int64.

nathanielc:
Fixes #561 

Changes the error detection to compile time if possible.

Also refactors the `invalid ... operator` errors to be left/right symmetric and correctly report Regex errors.

### DIFF
--- a/tick/stateful/expr_test.go
+++ b/tick/stateful/expr_test.go
@@ -2,8 +2,8 @@ package stateful_test
 
 import (
 	"errors"
+	"fmt"
 	"regexp"
-	"strings"
 	"testing"
 	"time"
 
@@ -18,7 +18,7 @@ type keyStruct struct {
 }
 
 func TestExpression_EvalNum_KeepsFunctionsState(t *testing.T) {
-	se := mustCompileExpression(t, &tick.FunctionNode{
+	se := mustCompileExpression(&tick.FunctionNode{
 		Func: "sigma",
 		Args: []tick.Node{&tick.ReferenceNode{Reference: "value"}},
 	})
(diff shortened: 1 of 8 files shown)


nathanielc:
@yosiat I refactored a bunch in the eval_binary_node, if you wanted to take a look it would be much appreciated.

tick/stateful/expr_test.go
```diff
@@ -1321,10 +1334,10 @@ func runCompiledEvalTests(
 	}
 }
 
-func mustCompileExpression(t *testing.T, node tick.Node) stateful.Expression {
+func mustCompileExpression(node tick.Node) stateful.Expression {
 	se, err := stateful.NewExpression(node)
 	if err != nil {
-		t.Fatalf("Failed to compile expression: %v", err)
+		panic(fmt.Sprintf("Failed to compile expression: %v", err))
```
yosiat: Why we do panic here? If one test is failing, we fail here the whole test suite.
By using "Fatal" we fail this test only.

yosiat:
@nathanielc LGTM, except the change in mustCompileExpression which I added a comment.

tick/stateful/expr_test.go
nathanielc: I changed it from panic so that you get an ugly stack trace ;)  The stack trace makes it easier to figure out which expression didn't compile. Otherwise you have to look through all the expressions in a test. 

Also it should only fail when you are changing a test and accidentally create an expression that fails to compile. If its possible for a given expression not to compile then don't use this method and check the error explicitly.

PR #578, opened 2016-05-26
Build updates

nathanielc:
Fix build.py to `go generate` before `go get`  so that the uncommitted flag can detect generate changes.

Update deps.


nathanielc:
@rossmcdonald How do the build.py changes look?

PR #593, opened 2016-06-02
Handle default values for InfluxDB Service Configs

titilambert:
- [ ] Rebased/mergable
- [ ] Tests pass
- [ ] CHANGELOG.md updated

### DIFF
--- a/services/influxdb/config.go
+++ b/services/influxdb/config.go
@@ -63,6 +63,18 @@ func NewConfig() Config {
 	}
 }
 
+func (c *Config) SetDefaultValues() {
+	if c.UDPBuffer == 0 {
+		c.UDPBuffer = udp.DefaultBuffer
+	}
+	if c.StartUpTimeout == toml.Duration(0) {
+		c.StartUpTimeout = toml.Duration(DefaultStartUpTimeout)
+	}
+	if c.SubscriptionProtocol == "" {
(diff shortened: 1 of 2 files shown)


nathanielc:
@titilambert This looks good. Do we care that it prevents the zero values from ever being set? For example with this change it is impossible to set an empty list of exclusions as it will always add back in the stats db.

services/influxdb/config.go
```diff
@@ -63,6 +63,23 @@ func NewConfig() Config {
 	}
 }
 
+func (c *Config) SetDefaultValues() {
+	if len(c.ExcludedSubscriptions) == 0 {
+		c.ExcludedSubscriptions = map[string][]string{
+			stats.DefaultDatabse: []string{stats.DefaultRetentionPolicy},
+		}
+	}
```
titilambert: @nathanielc Oh ! I think you're right, maybe I have to remove these lines ? (67-71)

nathanielc: I think we should remove the lines. While it would be nice to be able to set this default, I think its worse to not be able to set an empty set.

titilambert:
@nathanielc changes done

nathanielc:
@titilambert Thanks, I'll merge on green.

titilambert:
@nathanielc it's green ! yeah !

Issue #553, opened 2016-05-17
Periodically check if new InfluxDB DBRPs have been created


nathanielc:
Currently Kapacitor only creates subscriptions on startup. We should add both and API endpoint to tell Kapacitor to add update subscriptions as well as optionally do so on a periodic basis.

titilambert:
@nathanielc Are you working on it ? If not, I can get it (I really need this one)

nathanielc:
@titilambert I am not working on this ATM. Thanks for taking a look!

titilambert:
@nathanielc I already got something, I just need more tests before create a PR

schmurfy:
What about the api endpoints, was it added too ? I can't find it in the documentation but it may just not be documented so I prefer to ask directly. If There isn't it would be nice to have a way to tell kapacitor to updates its subscriptions.

nathanielc:
@schmurfy It has been added and was left undocumented.

To trigger an explicit event to update subscriptions make an empty bodied POST request to `/kapacitor/v1/subscriptions`

nathanielc:
@schmurfy I should have also mentioned that as of v1.1.0 which is has an rc2 release out now there is a new API for configuring Kapacitor. Since the API allows adding/removing and modifying InfluxDB clusters in the Kapacitor config it naturally needed to support dynamically linking subscriptions. Depending on why you needed to update them this may be a better option. See the CHANGELOG for more details.

schmurfy:
thakns I will have a look at it, for now this is for testing purposes, I create the influxdb database on the fly and was looking for a way for kapacitor to immediately notice them.

PR #604, opened 2016-06-03
Periodically check if new InfluxDB DBRPs have been created #553

titilambert:
- [ ] Rebased/mergable
- [ ] Tests pass
- [ ] CHANGELOG.md updated

### DIFF
--- a/services/influxdb/config.go
+++ b/services/influxdb/config.go
@@ -13,7 +13,8 @@ import (
 
 const (
 	// Maximum time to try and connect to InfluxDB during startup.
-	DefaultStartUpTimeout = time.Minute * 5
+	DefaultStartUpTimeout           = time.Minute * 5
+	DefaultSubscriptionSyncInterval = time.Minute * 1
 
 	DefaultSubscriptionProtocol = "http"
 )
@@ -34,15 +35,16 @@ type Config struct {
 	// Use SSL but skip chain & host verification
 	InsecureSkipVerify bool `toml:"insecure-skip-verify"`
 
-	Timeout               toml.Duration       `toml:"timeout"`
-	DisableSubscriptions  bool                `toml:"disable-subscriptions"`
-	SubscriptionProtocol  string              `toml:"subscription-protocol"`
-	Subscriptions         map[string][]string `toml:"subscriptions"`
-	ExcludedSubscriptions map[string][]string `toml:"excluded-subscriptions"`
-	UDPBind               string              `toml:"udp-bind"`
-	UDPBuffer             int                 `toml:"udp-buffer"`
-	UDPReadBuffer         int                 `toml:"udp-read-buffer"`
-	StartUpTimeout        toml.Duration       `toml:"startup-timeout"`
+	Timeout                  toml.Duration       `toml:"timeout"`
+	DisableSubscriptions     bool                `toml:"disable-subscriptions"`
+	SubscriptionProtocol     string              `toml:"subscription-protocol"`
+	Subscriptions            map[string][]string `toml:"subscriptions"`
+	ExcludedSubscriptions    map[string][]string `toml:"excluded-subscriptions"`
+	UDPBind                  string              `toml:"udp-bind"`
+	UDPBuffer                int                 `toml:"udp-buffer"`
+	UDPReadBuffer            int                 `toml:"udp-read-buffer"`
+	StartUpTimeout           toml.Duration       `toml:"startup-timeout"`
+	SubscriptionSyncInterval toml.Duration       `toml:"subscriptions-retry-interval"`
 }
 
 func NewConfig() Config {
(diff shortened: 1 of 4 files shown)


services/influxdb/service.go
```diff
@@ -187,7 +189,14 @@ type subInfo struct {
...
 	if !s.disableSubs {
-		return s.linkSubscriptions()
+		err := s.linkSubscriptions()
+		ticker := time.NewTicker(s.subscriptionsRetryInterval)
+		go func() {
+			for _ = range ticker.C {
+				s.linkSubscriptions()
+			}
+		}()
+		return err
```
nathanielc: Can you add the matching logic to stop the ticker in the `Close` method?

nathanielc:
@titilambert Overall looks good. A few comments

1.  I think we should rename the variable SubscriptionRetryInterval to SubscriptionSyncInterval since its not retrying a failed task but rather polling for new information. 
2. We should check if the interval is 0 thus disabling subscription syncing. 
3. The current implementation of `linkSubscriptions` assumes that no services are running. But that is no longer the case, we need to update `linkSubscriptions` to keep track of which subs are already running and which are not.

nathanielc:
Here at this line https://github.com/titilambert/kapacitor/blob/c673874f60cfa5123bca73b046f5da419468fb03/services/influxdb/service.go#L417

We need to check if the UDP service is already running before starting. A map[subEntry]bool on the `influxdb` struct should work.

titilambert:
@nathanielc I fixed comment 1 and 2
About comment 3: I'm not sure to understand what is purpose of `startedSubs` variable

titilambert:
@nathanielc I made a new commit which should include all your comments

nathanielc:
@titilambert Looking good, I think you also need to mark the sub as running [here](https://github.com/titilambert/kapacitor/blob/41926fbdd3324c1fed9359f7cc04e4fb22bea917/services/influxdb/service.go#L409) too. The difference between those two loops is whether the sub has been created or not. We could probably simplify the loops a bit, but for now just keeping track in both loops should work.

titilambert:
@nathanielc Done !

services/influxdb/service.go
```diff
@@ -144,22 +145,25 @@ func (s *Service) NewNamedClient(name string) (client.Client, error) {
...
+	httpPort                 int
+	logger                   *log.Logger
+	protocol                 string
+	udpBind                  string
+	udpBuffer                int
+	udpReadBuffer            int
+	startupTimeout           time.Duration
+	subscriptionSyncInterval time.Duration
+	disableSubs              bool
+	runningSubs              map[subEntry]bool
```
nathanielc: This map needs to be initialized or you get a `panic: assignment to entry in nil map`.

services/influxdb/service.go
```diff
@@ -187,13 +191,25 @@ type subInfo struct {
...
+			}()
+		}
+		return err
 	}
 	return nil
 }
 
 func (s *influxdb) Close() error {
 	var lastErr error
+	if s.subscriptionSyncInterval != time.Duration(0) {
```
nathanielc: I prefer `if s.subscriptionSyncInterval != 0` as there is less to read and as a result reads cleaner.

services/influxdb/service.go
```diff
@@ -187,13 +191,25 @@ type subInfo struct {
 
 func (s *influxdb) Open() error {
 	if !s.disableSubs {
-		return s.linkSubscriptions()
+		err := s.linkSubscriptions()
+		if s.subscriptionSyncInterval != time.Duration(0) {
```
nathanielc: Here too...

nathanielc:
@titilambert Just a few small fixes but overall looks ready. Once those things are fixed I'll test locally.

titilambert:
@nathanielc done.
I will squash it when you want.

titilambert:
@nathanielc I got a timeout on circle. I can not see where restart the build. Could you do that ?

services/influxdb/config.go
```diff
@@ -34,15 +35,16 @@ type Config struct {
...
+	Timeout                  toml.Duration       `toml:"timeout"`
+	DisableSubscriptions     bool                `toml:"disable-subscriptions"`
+	SubscriptionProtocol     string              `toml:"subscription-protocol"`
+	Subscriptions            map[string][]string `toml:"subscriptions"`
+	ExcludedSubscriptions    map[string][]string `toml:"excluded-subscriptions"`
+	UDPBind                  string              `toml:"udp-bind"`
+	UDPBuffer                int                 `toml:"udp-buffer"`
+	UDPReadBuffer            int                 `toml:"udp-read-buffer"`
+	StartUpTimeout           toml.Duration       `toml:"startup-timeout"`
+	SubscriptionSyncInterval toml.Duration       `toml:"subscriptions-retry-interval"`
```
nathanielc: The config options still says `retry`.

services/influxdb/config.go
```diff
@@ -73,6 +76,9 @@ func (c *Config) SetDefaultValues() {
 	if c.SubscriptionProtocol == "" {
 		c.SubscriptionProtocol = DefaultSubscriptionProtocol
 	}
+	if c.SubscriptionSyncInterval == toml.Duration(0) {
```
nathanielc: And this one too.

nathanielc:
@titilambert I think we are close:

1. Can you add a CHANGELOG entry under the 0.14 features section?
2. Can you update the example config in etc/kapacitor/kapacitor.conf with a simple explanation about the new config setting
3. Can you change [this](https://github.com/titilambert/kapacitor/blob/23cc44d6107dcb21a4c5d1c1a04f0d9d00dea41c/services/influxdb/service.go#L250) log to a `D!` debug log since its a much more frequent log.

Testing locally its working great. Thanks!

titilambert:
@nathanielc BTW 1m0s as default value is OK for you ?

nathanielc:
@titilambert Yes, 1m seems like a reasonable default.

nathanielc:
@titilambert Thanks! I'll merge on green.

titilambert:
Thanks ! I will squash it just before ;)

nathanielc:
@titilambert Github can now squash automatically before a merge ;) I got this. Thanks again!

titilambert:
Cool ! Thanks !

PR #606, opened 2016-06-03
Add HoltWinters method

nathanielc:
Add Holt Winters InfluxQL function

For example one can now define an alert based off forecasted future values.

The following TICKscript will take the last 30 days of disk usage stats and using holt-winters forecast the next 7 days. If the forecasted value crosses a threshold an alert is triggered.

The result is now Kapacitor will alert you 7 days in advance of a disk filling up. This assumes a slow growth but by changing the vars in the script you could check for shorter growth intervals.

```go
// The interval on which to aggregate the disk usage
var agg_interval = 1d
// The number of `agg_interval`s to forecast into the future
var forecast_count = 7
// The amount of historical data to use for the fit
var history = 30d

// The critical threshold on used_percent
var threshold = 90.0

batch
    |query('''
    SELECT max(used_percent) as used_percent
    FROM "telegraf"."default"."disk"
''')
        .period(history)
        .every(agg_interval)
        .align()
        .groupBy(time(agg_interval), *)
    |holtWinters('used_percent', forecast_count, 0, agg_interval)
        .as('used_percent')
    |max('used_percent')
        .as('used_percent')
    |alert()
         // Trigger alert if the forecasted disk usage is greater than threshold
        .crit(lambda: "used_percent" > threshold)
        .slack()
```
- [x] Rebased/mergable
- [x] Tests pass
- [x] CHANGELOG.md updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -110,6 +110,47 @@ On startup Kapacitor will detect the change and recreate the subscriptions in In
 >NOTE: While HTTP itself is a TCP transport such that packet loss shouldn't be an issue, if Kapacitor starts to slow down for whatever reason, InfluxDB will drop the subscription writes to Kapacitor.
 In order to know if subscription writes are being dropped you should monitor the measurement `_internal.monitor.subscriber` for the field `writeFailures`.
 
+#### Holt-Winters Forecasting
+
+This release contains an new Holt Winters InfluxQL function.
+
+With this forecasting method one can now define an alert based off forecasted future values.
+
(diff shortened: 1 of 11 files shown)


elvarb:
Is holtWinters out of the project? it does not seem to be available nor in the documentation except for version 0.10

nathanielc:
@elvarb Its part of the 1.0 series. https://docs.influxdata.com/kapacitor/v1.0/nodes/influx_q_l_node/#holtwinters

elvarb:
Ah, all makes sense now. Thanks :+1:

PR #610, opened 2016-06-03
Expire groups after X number of zero events

nathanielc:
Expire a group after X number of stats points are received with zero value.

This allows for deaman groups to expire after giving an alert.

TODO
- [ ]  Test inverted deadman, aka alert on high counts
- [ ] Be able to disable expiring alerts for deadman method
- [x] Rebased/mergable
- [ ] Tests pass
- [x] CHANGELOG.md updated

Some thoughts:

Maybe we want to extend this kind of logic beyond just the StatsNode. All nodes would benefit from being able to free up resources for expired groups. There should probably be two ways to expire groups.
1. Implicitly via some sort of timeout or zero count (like the PR is already doing)
2. Explicitly via an API call, so it can be managed externally to Kapacitor. (i.e. shutdown hook when a server is being retired)


nathanielc:
Closing since this is no longer a valid solution. See #1157

PR #611, opened 2016-06-03
Add bash completetion, fix a few oddities with the cli

nathanielc:
Adds bash completion to the `kapacitor` cli command.

- [x] Rebased/mergable
- [x] Tests pass
- [x] CHANGELOG.md updated
- [x] Update build.py to package the bash_completion script


nathanielc:
@rossmcdonald Could you test this out when you get a sec? It seems to work well on my host but I'd like it to get a little more exposure before merging. I don't trust my ability to write portable bash ;)

PR #629, opened 2016-06-09
Bash auto completion: Path fixes

bbczeuz:
I was annoyed of typing long lines at the CLI, so I started to implement a bash auto completion script for kapacitor, only to find that there is already one, well hidden in the git repo (of course I figured that out *after* I finished my own script). It seems not to be installed by the current .rpm package. Anyhow. I now base on the existing script with improved file name auto completion for ```-tick``` and ```-vars``` parameters of ```kapacitor define```.

- [x] Rebased/mergable
- [x] Tests pass
- [ ] CHANGELOG.md updated

###### Comments

- How shall I update the changelog? This pull request does not yet have a number.
- At $packager: The bash auto completion script seems not to be installed by default

### DIFF
--- a/usr/share/bash-completion/completions/kapacitor
+++ b/usr/share/bash-completion/completions/kapacitor
@@ -1,3 +1,10 @@
+__kapacitor_filter_ticks()
+{
+#$1=cmd
+#$2=word
+#$3=prev
+	return 0
+}
 
 # Bash completion for the kapacitor command.
 _kapacitor()
@@ -49,7 +56,10 @@ _kapacitor()
                         words=$(_kapacitor_dbrps)
                         ;;
                     -tick)
-                        COMPREPLY=($(compgen -o filenames -A file -X '!*.tick' -- "$cur"))
+                        COMPREPLY=($(compgen -o filenames -A file -X '!*.tick' -S ' ' -- "$cur") )
+                        didi=$(compgen -d -S '/' -- "$cur")
+                        COMPREPLY=(${COMPREPLY[@]:-} $didi )
+                        compopt -o nospace
                         ;;
                     -type)
                         words='batch stream'
--- a/usr/share/bash-completion/completions/kapacitor
+++ b/usr/share/bash-completion/completions/kapacitor
@@ -1,11 +1,3 @@
-__kapacitor_filter_ticks()
-{
-#$1=cmd
-#$2=word
-#$3=prev
-	return 0
-}
-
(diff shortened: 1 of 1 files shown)


usr/share/bash-completion/completions/kapacitor
```diff
@@ -49,7 +48,7 @@ _kapacitor()
                         words=$(_kapacitor_dbrps)
                         ;;
                     -tick)
-                        COMPREPLY=($(compgen -o filenames -A file -X '!*.tick' -- "$cur"))
+			_kapacitor_tick_files "${cur}"
```
nathanielc: The -tick is also part of the `define-template` section. Can you add this fix there too?

usr/share/bash-completion/completions/kapacitor
```diff
@@ -165,6 +164,38 @@ _kapacitor()
...
+	#Add a slash (directories) or space (files except directories) after name
+	local c i=0
+	for c in "${COMPREPLY[@]}"; do
+		if [ -d "${c}" ]; then
+			COMPREPLY[i++]="${c}/";
+		else
+			COMPREPLY[i++]="${c} ";
+		fi
+	done
+        compopt -o nospace
```
nathanielc: Isn't this the default? What is this doing? This is my first bash completion script so I don't really know what I am talking about..Thanks for improving it ;)

bbczeuz: The original script didn't support directory auto completion. So one had to adapt the compgen part to get filenames and directories. Per default, bash adds a space, so tabbing though a directory tree wouldn't work. So I disabled this via 'nospace' and manually add spaces and slashes.

nathanielc: OK, so why is it after the loop? Looks like its only set after compgen has already been called.

usr/share/bash-completion/completions/kapacitor
bbczeuz: Done

usr/share/bash-completion/completions/kapacitor
bbczeuz: It just happened to be there. It must be somewhere to avoid space trailing, but it could be placed before the loop, too.

nathanielc:
@bbczeuz Look good to go. I had an upstream build failure which is causing your PR to fail. Can you rebase off lastest master, that should fix it? I'll merge on green.

PR #630, opened 2016-06-09
Update proto for udf

nathanielc:
protoc-gen-go updated and we need a new udf.pb.go file to match.

Need to figure out a simple way of locking the version of protoc-gen-go that we use. Should I just vendor it?


nathanielc:
I didn't like where that was going... :(

PR #632, opened 2016-06-09
Change to manual vendored deps

nathanielc:
The protoc-gen-go code updated upstream and it broke builds again. So I went hunting for a simple go vendor depedency manager. Nothing fit my needs to I have switched to simple vendor based management via git-subrepo.

@rossmcdonald FYI here are some changes to build.py, but they are all specific to Kapacitor.


rossmcdonald:
`build.py` changes look good. And I like the new log message:

```
Pretending to retrieve Go dependencies...
```

😄

PR #644, opened 2016-06-13
JIRA plugin for kapacitor

ljagiello:
- [X] Rebased/mergable
- [X] Tests pass
- [X] CHANGELOG.md updated

The general idea is to send alert to JIRA and create a ticket or change a ticket priority based on alert state.

It's still in progress but I have a question ( probably @nathanielc can answer ).
* I used third-party packages (github.com/andygrunwald/go-jira) should I rewrite that based on standard libs inside that plugin?

### DIFF
--- a/pipeline/alert.go
+++ b/pipeline/alert.go
@@ -753,9 +753,9 @@ type HipChatHandler struct {
 //      username = ""
 //      password = ""
 //      project = ""
-//      issue_type = ""
-//      priority_warn = ""
-//      priority_crit = ""
+//      issue-type = ""
+//      priority-warn = ""
+//      priority-crit = ""
 //      global = false
 //
 // Example:
@@ -775,9 +775,9 @@ type HipChatHandler struct {
 //      username = "test-username"
 //      password = "test-password"
 //      project = "TEST"
-//      issue_type = "Bug"
-//      priority_warn = "Normal"
-//      priority_crit = "Urgent"
+//      issue-type = "Bug"
+//      priority-warn = "Normal"
+//      priority-crit = "Urgent"
 //      global = true
 //
 // Example:
@@ -792,9 +792,9 @@ type HipChatHandler struct {
 //         |alert()
 //             .jira()
 //             .project('TEST2')
-//             .issue_type('Bug2')
-//             .priority_warn('Normal2')
-//             .priority_crit('Urgent2')
+//             .issueType('Bug2')
+//             .priorityWarn('Normal2')
+//             .priorityCrit('Urgent2')
 //
 // Send alerts to JIRA where:
 //  - project: 'TEST2'
@@ -821,15 +821,15 @@ type JiraHandler struct {
 
 	// JIRA issue type in which to post messages.
 	// If empty uses the issue type from the configuration.
-	Issue_type string
+	IssueType string
 
 	// JIRA warning priority in which to post messages.
 	// If empty uses the warning priority from the configuration.
-	Priority_warn string
+	PriorityWarn string
 
 	// JIRA critical priority in which to post messages.
 	// If empty uses the critical priority from the configuration.
-	Priority_crit string
+	PriorityCrit string
 }
 
 // Send the alert to Alerta.
--- a/pipeline/alert.go
+++ b/pipeline/alert.go
@@ -28,7 +28,7 @@ const defaultLogFileMode = 0600
 // See AlertNode.Info, AlertNode.Warn, and AlertNode.Crit below.
 //
 // Different event handlers can be configured for each AlertNode.
-// Some handlers like Email, HipChat, Sensu, Slack, OpsGenie, VictorOps, PagerDuty and Talk have a configuration
+// Some handlers like Email, HipChat, Sensu, Slack, OpsGenie, VictorOps, Jira, PagerDuty and Talk have a configuration
 // option 'global' that indicates that all alerts implicitly use the handler.
 //
 // Available event handlers:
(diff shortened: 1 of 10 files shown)


nathanielc:
> I used third-party packages (github.com/andygrunwald/go-jira) should I rewrite that based on standard libs inside that plugin?

I would prefer to use the the standard libs like the `net/http` package to communicate with the JIRA REST API. If that represents a significant amount of work we can reconsider but if all Kapacitor is doing it creating and editing issues it should be simple enough to do directly.

pipeline/alert.go
```diff
@@ -733,6 +737,65 @@ type HipChatHandler struct {
 	Token string
 }
 
+// Send the alert to JIRA.
```
nathanielc: Does anything need to be done on the JIRA side of things to enable this plugin?

If so can you add some steps on how to do so here?

pipeline/alert.go
```diff
@@ -733,6 +737,65 @@ type HipChatHandler struct {
...
+	}
+	a.JiraHandlers = append(a.JiraHandlers, jira)
+	return jira
+}
+
+// tick:embedded:AlertNode.Jira
+type JiraHandler struct {
+	*AlertNode
+
+	Project string
```
nathanielc: I think the issue_type, priority_warn, and priority_crit should all be configurable per alert, and if its empty then use the value from the config.

pipeline/alert.go
```diff
@@ -733,6 +737,65 @@ type HipChatHandler struct {
...
+//
+// Example:
+//    [jira]
+//      enabled = true
+//      url = "https://your.jira.host"
+//      username = "test-username"
+//      password = "test-password"
+//      project = "TEST"
+//      issue_type = "Bug"
+//      priority_warn = "Normal"
```
nathanielc: Do we need priorities for the OK, and INFO alert levels too?

nathanielc:
@ljagiello I took a quick look and added some comments. Overall it look great! Thanks for the PR.

pipeline/alert.go
ljagiello: Fixed with https://github.com/influxdata/kapacitor/pull/644/commits/22854adb636f6bc51a7a22b760ce06d517973376. I think that should be clear enough.

pipeline/alert.go
ljagiello: For now I would leave only WARNING/CRITICAL. It's hard for me to imagine reason why JIRA ticket should be created in case of priority OK/INFO.

ljagiello:
@nathanielc I think all issues were fixed, could you please code review one more time.

nathanielc:
@ljagiello Thanks for the update. I took another look and have a few questions:

1. Can you change variable names to match the camel case style, (i.e issue_type -> issueType)? Also all toml keys should use `-` instead of `_`.
2. How does an issue get resolved? I saw the switch statement on the alert level but no case for when the Alert goes back to the OK state. Does close the issue or change its priority to something low?

Thanks for removing the third party package. Looks like it was pretty straightforward to call their REST API.

ljagiello:
@nathanielc Thanks for questions. 
* Camel case style - fixed.
* Current code doesn't handle `OK` state. I will add that part.

ljagiello:
@nathanielc It should supports now: resolve and alert level changes. Tests are not really great for that and I don't really know how to make them better :(

services/jira/service.go
```diff
@@ -0,0 +1,239 @@
...
+	escapeKeys := [...]string{"[", "]", " "}
+	for _, v := range escapeKeys {
+		searchSummary = strings.Replace(searchSummary, v, "\\\\"+v, -1)
+	}
+	searchData["jql"] = "project = " + project + " AND summary ~ " + searchSummary + " AND status != " + s.issueFinalStatus
+
+	resp, err := s.RequestPost("POST", "/rest/api/2/search", searchData)
+	body, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		panic(err.Error())
```
nathanielc: Why panic instead of returning the error?

ljagiello: I think i left those from some previous debugging. Fixed - https://github.com/influxdata/kapacitor/pull/644/commits/2125e71d259c1af35af6d544ea52a9248353be74

nathanielc:
@ljagiello Is there a way to use the `ID` to uniquely identify the issue so we do not have to search existing issues? Searching issues seems fragile as any number of project specific customization could return invalid results. 

Also to make sure I understand the current implementation...

* If Kapacitor get a new alert it creates a new JIRA issue
* If Kapacitor changes the state of an alert it changes the issue priority.
* If the alert is resolved then the issue is transitioned to a final state.

Are all those statements correct?

ljagiello:
@nathanielc Unfortunately `ID` is not good enough to find correct ticket (modify/close issue). JIRA REST API requires: `issueID` or `Key` (https://docs.atlassian.com/jira/REST/latest/#api/2/issue-doTransition) - both are unknown when you have only information from kapacitor and both are autogenerated when issue is created (you cannot really predict that) :(

All statements you've made are correct. That's all what plugin is doing.

nathanielc:
I am concerned that this integration is not going to be robust enough for general use. 

Specifically the searching aspect and the mapping of Kapacitor events to both transitions and priority changes seems too specific to a JIRA workflow. JIRA can be used in many ways and there doesn't seem to be a clean mapping from Kapacitor alerts to JIRA issue workflows.  

 This still leaves the option to integrate with JIRA via the AlertNode [exec](https://docs.influxdata.com/kapacitor/v0.13/nodes/alert_node/#exec) handler, for which most of this code can be re-purposed.

nathanielc:
Closing for now since there hasn't been anymore activity.

PR #680, opened 2016-06-27
Telegram Alerting

burdandrei:
- [x] Rebased/mergable
- [x] Tests pass
- [x] CHANGELOG.md updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@
 ### Features
 
 - [#662](https://github.com/influxdata/kapacitor/pull/662): Add `-skipVerify` flag to `kapacitor` CLI tool to skip SSL verification.
+- [#680](https://github.com/influxdata/kapacitor/pull/680): Add Telegram Alerting option
+
 
 ### Bugfixes
 
--- a/integrations/streamer_test.go
+++ b/integrations/streamer_test.go
@@ -31,6 +31,7 @@ import (
 	"github.com/influxdata/kapacitor/services/sensu"
 	"github.com/influxdata/kapacitor/services/slack"
 	"github.com/influxdata/kapacitor/services/talk"
+	"github.com/influxdata/kapacitor/services/telegram"
 	"github.com/influxdata/kapacitor/services/victorops"
 	"github.com/influxdata/kapacitor/udf"
 	"github.com/influxdata/kapacitor/udf/test"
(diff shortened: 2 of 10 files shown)


services/telegram/service.go
```diff
@@ -0,0 +1,81 @@
...
+		return err
+	}
+	defer resp.Body.Close()
+	if resp.StatusCode != http.StatusOK {
+		body, err := ioutil.ReadAll(resp.Body)
+		if err != nil {
+			return err
+		}
+		type response struct {
+			Error string `json:"error"`
```
nathanielc: As per the docs here is looks like errors are of the form 

```json
{
  "ok" : false,
  "description" : "error message here",
  "error_code": 576
}
```

Can you update the response struct accordingly?

services/telegram/service.go
```diff
@@ -0,0 +1,81 @@
...
+func (s *Service) Global() bool {
+	return s.global
+}
+func (s *Service) StateChangesOnly() bool {
+	return s.stateChangesOnly
+}
+
+func (s *Service) Alert(chatId, message string) error {
+	if chatId == "" {
+		chatId = s.chat_id
```
nathanielc: Can you use camel case for all vars i.e. `s.chatId`?

services/telegram/service.go
```diff
@@ -0,0 +1,81 @@
...
+	}
+	postData := make(map[string]interface{})
+	postData["chat_id"] = chatId
+	postData["text"] = message
```
nathanielc: From the docs it looks like the json body needs to be on the form 

```json
{
   "message": {
      "chat_id" : "asdf",
      "text": "CRITICAL ..."
   }
}
```

 I could be reading it wrong. 

On a related note, Is there an easy way I can test against a  Telegram server? Something like a docker container or public testing endpoint?

nathanielc:
@burdandrei Thanks for the PR! Overall things look to be in place. I have made a few comments here and there.

burdandrei:
@nathanielc  thanks for the review,
Error responce and camel case i'll do, missed it with all the copy-pasting from other services ;)
about the message text, i can [see](https://docs.influxdata.com/kapacitor/v0.13//nodes/alert_node/#message) that there is Level in default Message, so i think leaving it as is, OK?

services/telegram/service.go
burdandrei: They have [no server](https://telegram.org/faq#q-can-i-run-telegram-using-my-own-server) available in open source, only clients. So I'm afraid we can only can mock it in tests.

nathanielc:
@burdandrei Sounds good. As for the message text you are right that the level can be added to the message and is there by default. The part the seemed incorrect was the structure of the JSON post data itself. Specifically that their docs imply that it needs to be wrapped in an object under the `message` field. They don't provide examples and I haven't found a way to test locally so I am not sure. Does it work for you?

CHANGELOG.md
```diff
@@ -18,6 +18,7 @@
 
 - [#636](https://github.com/influxdata/kapacitor/pull/636): Change HTTP logs to be in Common Log format.
 - [#652](https://github.com/influxdata/kapacitor/pull/652): Add optional replay ID to the task API so that you can get information about a task inside a running replay.
+- [#680](https://github.com/influxdata/kapacitor/pull/680): Add Telegram Alerting option
```
nathanielc: Also, can you rebase off latest master and move this line to the beta3 section since beta2 has now been released?

burdandrei:
@nathanielc according to the [docs](https://core.telegram.org/bots/api#sendmessage) Only required fields are chat_id and text, so I'm just pushing message to text and that's it.

CHANGELOG.md
burdandrei: Sure, you guys are fast ;)

burdandrei:
@nathanielc Take a look now

pipeline/alert.go
```diff
@@ -936,6 +941,91 @@ type SlackHandler struct {
...
+	ParseMode string
+	// Web Page preview
+	// If empty uses the disable-web-page-preview from the configuration.
+	IsDisableWebPagePreview bool `tick:"DisableWebPagePreview"`
+	// Disables Notification
+	// If empty uses the disable-notification from the configuration.
+	IsDisableNotification bool `tick:"DisableNotification"`
+}
+
+func (tel *TelegramHandler) DisableNotification() *TelegramHandler {
```
nathanielc: Isn't the default true? How would you set this to false?

services/telegram/service.go
```diff
@@ -0,0 +1,101 @@
...
+}
+
+func (s *Service) Alert(chatId, parseMode, message string, disableWebPagePreview, disableNotification bool) error {
+	if chatId == "" {
+		chatId = s.chatId
+	}
+
+	if parseMode == "" {
+		parseMode = s.parseMode
+	}
```
nathanielc: Can we validate parse mode is one of `Markdown` or `HTML`?

services/telegram/service.go
```diff
@@ -0,0 +1,101 @@
...
+
+	postData := make(map[string]interface{})
+	postData["chat_id"] = chatId
+	postData["text"] = message
+	postData["parse_mode"] = parseMode
+	postData["disable_web_page_preview"] = disableWebPagePreview
```
nathanielc: Shouldn't these booleans default to the config option?

pipeline/alert.go
burdandrei: You are right, it's default false, changed it

services/telegram/service.go
burdandrei: I'm a bit confused how to check it, cause they can't be nil, and if they are not passed in tick, they are false. Maybe this is OK?

burdandrei:
@nathanielc  i added validation of parseMode and the checks of disableNotification and disableWebPagePreview, to take the defaults from the config.
Let me know what you think  =)

services/telegram/service.go
```diff
@@ -0,0 +1,113 @@
...
+	if chatId == "" {
+		chatId = s.chatId
+	}
+
+	if parseMode == "" {
+		parseMode = s.parseMode
+	}
+
+	if parseMode != "Markdown" && parseMode != "HTML" {
+		return errors.New(fmt.Sprintf("parseMode %s is not valid, please use 'Markdown' or 'HTML'", parseMode))
```
nathanielc: You can use fmt.Errorf here

```
fmt.Errorf("parseMode %s is not valid, please use 'Markdown' or 'HTML'", parseMode)
```

pipeline/alert.go
nathanielc: Can you add a comment to these functions?

pipeline/alert.go
```diff
@@ -936,6 +941,91 @@ type SlackHandler struct {
...
+
+	// Telegram user/group ID to post messages to.
+	// If empty uses the chati-d from the configuration.
+	ChatId string
+	// Parse node, defaults to Mardown
+	// If empty uses the parse-mode from the configuration.
+	ParseMode string
+	// Web Page preview
+	// If empty uses the disable-web-page-preview from the configuration.
+	IsDisableWebPagePreview bool `tick:"DisableWebPagePreview"`
```
nathanielc: Can you add a `// tick:ignore` comment line to the `IsDisabled*` fields so that the auto gen docs skips these place holder fields?

For example

```go
// Web Page preview
// If empty uses the disable-web-page-preview from the configuration.
// tick:ignore
IsDisableWebPagePreview bool `tick:"DisableWebPagePreview"`
```

nathanielc:
@burdandrei Looks good just a few small changes left...

integrations/streamer_test.go
```diff
@@ -3793,6 +3794,85 @@ stream
...
+	|count('value')
+	|alert()
+		.id('kapacitor/{{ .Name }}/{{ index .Tags "host" }}')
+		.info(lambda: "count" > 6.0)
+		.warn(lambda: "count" > 7.0)
+		.crit(lambda: "count" > 8.0)
+		.telegram()
+		.chatId('12345678')
+                .disableNotification()
+                .parseMode('HTML')
```
nathanielc: Can you add another `.telegram` call so we can show that it works with multiple handlers configured?

task_master.go
```diff
@@ -82,6 +82,11 @@ type TaskMaster struct {
 		StateChangesOnly() bool
 		Alert(channel, message string, level AlertLevel) error
 	}
+	TelegramService interface {
+		Global() bool
+		StateChangesOnly() bool
+		Alert(chatId, parseMode, message string, disableWebPagePreview, disableNotification bool) error
+	}
```
nathanielc: Can you also add the TelegramService to the `New` method below so cloned task master preserve this service?

I always forget this one ;)

burdandrei:
@nethanielc Done, thanks for great tips!

nathanielc:
@burdandrei Thanks for the PR!

PR #702, opened 2016-07-06
Initial auth plumbing

nathanielc:
Adds basic plumbing for authenticating and authorizing requests.

The basic design has two kinds of user actions that can be controlled.

1. API calls are controlled by their HTTP method and resource path.
2. The /write endpoint further checks permissions for write access to the database.

- [x] Rebased/mergable
- [x] Tests pass
- [x] CHANGELOG.md updated
- [x] Vendor jwt dep to version 3.0.0
- [x] Add tests to auth package.


auth/auth.go
```diff
@@ -0,0 +1,157 @@
...
+	case "GET":
+		return ReadPrivilege, nil
+	case "POST", "PATCH":
+		return WritePrivilege, nil
+	case "DELETE":
+		return DeletePrivilege, nil
+	default:
+		return AllPrivileges, fmt.Errorf("unknown method %q", m)
+	}
+}
```
dgnorton: Wondering if authorizing the HTTP request is the right way to go or if it should be somewhere below the transport layer. Thoughts?

auth/auth.go
```diff
@@ -0,0 +1,157 @@
...
+	"fmt"
+	"path"
+	"strings"
+
+	"github.com/influxdata/influxdb/influxql"
+	"github.com/pkg/errors"
+)
+
+// Interface for authenticating and retrieving users.
+type Interface interface {
```
dgnorton: Not crazy about using `Interface` as the interface name. Reads okay in the code but feels weird. Also, if the `auth` package ever needed a second interface for something, it would end up with a regular name, which would be inconsistent.

PR #735, opened 2016-07-21
Fix missing dot causes confusing sentence in AlertNode doc

minhdanh:
The missing dot causes the generated doc to include a confusing sentence:

screenshot from 2016-07-21 16 57 30


nathanielc:
@minhdanh Thanks!

Issue #614, opened 2016-06-04
alerts: threshold values for lowering alert levels


melor:
## Currently

Alert is generated at every alert level change when going up info -> warn -> critical and when coming down critical -> warn -> info -> ok.

## Problem

When monitoring for example free disk space level, the value often hovers both below and above the threshold values, causing e.g. a warn -> ok -> warn -> ok -> warn ... cycle. Flapping() percentages cause delays and are not optimal for every purpose.

Problem example:

```
stream
    |from()
        .measurement('disk')
    |alert()
        // values 20.0,19.99,20.0,19.99,... each cause an alert level change 
        .warn(lambda: "free" < 20)  
```

## Proposed solution

provide optional, separate threshold functions for resetting a higher severity to a lower one. 

Example:

```
stream
    |from()
        .measurement('disk')
    |alert()
        .warn(lambda: "free" < 20)
        .warn_reset(lambda: "free" > 25)
        .crit(lambda: "free" < 10)
        .crit_reset(lambda: "free" > 15)
```

In this example the value, once going below 20 and causing the initial "WARNING" alert, would be allowed to fluctuate between 10 and 25 without changing the alert level from warning.

minhdanh:
I'm experiencing the same problem. This would be a very useful feature to be implemented to avoid continuous alerts on changes hover up and down the threshold values.

minhdanh:
@rossmcdonald I would like to make a PR for this feature. Can you please tell me generally where should I begin? Which files should be involved?

melor:
Perhaps the syntax could simply be with an optional extra argument to the alert level functions, e.g. .warn(<alert criteria>, [<alert reset criteria>])

Example:

```
stream
    |from()
        .measurement('disk')
    |alert()
        .warn(lambda: "free" < 20, lambda: "free" > 25)
        .crit(lambda: "free" < 10, lambda: "free" > 15)
```

nathanielc:
@melor  I think using `warn_reset` is more explicit and cleaner to implement internally.

@minhdanh Thanks for stepping up.

There is a function `determineLevel` that is responsible for determining the alert level of a data point. It probably needs another arg for the current level...

It can be found here:
https://github.com/influxdata/kapacitor/blob/master/alert.go#L554

Then in the `pipeline/alert.go` file you will need to field for the *Reset versions of the Alert node:
https://github.com/influxdata/kapacitor/blob/master/pipeline/alert.go#L204

[...]

PR #740, opened 2016-07-22
Resetting alert levels

minhdanh:
In regard to this issue https://github.com/influxdata/kapacitor/issues/614
This pull request should enable lowering alert levels using lamda expressions.
Only WARN and CRIT levels are supported, INFO level is not.

I'm still trying to come up with some tests for this.

- [x] Rebased/mergable
- [x] Tests pass
- [x] CHANGELOG.md updated


minhdanh:
This is what I think of, however I believe it needs more improvements. Will make some more changes in the next days : )

nathanielc:
@minhdanh Looks like a good start. FYI #732 has been merged so you can rebase your PR to see only your changes.

alert.go
```diff
@@ -116,6 +116,8 @@ type AlertNode struct {
 	critsTriggered  *expvar.Int
 
 	bufPool sync.Pool
+
+	level_resets []stateful.Expression
```
nathanielc: Go is an opinionated language and prefers camelCase names. Can you change this to `levelResets`?

pipeline/alert.go
```diff
@@ -209,6 +206,11 @@ type AlertNode struct {
 	// An empty value indicates the level is invalid and is skipped.
 	Crit *ast.LambdaNode
 
+	// Filter expression for reseting the WARNING alert level to OK.
```
nathanielc: Doesn't a reset set it to a different level not necessarily the OK state? Meaning a CRITICAL remains CRITICAL until the reset condition is false, then the level is determined again using the previous behavior. Which could be WARNING, INFO or OK.

pipeline/alert.go
```diff
@@ -209,6 +206,11 @@ type AlertNode struct {
...
+	Warn_reset *ast.LambdaNode
```
nathanielc: Again can we name these WarnReset etc? Which in TICKscript will read `.warnReset`.

pipeline/alert.go
```diff
@@ -209,6 +206,11 @@ type AlertNode struct {
...
+	// Filter expression for reseting the CRITICAL alert level to OK.
+	Crit_reset *ast.LambdaNode
```
nathanielc: Do we also want an InfoReset?

alert.go
minhdanh: Yup. I've changed the variables to camelCase format. : )

pipeline/alert.go
minhdanh: That makes sense. This makes it easier to implement the feature.

alert.go
```diff
@@ -336,6 +354,13 @@ func newAlertNode(et *ExecutingTask, n *pipeline.AlertNode, l *log.Logger) (an *
 		}
 		an.levels[CritAlert] = statefulExpression
 		an.scopePools[CritAlert] = stateful.NewScopePool(stateful.FindReferenceVariables(n.Crit.Expression))
```
nathanielc: You will need separate scopePools for the reset expressions.

alert.go
```diff
@@ -563,6 +588,18 @@ func (a *AlertNode) determineLevel(now time.Time, fields models.Fields, tags map
 			continue
 		} else if pass {
 			return AlertLevel(l)
+		} else {
+			lrse := a.levelResets[l]
```
nathanielc: Doesn't the determineLevel function need to know the current level? This logic doesn't seem right but maybe I am misunderstanding the purpose of the reset expressions.

My thinking was that if an alert enters a state with a reset expression it can't leave that state until the reset function is false. Meaning the determineLevel would check for any reset expressions for the current level first and only evaluate other expressions if the reset expression is false.

Maybe it would help to write up the tests at this point so we can agree on the expected behavior before we dig into the implementation details too far.

alert.go
minhdanh: Yeah, at first I thought the reset expression is kind of "opposite" to the normal expression, so it may use the exact fields and tags from the scope pool of the normal expression. I'll have this updated.

alert.go
minhdanh: I get you. My idea was that when a normal expression returns false, a data point should be evaluated for lower levels only if a reset expression doesn't exists, or if it exists and return true.
I'll try to come up with the tests. Feel free to point out where I get it wrong. : )

minhdanh: God, that logic is terrible. Your thinking is much better. I need to write the tests now.

minhdanh:
@nathanielc I've just pushed a test with its sample data. Please have a look at them to see it's is the expected behavior of this feature.
There's one thing confuses me, as you said:
``` 
Meaning the determineLevel would check for any reset expressions for the current level first and only evaluate other expressions if the reset expression is false.
```
Why would it only evaluate other expressions if the reset expression is false? I mean why not true?
For example I have the following rules in a TICK script:
```
.info(lambda: "value" > 40)
.infoReset(lambda: "value" < 10)
.warn(lambda: "value" > 60)
.warnReset(lambda: "value" < 20)
.crit(lambda: "value" > 80)
.critReset(lambda: "value" < 30)
```
Say first we have a data point with a value of `70`, that will cause the alert level to be WARNING. Next if we have a data point with a value of `15`, the reset expression of the WARNING level will be evaluated. It would return true (15 < 20 = true). Here we'll want the alert to escape the WARNING state so that the other expressions can be evaluated.
Or let's say we have a value of `40` instead of `15`. It would cause the reset evaluation to return false (40 < 20 = false). At this point wouldn't we want the `determineLevel` function to return the current level which is WARNING?

Let's examine another data point with a value of `85` instead of `15` in the above case. Now the reset expression of the WARNING level would return false (85 < 20 = false). Here if we return the current level (WARNING), it will be incorrect as the level now should be CRITICAL, meaning the crit expression should be evaluated as well. So I think in case a reset expression returns false, we would check if the most highest level possible is greater than the current level. If it is, we return that level instead of the current level. If it's not, we just return the current level.

nathanielc:
@minhdanh Thanks for the detailed example. I think we are on the same page. Your example makes sense to me, (I probably inverted the logic earlier when I was talking about the true/false value of the reset expression)

To make sure we are on the same page here are a few more examples:

```
.info(lambda: "value" > 40)
.infoReset(lambda: "value" < 10)
.warn(lambda: "value" > 60)
.warnReset(lambda: "value" < 20)
.crit(lambda: "value" > 80)
.critReset(lambda: "value" < 30)
```

Given the above criteria the following data points create the following alert levels.

```
45 40 30 9
In In In OK
```

```
45 61 30 19
In Wn Wn OK
```

```
45 61 81 29
In Wn Cr OK
```

Now for a slightly different example

```
.info(lambda: "value" > 40)
.infoReset(lambda: "value" < 30)
.warn(lambda: "value" > 60)
.warnReset(lambda: "value" < 50)
.crit(lambda: "value" > 80)
.critReset(lambda: "value" < 70)
```

```
45 40 30 29
In In In OK
```

```
45 61 49 29
In Wn In OK
```

```
45 61 81 69 50 41 25
In Wn Cr Wn Wn In OK
```

This is how I would expect it to work.

minhdanh:
@nathanielc I updated the test and its sample data with the values in your first example. I could make the test pass on my machine. Should I write another test for the latter example, or just add another tick script and call the same functions in `TestStreamer_Alert_WithReset` function?

nathanielc:
@minhdanh I'd prefer each test to test a single aspect of the behavior, so creating another test is fine.

 I took a quick look at the tests and they look good.

integrations/streamer_test.go
```diff
@@ -4110,7 +4110,7 @@ stream
 	}
 }
 
-func TestStream_Alert_WithReset(t *testing.T) {
+func TestStream_Alert_WithReset_0(t *testing.T) {
 	requestCount := int32(0)
```
minhdanh: @nathanielc I'm not sure how to name two test functions of the same feature so I split them into `TestStream_Alert_WithReset_0` and `TestStream_Alert_WithReset_1`. What do you think?

alert.go
```diff
@@ -577,7 +590,29 @@ func (a *AlertNode) handleAlert(ad *AlertData) {
...
+				}
+				if pass, err := EvalPredicate(se, a.scopePools[l], now, fields, tags); err != nil {
+					a.logger.Printf("E! error evaluating expression for level %v: %s", AlertLevel(l), err)
+					continue
+				} else if pass {
+					if currentLevel < AlertLevel(l) {
+						return AlertLevel(l)
+					}
+				}
+			}
```
minhdanh: @nathanielc There's a little duplication of code here for checking the expressions of the levels. Can you recommend on how to refactor this?

nathanielc: @minhdanh Now that I see how this is working I agree there is a simpler way to perform this logic.

_A reset expressions only prevents an alert from being lowered in severity_

This mean to simplify we can check just the higher level expressions first and only eval the reset expressions if the new level is not higher.

Psudeo code:

```go
for l = CRITICAL; l > currentLevel; l--{
   eval normal expressions, return the first level that evals to true
}

if resetExpression exists {
 if  !resetExpression {
    return currentLevel
 } else {
   for l = currentLevel - 1; l > OK; l-- {
     eval each lower expression and return the first that is true.
  }
} 
```

Since there are two loops that loops over a subset of levels and return the first that matches you could create a function for that logic. something like `findFirstMatch(start , stop AlertLevel,  all other needed data to eval...) (AlertLevel, error)` 

>NOTE: In my pseudo code I am using the `AlertLevel` type instead of `int` I think using the actual `AlertLevel` type will be more clear instead of `len(a.levels)- 1` since we are now comparing against `currentLevel`

Thanks for taking the time to get this right.

minhdanh:
@nathanielc I've implemented your new way. All the tests pass except `TestStream_AlertDuration`. 
I tried to track the problem down but I could not go further. It seems to be a problem with evaluating the normal expressions (https://github.com/influxdata/kapacitor/pull/740/commits/f54b0a2b6555a36ed80b6d5d957a6db7d1cbfc05#diff-d02710009bb4a30f396667696f963627R620)

Here is the output for that test:
```
[edge:task_master:testStreamer|TestStream_AlertDuration->stream] 2016/08/04 13:16:51 D! closing c: 12 e: 10
[edge:task_master:testStreamer|write_points->stream] 2016/08/04 13:16:51 D! closing c: 0 e: 0
[edge:TestStream_AlertDuration|stream->stream0] 2016/08/04 13:16:51 D! closing c: 12 e: 12
[edge:TestStream_AlertDuration|stream0->from1] 2016/08/04 13:16:51 D! closing c: 12 e: 1
[edge:TestStream_AlertDuration|from1->alert2] 2016/08/04 13:16:51 D! closing c: 12 e: 0
[TestStream_AlertDuration:alert2] 2016/08/04 13:16:51 D! CRITICAL alert triggered id:kapacitor/cpu/serverA msg:kapacitor/cpu/serverA is CRITICAL data:&{cpu map[host:serverA type:idle] [time value] [[1971-01-01 00:00:00 +0000 UTC 9]] <nil>}
[TestStream_AlertDuration:alert2] 2016/08/04 13:16:51 D! OK alert triggered id:kapacitor/cpu/serverA msg:kapacitor/cpu/serverA is OK data:&{cpu map[type:idle host:serverA] [time value] [[1971-01-01 00:00:02 +0000 UTC 8]] <nil>}
[TestStream_AlertDuration:alert2] 2016/08/04 13:16:51 D! WARNING alert triggered id:kapacitor/cpu/serverA msg:kapacitor/cpu/serverA is WARNING data:&{cpu map[host:serverA type:idle] [time value] [[1971-01-01 00:00:03 +0000 UTC 8]] <nil>}
[TestStream_AlertDuration:alert2] 2016/08/04 13:16:51 D! OK alert triggered id:kapacitor/cpu/serverA msg:kapacitor/cpu/serverA is OK data:&{cpu map[type:idle host:serverA] [time value] [[1971-01-01 00:00:04 +0000 UTC 6]] <nil>}
[TestStream_AlertDuration:alert2] 2016/08/04 13:16:51 D! WARNING alert triggered id:kapacitor/cpu/serverA msg:kapacitor/cpu/serverA is WARNING data:&{cpu map[host:serverA type:idle] [time value] [[1971-01-01 00:00:05 +0000 UTC 8]] <nil>}
[TestStream_AlertDuration:alert2] 2016/08/04 13:16:51 D! OK alert triggered id:kapacitor/cpu/serverA msg:kapacitor/cpu/serverA is OK data:&{cpu map[host:serverA type:idle] [time value] [[1971-01-01 00:00:08 +0000 UTC 3]] <nil>}
[edge:TestStream_AlertDuration|alert2->http_out3] 2016/08/04 13:16:51 D! closing c: 6 e: 5
[httpd] ::1 - - [04/Aug/2016:13:16:51 +0700] "GET /kapacitor/v1/tasks/TestStream_AlertDuration/TestStream_AlertDuration HTTP/1.1" 200 144 "-" "Go-http-client/1.1" 0880f51a-5a0b-11e6-8088-000000000000 1584
[task_master:testStreamer] 2016/08/04 13:16:51 I! Stopped task: TestStream_AlertDuration
[task_master:testStreamer] 2016/08/04 13:16:51 I! closed
--- FAIL: TestStream_AlertDuration (0.01s)
        streamer_test.go:6512: digraph TestStream_AlertDuration {
                stream0 -> from1;
                from1 -> alert2;
                alert2 -> http_out3;
                }
        streamer_test.go:4980: unexpected alert data for request: 2
                got {kapacitor/cpu/serverA kapacitor/cpu/serverA is OK details 1971-01-01 00:00:02 +0000 UTC 2s OK {0 [] [] <nil>} {{{   map[]}  map[]  {0 0 <nil>}} }}
                exp {kapacitor/cpu/serverA kapacitor/cpu/serverA is WARNING details 1971-01-01 00:00:02 +0000 UTC 2s WARNING {0 [] [] <nil>} {{{   map[]}  map[]  {0 0 <nil>}} }}
        streamer_test.go:4980: unexpected alert data for request: 3
                got {kapacitor/cpu/serverA kapacitor/cpu/serverA is WARNING details 1971-01-01 00:00:03 +0000 UTC 0 WARNING {0 [] [] <nil>} {{{   map[]}  map[]  {0 0 <nil>}} }}
                exp {kapacitor/cpu/serverA kapacitor/cpu/serverA is OK details 1971-01-01 00:00:04 +0000 UTC 4s OK {0 [] [] <nil>} {{{   map[]}  map[]  {0 0 <nil>}} }}
        streamer_test.go:4980: unexpected alert data for request: 4
                got {kapacitor/cpu/serverA kapacitor/cpu/serverA is OK details 1971-01-01 00:00:04 +0000 UTC 1s OK {0 [] [] <nil>} {{{   map[]}  map[]  {0 0 <nil>}} }}
                exp {kapacitor/cpu/serverA kapacitor/cpu/serverA is WARNING details 1971-01-01 00:00:05 +0000 UTC 0 WARNING {0 [] [] <nil>} {{{   map[]}  map[]  {0 0 <nil>}} }}
        streamer_test.go:4980: unexpected alert data for request: 5
                got {kapacitor/cpu/serverA kapacitor/cpu/serverA is WARNING details 1971-01-01 00:00:05 +0000 UTC 0 WARNING {0 [] [] <nil>} {{{   map[]}  map[]  {0 0 <nil>}} }}
                exp {kapacitor/cpu/serverA kapacitor/cpu/serverA is OK details 1971-01-01 00:00:08 +0000 UTC 3s OK {0 [] [] <nil>} {{{   map[]}  map[]  {0 0 <nil>}} }}
        streamer_test.go:4980: unexpected alert data for request: 6
                got {kapacitor/cpu/serverA kapacitor/cpu/serverA is OK details 1971-01-01 00:00:08 +0000 UTC 3s OK {0 [] [] <nil>} {{{   map[]}  map[]  {0 0 <nil>}} }}
                exp {   0001-01-01 00:00:00 +0000 UTC 0 OK {0 [] [] <nil>} {{{   map[]}  map[]  {0 0 <nil>}} }}
        streamer_test.go:5023: got 6 exp 5
```

alert.go
```diff
@@ -552,8 +590,29 @@ func (a *AlertNode) handleAlert(ad *AlertData) {
...
+	if higherLevel, err := a.findFirstMatchLevel(CritAlert, currentLevel-1, now, fields, tags); err == nil {
+		return higherLevel
+	}
+	if rse := a.levelResets[currentLevel]; rse != nil {
+		if pass, err := EvalPredicate(rse, a.lrScopePools[currentLevel], now, fields, tags); err != nil {
+			a.logger.Printf("E! error evaluating reset expression for current level %v: %s", currentLevel, err)
+		} else if !pass {
+			return currentLevel
+		} else {
+			if newLevel, err := a.findFirstMatchLevel(currentLevel-1, OKAlert, now, fields, tags); err == nil {
```
nathanielc: try this. The change is that if the reset expression does not exist then we need to check the rest of the normal expressions below or equal to our current level. 

```go
    if higherLevel, err := a.findFirstMatchLevel(CritAlert, currentLevel-1, now, fields, tags); err == nil {
		return higherLevel
	}
	if rse := a.levelResets[currentLevel]; rse != nil {
		if pass, err := EvalPredicate(rse, a.lrScopePools[currentLevel], now, fields, tags); err != nil {
			a.logger.Printf("E! error evaluating reset expression for current level %v: %s", currentLevel, err)
		} else if !pass {
			return currentLevel
		}
		// fall through to check the rest of the levels.
	}
    if newLevel, err := a.findFirstMatchLevel(currentLevel, OKAlert, now, fields, tags); err == nil {
		return newLevel
	}
    return OKAlert

```

alert.go
```diff
@@ -442,7 +476,11 @@ func (a *AlertNode) runAlert([]byte) error {
 			var highestPoint *models.BatchPoint
 
 			for i, p := range b.Points {
-				l := a.determineLevel(p.Time, p.Fields, p.Tags)
+				var currentLevel AlertLevel
+				if _, ok := a.states[b.Group]; ok {
+					currentLevel = a.states[b.Group].currentLevel()
```
nathanielc: Here and above instead of assigning the result to `_` and then doing another map lookup simply keep the state var around.

```go
if state, ok := a.states[b.Group]; ok {
	currentLevel = state.currentLevel()
}
```

alert.go
minhdanh: This works like a charm. Thank you so much :dancer:

alert.go
minhdanh: This makes sense. Thanks

minhdanh:
@nathanielc I've just applied the fixes and updated CHANGELOG.md. Is there anything else need to be done?

nathanielc:
@minhdanh OK, I have re-reviewed the entire PR. I just have two small recommendations:

1. Can you add the example from the CHANGELOG [here](https://github.com/influxdata/kapacitor/blob/master/pipeline/alert.go#L89) as well? That way it will part of the official docs.
2. The logic for the `findFirstMatchLevel` is a little confusing since it uses the error return to indicate that it didn't find a match. I would be fine with either: adding a bool return value to indicate a match was found, or returning the `stop` level passed in and then comparing against that in the caller. The error return should only be set if an unrecoverable error occurs.

Thanks a ton!

minhdanh:
@nathanielc I changed the error type to bool type to indicate that if a match was found. The function `findFirstMatchLevel` doesn't return an unrecoverable error so I think this makes sense. As now no error returned so I don't do error handling in the callers (which makes the code a little shorter :D). What do you think?

nathanielc:
@minhdanh LGTM, thanks for all the work!

minhdanh:
My pleasure, sir. I couldn't make it without your help. Thank you very much.

melor:
Good stuff! Thanks @minhdanh and @nathanielc !

PR #745, opened 2016-07-23
Adding "if" function to tick

### LINKED ISSUE #741: [Feature request] Support if/else in evalNode
I want to use evalNode like this, 

```
stream
    |eval(lambda:"value" > 100? 30:40)
        .as("substitution")
```
It is just an example, I hope kapacitor's evalNode can have this feature. It's a greate promotion for evalNode and very useful.
^ - ^

yosiat:
- [x] Rebased/mergable
- [x] Tests pass
- [x] CHANGELOG.md updated

Fixes #741

### DIFF
--- a/tick/stateful/functions.go
+++ b/tick/stateful/functions.go
@@ -4,6 +4,7 @@ import (
 	"errors"
 	"fmt"
 	"math"
+	"reflect"
 	"strconv"
 	"time"
 
@@ -87,6 +88,9 @@ func init() {
 
 	// Humanize functions
 	statelessFuncs["humanBytes"] = &humanBytes{}
+
+	// Conditionals
+	statelessFuncs["if"] = &ifFunc{}
 }
 
 // Return set of built-in Funcs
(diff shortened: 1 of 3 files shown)


tick/stateful/functions.go
```diff
@@ -624,3 +627,26 @@ func (*humanBytes) Call(args ...interface{}) (v interface{}, err error) {
...
+type ifFunc struct {
+}
+
+func (*ifFunc) Reset() {
+
+}
+
+func (*ifFunc) Call(args ...interface{}) (interface{}, error) {
+	if len(args) != 3 {
+		return nil, errors.New("if expects exactly three argument")
```
nathanielc: nit: Can you change `argument` to `arguments`?

tick/stateful/functions_test.go
```diff
@@ -485,3 +486,63 @@ func Test_Duration(t *testing.T) {
...
+			exp:  1,
+		},
+		{
+			args: []interface{}{false, 1, 2},
+			exp:  2,
+		},
+
+		// multiple types
+		{
+			args: []interface{}{false, 1, "1"},
```
nathanielc: Do we want to support this? I can't think of a valid use case right now and it could lead to confusing results later on.

For example:

```
|eval(lambda: if("value", 1, 2.5))
```

The type would alternate between int and float. Which would be very confusing, and hard to debug.

What if we added a check that enforced that the two last args are the same type? Then we could catch the error sooner and help the user fix the mistake. Thoughts?

nathanielc:
@yosiat Looks great! A few small questions but overall this looks ready to go.

yosiat:
@nathanielc fixed everything 👍

tick/stateful/functions.go
```diff
@@ -624,3 +628,33 @@ func (*humanBytes) Call(args ...interface{}) (v interface{}, err error) {
...
+
+	var condition bool
+	var isBool bool
+
+	if condition, isBool = args[0].(bool); !isBool {
+		return nil, fmt.Errorf("first argument to if must be a condition with type of bool - got %T", args[0])
+	}
+
+	if reflect.TypeOf(args[1]) != reflect.TypeOf(args[2]) {
+		return nil, fmt.Errorf("Multiple return types isn't supported - second argument is %T and third argument is %T", args[1], args[2])
```
nathanielc: nit:

I think this would read better if it read: `Different return types are not supported - ...`

tick/stateful/functions_test.go
```diff
@@ -485,3 +486,63 @@ func Test_Duration(t *testing.T) {
...
+			err:  errors.New("Multiple return types isn't supported - second argument is int and third argument is string"),
```
nathanielc: Tests are failing on this error message.

PR #753, opened 2016-07-25
Fix broken deadman extra expressions

### LINKED ISSUE #718: Deadman trigger error on lambda statement
I am hoping that someone can point me in the right direction for this, I have been working on this problem for a number of days and can't seem to make the deadman trigger adhere to a lambda statement in the 3rd parameter call.

I have only used the pre-built binaries but have tried 0.13.1, the nightly and 1.0.0-beta2 versions with differing errors.  I have written the script as both a batch and a stream and both appear to have the same errors.  I'm really hoping that I'm just doing this wrong but I have exhausted all the permutations that I can think of and I'm not getting very far.

Please note that the deadman switch works without the lambda statement for me.

**batch script:**

```
> var adt = batch
>         |query('SELECT count("messagecount") from "messages"."default"."messagestats"')
>         .period(1m)
>         .every(1m)
>         .groupBy(time(1m))
> 
> adt
>         |deadman(100.0, 1m, lambda: hour("time") > 20)
>                 .id('TESTING')
>                 .log('/tmp/testing.log')
```

**stream script:**

```
> var adt = stream
>         |from()
>                 .database('messages')
>                 .where(lambda: "server"=='JCPP01' AND "domain"=='ADT_PROD01_19100'
>                         AND "messageto" == 'medipath' AND "messagegenerictype" == 'ADT')
>         .groupBy(time(1m))
> adt
>         |deadman(0.0, 1m, lambda: hour("time") >= 20)
>                 .message('''Detection status update for ADT to Medipath between 7am and 5pm.   We {{ if eq .Level "OK" }}HAVE{{ else }}HAVEN'T{{ end }} sent ADT to Medipath in the last 5 minutes''')
>                 .log('/tmp/medipath2.log')
```

In the 0.13.1 version, for both of these scrits the kapacitor log shows:
`[medipath4:alert5] 2016/07/17 22:37:02 E! error evaluating expression: Failed to handle 1 argument: expression returned unexpected type invalid type`

every minute

In both the nightly and the 1.0.0-beta2 versions, the log shows:

`[medipath4:alert5] 2016/07/17 23:00:00 E! error evaluating expression: Failed to handle 1 argument: name "time" is undefined. Names in scope: emitted`

Again, this is also every minute

Please let me know if there is any further information needed.  Any and all help is appreciated.

Andrew

nathanielc:
Fixes #718 

We need to add tests for this but its not easy until the deadman can have its own clock. I'll create an issue to track this. #754

- [x] Rebased/mergable
- [x] Tests pass
- [x] CHANGELOG.md updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -68,6 +68,7 @@ stream
     The breaking change is that expression evaluation order changed.
     As a result stateful expressions that relied on that order are broken.
 - [#749](https://github.com/influxdata/kapacitor/issues/749): Fix issue with tasks with empty DAG.
+- [#718](https://github.com/influxdata/kapacitor/issues/718): Fix broken extra expressions for deadman's switch.
 
 ## v1.0.0-beta3 [2016-07-09]
 
--- a/expr.go
+++ b/expr.go
@@ -32,7 +32,7 @@ func fillScope(vars *stateful.Scope, referenceVariables []string, now time.Time,
 			continue
 		}
 
-		// Support the error with tags/fields collison
+		// Support the error with tags/fields collision
 		var fieldValue interface{}
 		var isFieldExists bool
 		var tagValue interface{}
--- a/pipeline/node.go
+++ b/pipeline/node.go
@@ -280,7 +280,7 @@ func (n *node) Deadman(threshold float64, interval time.Duration, expr ...*ast.L
 		critExpr = &ast.BinaryNode{
 			Operator: ast.TokenAnd,
 			Left:     critExpr,
-			Right:    e,
+			Right:    e.Expression,
 		}
 	}
 	an.Crit = &ast.LambdaNode{Expression: critExpr}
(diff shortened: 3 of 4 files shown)


gargooza:
Just to confirm, will this fix this error?
`<30> Jul 25 18:15:07.106552 -04:00 kapacitord: [aCSE:alert5] 2016/07/25 18:15:07 E! error evaluating expression: Failed to handle 1 argument: name "time" is undefined. Names in scope: emitted`

nathanielc:
Yes.

FWIW the cause of the error was the the LambdaNode itself was being appended to the alert condition instead of the expressions contained in the LambdaNode. Essentially doing this:

```
|alert()
   .crit(lambda: "emitted" < threshold AND lambda: hour("time") >= 20)
```

Which is clearly wrong, and confused the evaluation of the crit expression.

bowdenandrew:
Thanks Nathaniel,

That definitely fixed the bug that I was seeing, I'll work up some more robust samples and if I have any further problems I'll let you know.

Thanks so much for making this code change.

Andrew

PR #771, opened 2016-08-03
Fix logrotate does not work with default log file.

wutaizeng:
- [x] Rebased/mergable
- [ ] Tests pass
- [ ] CHANGELOG.md updated


nathanielc:
@wutaizeng Ah, good catch, but we need both the .err and .log files to be rotated. Can you update the PR to simply add the .log file?

nathanielc:
@wutaizeng Thanks!

PR #793, opened 2016-08-05
change build.py to print the output of the tests as they are run

nathanielc:
This fixes an issue where CircleCI builds are frequently timing out. Basically the race enabled test suite would take longer than expected and CircleCI would abort the test since no output was received. Now with this change test output is sent to STDOUT immediatly instead of after the entire test suite has finished.

- [x] Rebased/mergable
- [x] Tests pass
- [x] CHANGELOG.md updated

/cc @rossmcdonald


rossmcdonald:
👍  Great idea.

PR #800, opened 2016-08-10
Add zookeeper tickscripts

jackzampolin:
Have alerts for amount of data held and API latency

### DIFF
--- a/examples/telegraf/zookeeper/zookeeper_data_size_batch.tick
+++ b/examples/telegraf/zookeeper/zookeeper_data_size_batch.tick
@@ -1,0 +1,47 @@
+// zookeeper_data_size_batch
+
+// metric: "approximate_data_size"
+// available_fields: "avg_latency","ephemerals_count","max_file_descriptor_count","max_latency","min_latency","num_alive_connections","open_file_descriptor_count","outstanding_requests","packets_received","packets_sent","version","watch_count","znode_count" 
+
+// TELEGRAF CONFIGURATION
+// [inputs.zookeeper]
+//   servers = [":2181"]
+
+// DEFINE: kapacitor define zookeeper_data_size_batch -type batch -tick zookeeper/zookeeper_data_size_batch.tick -dbrp telegraf.autogen
+// ENABLE: kapacitor enable zookeeper_data_size_batch
+
+// Parameters
+var info = 100000000 // 100MB
+var warn = 1000000000 // 1GB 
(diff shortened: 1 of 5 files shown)


examples/telegraf/zookeeper/zookeeper_data_size_batch.tick
```diff
@@ -0,0 +1,47 @@
...
+
+// TELEGRAF CONFIGURATION
+// [inputs.zookeeper]
+//   servers = [":2181"]
+
+// DEFINE: kapacitor define zookeeper_data_size_batch -type batch -tick zookeeper/zookeeper_data_size_batch.tick -dbrp telegraf.autogen
+// ENABLE: kapacitor enable zookeeper_data_size_batch
+
+// Parameters
+var info = 100000000 // 100MB
```
nathanielc: We should make this easier to use. Possibly by defining a sizeUnit var and then multiplying the thresholds by it. For example

```
var mb = 1024*1024 // 1 MB
var gb = 1024*mb // 1 GB

var info = 100 * mb 
var warn = 1 * gb 
var crit = 10 * gb
```

It demonstrates some of the flexibility of TICKscript and it reads better to me. Thoughts?

jackzampolin: +100. Thats a WAY more readable format.

examples/telegraf/zookeeper/zookeeper_data_size_batch.tick
```diff
@@ -0,0 +1,47 @@
...
+// Dataframe
+var data = batch
+  |query('''SELECT mean(approximate_data_size) AS stat FROM "telegraf"."autogen"."zookeeper"''')
+    .period(period)
+    .every(every)
+    .groupBy('host')
+        
+// Thresholds
+var alert = data
+  |eval(lambda: sigma("stat"))
```
nathanielc: An alternative approach would be to dived the stat here by `mb` or `gb` and then set the thresholds in those units.

```
var info = 0.1 // 100 MB
var warn = 1 // 1 GB
var crit = 10 // 1 GB
```

```
|eval(lambda:"stat" / gb, lambda: sigma("stat"))
   .as('stat', 'sigma')
|...
```

jackzampolin: I went with the other option

nathanielc:
@jackzampolin Thanks!

PR #831, opened 2016-08-24
Improve join on docs

nathanielc:


pipeline/join.go
```diff
@@ -146,13 +146,21 @@ func (j *JoinNode) As(names ...string) *JoinNode {
...
 }
 
-// Join on specific dimensions.
+// Join on a subset of the group by dimensions.
+// This is a special case where you want a single point from one parent to join with multiple
+// points from a different parent.
+//
 // For example given two measurements:
 //
 // 1. building_power -- tagged by building, value is the total power consumed by the building.
```
rossmcdonald: Could we update this section to be:

> // 1. building_power (a single value) -- tagged by building, value is the total power consumed by the building.
> // 2. floor_power (multiple values) -- tagged by building and floor, values are the total power consumed by each floor.

I think that would help drive the point home.

PR #852, opened 2016-08-30
fix typo

jademcgough:
pulled from https://github.com/influxdata/docs.influxdata.com/pull/625

When I was in academia, people were pretty serious about data being plural, but a google search suggests that either is acceptable. I'll leave the grammar decisions up to you.


nathanielc:
@jademcgough Thanks!

Issue #625, opened 2016-06-08
IP Geotagging


nathanielc:
Add the ability to add fields or tags to data for is geo location based on IP.

PR #858, opened 2016-08-31
Update with changes from docs #625

rkuchan:
Port changes from https://github.com/influxdata/docs.influxdata.com/pull/625/files. I made three additional changes too.


nathanielc:
@rkuchan Thanks!

PR #873, opened 2016-09-06
Add TCP alert handler

rossmcdonald:
- [x] Rebased/mergable
- [x] Tests pass
- [x] CHANGELOG.md updated

### DIFF
--- a/alert.go
+++ b/alert.go
@@ -936,7 +936,7 @@ func (a *AlertNode) handleTcp(tcp *pipeline.TcpHandler, ad *AlertData) {
 	}
 	defer conn.Close()
 
-	buf.WriteByte("\n")
+	buf.WriteByte('\n')
 	conn.Write(buf.Bytes())
 
 	return
--- a/alert.go
+++ b/alert.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	html "html/template"
 	"log"
+	"net"
 	"net/http"
 	"os"
 	"os/exec"
@@ -171,6 +172,11 @@ func newAlertNode(et *ExecutingTask, n *pipeline.AlertNode, l *log.Logger) (an *
 	for _, post := range n.PostHandlers {
 		post := post
 		an.handlers = append(an.handlers, func(ad *AlertData) { an.handlePost(post, ad) })
+	}
+
+	for _, tcp := range n.TcpHandlers {
+		tcp := tcp
+		an.handlers = append(an.handlers, func(ad *AlertData) { an.handleTcp(tcp, ad) })
 	}
 
 	for _, email := range n.EmailHandlers {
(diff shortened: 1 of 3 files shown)


alert.go
```diff
@@ -910,6 +916,31 @@ func (a *AlertNode) handlePost(post *pipeline.PostHandler, ad *AlertData) {
...
+		a.bufPool.Put(bodyBuffer)
+	}()
+
+	err := json.NewEncoder(bodyBuffer).Encode(ad)
+	if err != nil {
+		a.logger.Println("E! failed to marshal alert data json", err)
+		return
+	}
+
+	conn, err := net.Dial("tcp", tcp.Address)
```
nathanielc: Would it be possible to reuse the same TCP connection and use the Write method to send individual packets for each event? Not sure if that would break the server side.

alert.go
```diff
@@ -910,6 +916,31 @@ func (a *AlertNode) handlePost(post *pipeline.PostHandler, ad *AlertData) {
...
+	if err != nil {
+		a.logger.Println("E! failed to connect", err)
+		return
+	}
+
+	fmt.Fprintf(conn, bodyBuffer+"\n")
```
nathanielc: This should probably be:

```
bodyBuffer.WriteByte('\n')
conn.Write(bodyBuffer.Bytes())
```

alert.go
```diff
@@ -910,6 +916,31 @@ func (a *AlertNode) handlePost(post *pipeline.PostHandler, ad *AlertData) {
...
+
+	defer conn.Close()
```
phemmer: Why a `defer` when the next line is `return`?

alert.go
rossmcdonald: I had considered that, but it seemed like it would be a pretty large undertaking. I'm assuming it would require some connection pooling per handler, with some mechanisms for checking the health of the connections (and maybe closing connections after a longer period of inactivity). Not sure if that's way off base or not, though. 

I'd be up for discussing it more in depth offline if you have some time.

alert.go
rossmcdonald: Fixed.

alert.go
rossmcdonald: I actually copied this convention from the `handlePost` method [above](https://github.com/influxdata/kapacitor/pull/873/files/fe4e9d5bb92515f74b42a3034997332394e3fc56#diff-d02710009bb4a30f396667696f963627R915). @nathanielc might be able to clarify, but I assume this was added to ensure `Close()` is always called and to prevent the socket from remaining in ESTABLISHED for longer than it needs to.

rossmcdonald: Actually I was incorrect, and the `defer` has now been moved to be before any data is written over the connection.

pipeline/alert.go
```diff
@@ -464,6 +470,26 @@ type PostHandler struct {
...
+	a.TcpHandlers = append(a.TcpHandlers, tcp)
+	return tcp
+}
+
+// tick:embedded:AlertNode.Tcp
+type TcpHandler struct {
+	*AlertNode
+
+	// The endpoint address.
+	// tick:ignore
```
nathanielc: This shouldn't be ignored.

rossmcdonald: Fixed.

PR #884, opened 2016-09-09
[WIP] Alerting system design doc

nathanielc:
We should probably flesh out this doc with more of the existing alert design. For now this doc contains the proposed enhancements to the alerting system.
 
@joelegasse

### DIFF
--- a/ALERT_DESIGN.md
+++ b/ALERT_DESIGN.md
@@ -1,0 +1,72 @@
+# Kapacitor Alerting
+
+
+Kapacitor should expose the current state of all alerts
+Kapacitor should allow alerts to be throttled/aggregated and sent to varying alert handlers without having to modify all tasks.
+
+## Implementation
+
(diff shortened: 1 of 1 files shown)


nathanielc:
A section on the write path has been added.

nathanielc:
@joelegasse I have updated this doc with information about existing alerts and this new alert system and some of the differences.

ALERT_DESIGN.md
```diff
@@ -0,0 +1,93 @@
...
+        {"pagerDuty": {"serviceKey":"XXX"}}
+    ]
+}
+```
+
+
+#### Implementation
+
+The underlying implementation will be a basic publish/subscribe system.
+Various subscriber can be defined via the above definitions which can in turn publish back to the internal system or send alerts to third parties.
```
nathanielc: The actual definitions need to be worked out and documented. The above are just example ideas.
Somethings not addressed by the above are how alerts are referenced, via a name/ID or something else.

dstreppa:
+1

PR #921, opened 2016-09-19
Add string manipulation functions

### LINKED ISSUE #777: Add string manipulation functions
Similar to how we have the Go math package exposed via TICKscript functions we should do the same for the strings package.

Some initial thoughts.

*  Only import functions that take strings args and other TICKscript primitives, not rune based functions. 
* Since the function names in the strings package are scoped by the `strings.` name in the Go language but TICKScript has global scope we need to decide on a consistent naming convention for these functions.  Probably prefixing the functions with `str` will be enough.

nathanielc:
Fixes #777 

There are three major changes in this PR:

1. Add a set of functions to the lambda expressions for string manipulation
2. Update the lexer/parser to allow for regex literals to be function arguments, (this was needed for one of the string manipulation functions).
3. Convert all stateless function objects to value receivers.

- [x] Rebased/mergable
- [x] Tests pass
- [x] CHANGELOG.md updated

### DIFF
--- a/tick/stateful/functions.go
+++ b/tick/stateful/functions.go
@@ -5,7 +5,9 @@ import (
 	"fmt"
 	"math"
 	"reflect"
+	"regexp"
 	"strconv"
+	"strings"
 	"time"
 
 	"github.com/dustin/go-humanize"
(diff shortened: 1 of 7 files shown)


nathanielc:
@joelegasse This is ready for review.

tick/stateful/functions.go
```diff
@@ -234,6 +256,232 @@ func (m *mathIF) Call(args ...interface{}) (v interface{}, err error) {
...
+	v = str[start:stop]
+	return
+}
+
+func (m *strSubstring) Reset() {}
+
+type regexReplace struct {
+}
+
+func (m *regexReplace) Call(args ...interface{}) (v interface{}, err error) {
```
joelegasse: regexReplace, strSubstring, and strReplace should all have value receivers as well as be used as values in map  above. There is no benefit in storing a pointer to an empty struct. The other types would be fine as value receivers as well, since they only store a string and a function pointer, and are immutable.

joelegasse: I'm also not sure the `newX` functions add anything to the code instead of doing `string2Bool{name:"strContains", f:strings.Contains}`

nathanielc: Good catch, I'll update them to value receivers.

nathanielc:
@joelegasse I have updated the function objects to be value receivers

PR #947, opened 2016-09-28
Help with adding a kafka alert

vistorve:
I will preface this by saying this is my first time writing Go so mistakes could easily be made.

I am trying to add a kafka alert node and I think I have followed the same pattern other new alert nodes have. But I am stuck on getting the following message when I try to create a tick script using it:

`invalid TICKscript: line 13 char 3: no method or property "kafka" on *pipeline.AlertNode`

Any help or guidance would be greatly appreciated.

Tick script:
```
stream
    // Select just the cpu measurement from our example database.
    |from()
        .measurement('cpu')
    |alert()
	.id('{{ index .Tags "cpu" }} on {{ index .Tags "host" }}')
	.message('{{ .ID }} has remaining idle {{ index .Fields "usage_idle" }}')
        .crit(lambda: "usage_idle" <  70)
	.kafka()

```

- [ ] Rebased/mergable
- [ ] Tests pass
- [ ] CHANGELOG.md updated

### DIFF
--- a/alert.go
+++ b/alert.go
@@ -269,6 +269,20 @@ func newAlertNode(et *ExecutingTask, n *pipeline.AlertNode, l *log.Logger) (an *
 		n.IsStateChangesOnly = true
 	}
 
+	for _, kafka := range n.KafkaHandlers {
+		kafka := kafka
+		an.handlers = append(an.handlers, func(ad *AlertData) { an.handleKafka(kafka, ad) })
+	}
+	if len(n.KafkaHandlers) == 0 && (et.tm.KafkaService != nil && et.tm.KafkaService.Global()) {
+		an.handlers = append(an.handlers, func(ad *AlertData) { an.handleKafka(&pipeline.KafkaHandler{}, ad) })
+	}
+	// If Kafka has been configured with state changes only set it.
+	if et.tm.KafkaService != nil &&
+		et.tm.KafkaService.Global() &&
+		et.tm.KafkaService.StateChangesOnly() {
+		n.IsStateChangesOnly = true
+	}
+
 	for _, alerta := range n.AlertaHandlers {
 		// Validate alerta templates
 		rtmpl, err := text.New("resource").Parse(alerta.Resource)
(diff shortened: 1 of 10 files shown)


alert.go
```diff
@@ -20,7 +20,7 @@ import (
 	imodels "github.com/influxdata/influxdb/models"
 	"github.com/influxdata/kapacitor/expvar"
 	"github.com/influxdata/kapacitor/models"
-	"github.com/influxdata/kapacitor/pipeline"
+	"./pipeline"
```
nathanielc: This change is incorrect, you can leave it how it was.

This may be an indication that your GOPATH is not setup correctly yet.

pipeline/alert.go
```diff
@@ -1254,3 +1259,37 @@ func (a *AlertNode) Talk() *TalkHandler {
...
+//
+// Example:
+//    stream
+//         |alert()
+//             .kafka()
+//
+// Send alerts to Kafka.
+//
+// tick:property
+func (a *AlertNode) Kafka(topics []string) *KafkaHandler {
```
nathanielc: This function signature should be:

`func (a *AlertNode) Kafka(topics ...string) *KafkaHandler {`

The `...` means is a [variadic](https://en.wikipedia.org/wiki/Variadic_function) function.

server/config.go
```diff
@@ -32,7 +32,7 @@ import (
 	"github.com/influxdata/kapacitor/services/udf"
 	"github.com/influxdata/kapacitor/services/udp"
 	"github.com/influxdata/kapacitor/services/victorops"
-
+	"../services/kafka"
```
nathanielc: Again this should be a fully qualified import path.

services/kafka/config.go
```diff
@@ -0,0 +1,22 @@
+package kafka
+
+
+const DEFAULT_TOPIC = "Important"
```
nathanielc: The Go naming convention here is `DefaultTopic`.

PR #1038, opened 2016-11-14
Update union node

### LINKED ISSUE #1100: UnionNode buffers points without warning
After 4 hours on this one, I concede defeat.

Consider this TICKscript:

```
var one = stream
    |from()
        .measurement('one')

var two = stream
    |from()
        .measurement('two')

one
    |union(two)
        .rename('input')
    @myUDF()
    |log()
```

It works perfectly on recorded data provided the record uses a query to select historical data. However calling `kapacitor record stream` does not return and no data is recorded even if a point is inserted into either `one` or `two`.

Further, enabling the task produces no log output **until `x` points are added to either `one` or `two`**. `x` is about 6, but I couldn't see consistent behaviour to determine a pattern. From there, every time a point is added to either `one` or `two`, the point `x` entries old gets processed. That is, it acts like a FIFO with `x` elements.

I could find no explanation of this behaviour in the docs, in the source code or via Google searches.

I finally traced it down to the `UnionNode` by looking at the DOT output of `kapacitor show my_task`:

```
digraph my_task {
graph [throughput="0.00 points/s"];

stream0 [avg_exec_time_ns="0s" ];
stream0 -> from2 [processed="15"];
stream0 -> from1 [processed="15"];

from2 [avg_exec_time_ns="2.67µs" ];
from2 -> union4 [processed="11"];

from1 [avg_exec_time_ns="1.894µs" ];
from1 -> union4 [processed="4"];

union4 [avg_exec_time_ns="0s" ];
union4 -> myTask5 [processed="9"];

calcOccupancy5 [avg_exec_time_ns="0s" ];
calcOccupancy5 -> log6 [processed="9"];

log6 [avg_exec_time_ns="0s" ];
log6 -> influxdb_out7 [processed="9"];

influxdb_out7 [avg_exec_time_ns="0s" points_written="8" write_errors="0" ];
}
```

This seems to indicate that the `FromNodes` are processing the points immediately, but the `UnionNode` is 6 points behind. Sure enough, if I remove the `UnionNode` and connect one of the `FromNode`'s directly, and then add a point to that `FromNode`, it is processed immediately:

```
digraph my_task {
graph [throughput="0.00 points/s"];

stream0 [avg_exec_time_ns="0s" ];
stream0 -> from2 [processed="1"];
stream0 -> from1 [processed="1"];

from2 [avg_exec_time_ns="0s" ];
from2 -> calcOccupancy3 [processed="1"];

calcOccupancy3 [avg_exec_time_ns="0s" ];
calcOccupancy3 -> log4 [processed="1"];

log4 [avg_exec_time_ns="0s" ];
log4 -> influxdb_out5 [processed="1"];

influxdb_out5 [avg_exec_time_ns="0s" points_written="1" write_errors="0" ];

from1 [avg_exec_time_ns="1.15µs" ];
}
```

Is there an explanation for this mysterious behaviour. Can it be disabled so that points are processed immediately?

###### URL for relevant page?
https://docs.influxdata.com/kapacitor/v1.1//nodes/union_node/

###### What products and version are you using?
InfluxDB 1.1.1
Kapacitor 1.1.1

###### Where did you look before opening the issue?
https://docs.influxdata.com/kapacitor/v1.1//nodes/union_node/
https://github.com/influxdata/kapacitor/blob/master/pipeline/union.go
https://github.com/influxdata/kapacitor/issues/361

nathanielc:
The union node was caching points for too long. This change makes it so that the union node emits points as soon as possible. 

Writing a test for this has been a bit challenging as the change in behavior is only related to the timing of emits, and cannot be easily tested with the existing integration tests.

I have put together a hacky test for now to demonstrate its working, but it probably shouldn't be merged in this state.

The ideal way to test this would be either: clean up the `ExecutingTask` code such that it is easier to setup a single node in isolation, or enhance the edge communication so it can carry metadata messages as well and add `barrier` messages. The reality is both probably need to happen.

- [x] Rebased/mergable
- [x] Tests pass
- [x] CHANGELOG.md updated

Fixes #1100

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -62,8 +62,8 @@ stream
 ### Bugfixes
 
 - [#1045](https://github.com/influxdata/kapacitor/issues/1045): Fix panic during replays.
-
 - [#1043](https://github.com/influxdata/kapacitor/issues/1043): logrotate.d ignores kapacitor configuration due to bad file mode
+- [#1100](https://github.com/influxdata/kapacitor/issues/1100): Fix issue with the Union node buffering more points than necessary.
 
 ## v1.1.1 [2016-12-02]
 
--- a/integrations/data/TestStream_Union_Stepped.srpl
+++ b/integrations/data/TestStream_Union_Stepped.srpl
@@ -1,0 +1,78 @@
+dbname
+rpname
+cpu,cpu=0 value=91 0000000000
+dbname
+rpname
+cpu,cpu=1 value=97 0000000000
+dbname
+rpname
+cpu,cpu=0 value=98 0000000001
+dbname
+rpname
+cpu,cpu=1 value=98 0000000001
+dbname
+rpname
+cpu,cpu=total value=92 0000000002
+dbname
+rpname
+cpu,cpu=1 value=92 0000000002
+dbname
+rpname
+cpu,cpu=0 value=95 0000000003
+dbname
+rpname
+cpu,cpu=1 value=95 0000000003
+dbname
+rpname
+cpu,cpu=total value=93 0000000004
+dbname
+rpname
+cpu,cpu=1 value=93 0000000004
+dbname
+rpname
+cpu,cpu=1 value=92 0000000005
+dbname
+rpname
+cpu,cpu=0 value=92 0000000005
+dbname
+rpname
+cpu,cpu=0 value=95 0000000006
+dbname
+rpname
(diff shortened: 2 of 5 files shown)


nathanielc:
I decided that we need this fix merged even if we can't write a stable test for it. Eventually we need to be able to step task execution but that represents a lot of work and should not block this positive fix.

For now the unstable test has been committed with a Skip message.

Issue #1052, opened 2016-11-17
[alerta] TLS Support


allen13:
Currently Kapacitor is sending the Alerta api token as a url param which is unencrypted even when ssl is used. Alerta supports tokens being provided via headers and I am working to make this change. In addition I am adding an option to skip TLS verification for convenience purposes.

nathanielc:
@allen13 Thanks!

PR #1053, opened 2016-11-17
Move alerta api token to header and add option to skip TLS verification.

allen13:
- [x] Rebased/mergable
- [x] Tests pass
- [x] CHANGELOG.md updated

Resolves issue #1052 

I have tested this change live with the custom kapacitor docker image allen13/kapacitor:latest

### DIFF
--- a/services/alerta/config.go
+++ b/services/alerta/config.go
@@ -11,6 +11,8 @@ type Config struct {
 	Enabled bool `toml:"enabled" override:"enabled"`
 	// The Alerta URL.
 	URL string `toml:"url" override:"url"`
+	// Whether to skip the tls verification of the alerta host
+	InsecureSkipVerify bool `toml:"insecure-skip-verify" override:"insecure-skip-verify"`
 	// The authentication token for this notification, can be overridden per alert.
 	Token string `toml:"token" override:"token,redact"`
 	// The environment in which to raise the alert.
--- a/services/alerta/service.go
+++ b/services/alerta/service.go
@@ -2,10 +2,10 @@ package alerta
 
 import (
 	"bytes"
+	"crypto/tls"
 	"encoding/json"
 	"errors"
 	"fmt"
-	"io"
 	"io/ioutil"
 	"log"
 	"net/http"
@@ -16,6 +16,7 @@ import (
 
 type Service struct {
 	configValue atomic.Value
+	clientValue atomic.Value
 	logger      *log.Logger
 }
 
@@ -24,6 +25,11 @@ func NewService(c Config, l *log.Logger) *Service {
 		logger: l,
 	}
 	s.configValue.Store(c)
+	s.clientValue.Store(&http.Client{
+		Transport: &http.Transport{
+			TLSClientConfig: &tls.Config{InsecureSkipVerify: c.InsecureSkipVerify},
+		},
+	})
 	return s
 }
 
@@ -95,7 +101,13 @@ func (s *Service) Update(newConfig []interface{}) error {
 		return fmt.Errorf("expected config object to be of type %T, got %T", c, newConfig[0])
 	} else {
 		s.configValue.Store(c)
+		s.clientValue.Store(&http.Client{
+			Transport: &http.Transport{
+				TLSClientConfig: &tls.Config{InsecureSkipVerify: c.InsecureSkipVerify},
+			},
+		})
 	}
+
 	return nil
 }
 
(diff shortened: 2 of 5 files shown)


services/alerta/config.go
```diff
@@ -11,6 +11,8 @@ type Config struct {
 	Enabled bool `toml:"enabled" override:"enabled"`
 	// The Alerta URL.
 	URL string `toml:"url" override:"url"`
+	// Whether to skip the tls verification of the alerta host
+	TLSSkipVerify bool `toml:"tls_skip_verify" override:"tls_skip_verify"`
```
nathanielc: Can you rename this option to `InsecureSkipVerify` and `insecure-skip-verify` so that it is consistent with other similar config options?

services/alerta/service.go
```diff
@@ -100,16 +100,23 @@ func (s *Service) Update(newConfig []interface{}) error {
...
 	}
 
-	url, post, err := s.preparePost(token, resource, event, environment, severity, group, value, message, origin, service, data)
+	req, err := s.preparePost(token, resource, event, environment, severity, group, value, message, origin, service, data)
 	if err != nil {
 		return err
 	}
 
-	resp, err := http.Post(url, "application/json", post)
+	client := &http.Client{
```
nathanielc: Can you create the client in the `NewService` method so that we don't have to create a new client for each new request?

Since the Alerta config can change dynamically you will also need to create a new client in the `Update` method if the config changed. And that also means that you will need to synchronize access to the client. For that I would recommend using atomic.Value, see how the `config` field is access for an example.

Let me know if I can clarify any of that better.

allen13: Yeah, good catch. I'm pretty sure I understand how the update method works. Let me fix that up.

services/alerta/config.go
allen13: Yup no problem.

allen13:
Updated to address above issues.

nathanielc:
Here is the failing test from `server/server_test.go`
```
 TestServer_UpdateConfig (0.25s)
	server_test.go:6230: unexpected defaults for alerta/: alerta: section element 0 are not equal: extra option "insecure-skip-verify" with value <nil>
	server_test.go:6245: unexpected update result 0 for alerta/: alerta: section element 0 are not equal: extra option "insecure-skip-verify" with value <nil>
```

allen13:
Tests passes this time!

services/alerta/service.go
```diff
@@ -17,13 +17,19 @@ import (
 type Service struct {
 	configValue atomic.Value
 	logger      *log.Logger
+	client      *http.Client
```
nathanielc: The client field here should be an `atomic.Value` type and to be consistent we should call it  `clientValue`.

services/alerta/service.go
```diff
@@ -17,13 +17,19 @@ import (
...
 }
 
 func NewService(c Config, l *log.Logger) *Service {
 	s := &Service{
 		logger: l,
 	}
 	s.configValue.Store(c)
+	s.client = &http.Client{
```
nathanielc: Here the use the `Store` method to store the new client. Much like the configValue line above.

services/alerta/service.go
```diff
@@ -95,7 +101,13 @@ func (s *Service) Update(newConfig []interface{}) error {
 		return fmt.Errorf("expected config object to be of type %T, got %T", c, newConfig[0])
 	} else {
 		s.configValue.Store(c)
+		s.client = &http.Client{
```
nathanielc: Same goes here, we should use the `Store` method.

services/alerta/service.go
```diff
@@ -104,12 +116,12 @@ func (s *Service) Alert(token, resource, event, environment, severity, group, va
...
 	}
 
-	url, post, err := s.preparePost(token, resource, event, environment, severity, group, value, message, origin, service, data)
+	req, err := s.preparePost(token, resource, event, environment, severity, group, value, message, origin, service, data)
 	if err != nil {
 		return err
 	}
 
-	resp, err := http.Post(url, "application/json", post)
+	resp, err := s.client.Do(req)
```
nathanielc: Lastly this becomes:

```
client := s.clientValue.Get().(*http.Client)
resp, err := client.Do(req)
```

All of these changes avoid a race condition on the `client` field, synchronizing access to the field.

allen13:
Race condition resolved.

allen13:
Thanks for the help!

Issue #1059, opened 2016-11-23
broken debian package build


earwin:
fpm `--directories` stanzas in `build.py` become a part of package description:
```
user@host:~$ apt show kapacitor
Package: kapacitor
Version: 1.1.0-1
License: MIT
Vendor: InfluxData
Maintainer: support@influxdb.com
Installed-Size: 33.7 MB
Homepage: github.com/influxdata/kapacitor
Priority: extra
Section: default
Download-Size: 9,557 kB
APT-Sources: https://repos.influxdata.com/ubuntu/ trusty/stable amd64 Packages
Description: var/log/kapacitor --directories var/lib/kapacitor --directories usr/lib/kapacitor/scripts --directories usr/lib/kapacitor --directories etc/kapacitor
```

rossmcdonald:
It looks like this was introduced with https://github.com/influxdata/kapacitor/commit/80d4e71e6e5aab53aefb4d03a20dad5ac59985fb. I'll have a fix up shortly.

PR #1060, opened 2016-11-23
Fix packaging arguments

rossmcdonald:
- [x] Rebased/mergable
- [x] Tests pass

Fixes https://github.com/influxdata/kapacitor/issues/1059

### DIFF
--- a/build.py
+++ b/build.py
@@ -59,7 +59,6 @@ fpm_common_args = "-f -s dir --log error \
  --maintainer {} \
  --config-files {} \
  --config-files {} \
- --config-files {} \
  --directories {} \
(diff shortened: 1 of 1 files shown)


nathanielc:
@rossmcdonald Thanks for fixing this. Why is bash completion considered config and not just part of the normal package files?

rossmcdonald:
Ah, good point. Actually if it's not a config, then it shouldn't need to be listed in the `fpm_common_args` variable at all. I'll make it so that's it just a regular file.

rossmcdonald:
@nathanielc I've updated it so that the bash completion file is treated like a normal part of the package, and not a configuration file.

PR #1097, opened 2016-12-20
Miscellaneous lint cleanup

mark-rushakoff:
- [x] Rebased/mergable
- [x] Tests pass

### DIFF
--- a/log.go
+++ b/log.go
@@ -40,7 +40,9 @@ func (s *LogNode) runLog([]byte) error {
 	case pipeline.StreamEdge:
 		for p, ok := s.ins[0].NextPoint(); ok; p, ok = s.ins[0].NextPoint() {
 			buf.Reset()
-			env.Encode(p)
+			if err := env.Encode(p); err != nil {
+				return err
+			}
 			s.logger.Println(key, buf.String())
 			for _, child := range s.outs {
 				err := child.CollectPoint(p)
@@ -52,7 +54,9 @@ func (s *LogNode) runLog([]byte) error {
 	case pipeline.BatchEdge:
 		for b, ok := s.ins[0].NextBatch(); ok; b, ok = s.ins[0].NextBatch() {
 			buf.Reset()
-			env.Encode(b)
+			if err := env.Encode(b); err != nil {
+				return err
+			}
 			s.logger.Println(key, buf.String())
 			for _, child := range s.outs {
 				err := child.CollectBatch(b)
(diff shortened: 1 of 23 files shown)


nathanielc:
@mark-rushakoff Thanks for cleaning up these issues. What specific  lint command do you run? I'd like to add it to the docs or possibly the build process.

mark-rushakoff:
I ran a few specific linters from [gometalinter](https://github.com/alecthomas/gometalinter) by hand, namely ineffassign, errcheck, gosimple, and staticcheck. It's probably reasonable to have any of those running in automation, but there will be some false positives to sift through.

nathanielc:
@mark-rushakoff Can you rebase these changes? Was just about to merge them in and realized my big PR the other day created some conflicts. Sorry about that.

log.go
```diff
@@ -40,7 +40,9 @@ func (s *LogNode) runLog([]byte) error {
 	case pipeline.StreamEdge:
 		for p, ok := s.ins[0].NextPoint(); ok; p, ok = s.ins[0].NextPoint() {
 			buf.Reset()
-			env.Encode(p)
+			if err := env.Encode(p); err != nil {
+				return err
```
nathanielc: Returning here will stop the task execution. The err should simply be logged a la

`s.logger.Println("E!", err)`

log.go
```diff
@@ -52,7 +54,9 @@ func (s *LogNode) runLog([]byte) error {
 	case pipeline.BatchEdge:
 		for b, ok := s.ins[0].NextBatch(); ok; b, ok = s.ins[0].NextBatch() {
 			buf.Reset()
-			env.Encode(b)
+			if err := env.Encode(b); err != nil {
+				return err
```
nathanielc: same as above

server/server.go
```diff
@@ -793,7 +793,9 @@ func (s *Server) startProfile(cpuprofile, memprofile string) {
 		}
 		s.Logger.Printf("I! writing CPU profile to: %s\n", cpuprofile)
 		prof.cpu = f
-		pprof.StartCPUProfile(prof.cpu)
+		if err := pprof.StartCPUProfile(prof.cpu); err != nil {
+			s.Logger.Fatalf("#! start cpu profile: %v", err)
```
nathanielc: Hmm, so I see that logger.Fatal is being use here but I'd rather remove it than add to it at this point.  Maybe we leave this little bit as is for now, or setup a clean shutdown mechanism here.

mark-rushakoff:
@nathanielc I updated the `return err` lines but I didn't totally follow the other comment about logger.Fatal. Feel free to take over the branch for any remaining small modifications.

nathanielc:
@mark-rushakoff Thanks!, I'll merge on green.

PR #1107, opened 2017-01-04
Example Foo alert handler

nathanielc:
The PR provides an example of adding an new alert handler to Kapacitor.

For an explanation of what is going on in this PR see https://github.com/influxdata/kapacitor/blob/master/alert/HANDLERS.md

This PR is not meant to actually be merged.


nathanielc:
Closing as out of date. We can update this and open a new PR, for now this is not helpful as its incomplete and out of date.

PR #1110, opened 2017-01-05
Add new query property for aligning group by intervals to start times

goakley:
Allow QueryNodes to automatically align GROUP BY time intervals to the start time for each query sent to influx.  This leverages the offset argument to `GROUP BY time()` to offset directly to the start time.

- [x] Rebased/mergable
- [x] Tests pass
- [x] CHANGELOG.md updated

### DIFF
--- a/pipeline/batch.go
+++ b/pipeline/batch.go
@@ -113,6 +113,10 @@ type QueryNode struct {
 	// 1 AM and the Offset is 1 hour. Then at 1 AM on Sunday the data from 12 AM will be queried.
 	Offset time.Duration
 
+	// Align the group by time intervals with the start time of the query
+	// tick:ignore
+	AlignGroupFlag bool `tick:"AlignGroup"`
+
 	// The list of dimensions for the group-by clause.
 	//tick:ignore
 	Dimensions []interface{} `tick:"GroupBy"`
@@ -178,6 +182,26 @@ func (n *QueryNode) ChainMethods() map[string]reflect.Value {
 //
 // It is recommended to use QueryNode.Align and QueryNode.Offset in conjunction with
 // group by time dimensions so that the time bounds match up with the group by intervals.
+// To automatically align the group by intervals to the start of the query time,
+// use QueryNode.AlignGroup. This is useful in more complex situations, such as when
+// the groupBy time period is longer than the query frequency.
+//
+// Example:
+//    batch
+//        |query(...)
+//            .period(5m)
+//            .every(30s)
+//            .groupBy(time(1m), 'tag1', 'tag2')
+//            .align()
+//            .alignGroup()
+//
+// For the above example, without QueryNode.AlignGroup, every other query issued by Kapacitor
+// (at :30 past the minute) will align to :00 seconds instead of the desired :30 seconds,
+// which would create 6 group by intervals instead of 5, the first and last of which
+// would only have 30 seconds of data instead of a full minute.
+// If the group by time offset (i.e. time(t, offset)) is used in conjunction with
+// QueryNode.AlignGroup, the alignment will occur first, and will be offset
+// the specified amount after.
 //
 // NOTE: Since QueryNode.Offset is inherently a negative property the second "offset" argument to the "time" function is negative to match.
 //
@@ -210,5 +234,12 @@ func (n *QueryNode) GroupByMeasurement() *QueryNode {
 // tick:property
 func (b *QueryNode) Align() *QueryNode {
 	b.AlignFlag = true
+	return b
+}
+
+// Align the group by time intervals with the start time of the query
+// tick:property
+func (b *QueryNode) AlignGroup() *QueryNode {
+	b.AlignGroupFlag = true
 	return b
 }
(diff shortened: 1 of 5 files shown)


nathanielc:
@goakley On a quick read through this looks great! I'll give it a full review soon. In the meantime could you sign the CLA?

goakley:
Signed 👍

query.go
```diff
@@ -96,6 +98,9 @@ func (q *Query) StopTime() time.Time {
 // Set the start time of the query
 func (q *Query) SetStartTime(s time.Time) {
 	q.startTL.Val = s
+	if q.alignGroup && q.groupByOffsetDL != nil {
+		q.groupByOffsetDL.Val = s.Sub(time.Unix(0, 0))
```
nathanielc: This produces queries like `... GROUP BY time(1m, 1484002977413326u)`. Can you perform the modulo so that the offset is always in the range `[-1m, 1m]` when the interval is 1m?

That should make debugging easier for an end user.

goakley: ✔️

Thanks for doing the arithmetic on that, I wasn't confident it was safe to do even though it seemed reasonable.  [-x,0] is covered by [0,x], so I implemented it as such.

query.go
```diff
@@ -192,6 +214,10 @@ func (q *Query) Dimensions(dims []interface{}) error {
 			}
 			// Add time dimension
 			hasTime = true
+			q.groupByOffsetDL.Val = dim.Offset
+			if q.alignGroup {
+				q.groupByOffsetDL.Val = q.StartTime().Sub(time.Unix(0, 0)) - dim.Offset
```
goakley: ✔️

nathanielc:
@goakley Thanks for the latest changes, they look good. Everything seems to be working well, but I am having a hard time finding good use cases for this feature that the normal `.align()` doesn't already cover.

Can you share some examples that lead you to creating this PR? Thanks

goakley:
The trouble is that InfluxDB ignores the query dates when it determines the start times of the group by time intervals.  If you execute a query with a minimum time set anywhere between 2017-01-10T18:06:00 and 2017-01-10T18:06:59 with a `GROUP BY time(1m)`, the interval will start at 2017-01-10T18:06:00, which means a query `time > 2017-01-10T18:06:30 AND time < 2017-01-10T18:08:30` will not only return 3 buckets instead of two, but 2 of those 3 buckets will be "incomplete" since InfluxDB's boundaries force the 1st and 3rd bucket to contain only 30 second of data instead of the full minute.

Here's part of a TICKscript we have for NGINX data processing:
```
var batchquery = batch
    |query('''
    SELECT DIFFERENCE(LAST(handled))-DIFFERENCE(LAST(accepts)) AS value
    FROM "telegraf"."default"."nginx"
    WHERE environment = 'production' GROUP BY time(1m), "host"
  ''')
        .groupBy(time(1m), 'host')
        .period(2m)
        .every(30s)
        .align()
```
We want to calculate the number of unhandled requests for the past minute, and without being able to change the group by time bucket alignment, every other query (the ones at 30s instead of 0s due to the `every(30s)`) gives partially incorrect data because.

pipeline/batch.go
```diff
@@ -178,6 +182,21 @@ func (n *QueryNode) ChainMethods() map[string]reflect.Value {
 //
 // It is recommended to use QueryNode.Align and QueryNode.Offset in conjunction with
 // group by time dimensions so that the time bounds match up with the group by intervals.
+// To automatically align the group by intervals to the start of the query time,
+// use QueryNode.AlignGroup. This is useful in more complex situations, such as when
+// the groupBy period is larger than the query interval.
```
nathanielc: As per your example can you change this to read `such as when the groupBy time period is longer than the query frequency.`

And maybe add something below the TICKscript example about how for every other query the groupBy time bounds would not be aligned (the ones ending on 30s instead of 0s) without the `.alignGroup` call.

Thanks

nathanielc:
You example makes sense to me. I added a few notes where that can be clarified in the doc comments. After that I think this will be ready to be merged. Thanks again!

pipeline/batch.go
goakley: ✔️

goakley:
@nathanielc should I resolve this conflict and squash the fixups?

nathanielc:
@goakley Yes please.

pipeline/batch.go
```diff
@@ -178,6 +182,25 @@ func (n *QueryNode) ChainMethods() map[string]reflect.Value {
...
+//            .every(30s)
+//            .groupBy(time(1m), 'tag1', 'tag2')
+//            .align()
+//            .alignGroup()
+//
+// For the above example, without QueryNode.AlignGroup, every other query issued by Kapacitor
+// (at :30 past the minute) will align to :00 seconds instead of the desired :30 seconds,
+// which would create 6 group by intervals instead of 5, the first and last of which
+// would only have 30 seconds of data instead of a full minute.
+// If QueryNode.Offset is used in conjunction with QueryNode.AlignGroup,
```
nathanielc: The QueryNode.Offset is different from the group by time offset. This should read 

`If the group by time offset (i.e. time(t, offset)) is used in conjunction with QueryNode.AlignGroup,`

goakley: ✔️

nathanielc:
@goakley Thanks!

PR #1162, opened 2017-02-02
Add Pushover integration for Kapacitor

desa:
The PR adds support for pushing alerts to the Pushover API.

- [x] Rebased/mergable
- [x] Tests pass
- [x] CHANGELOG.md updated

### DIFF
--- a/integrations/streamer_test.go
+++ b/integrations/streamer_test.go
@@ -39,6 +39,8 @@ import (
 	"github.com/influxdata/kapacitor/services/opsgenie/opsgenietest"
 	"github.com/influxdata/kapacitor/services/pagerduty"
 	"github.com/influxdata/kapacitor/services/pagerduty/pagerdutytest"
+	"github.com/influxdata/kapacitor/services/pushover"
+	"github.com/influxdata/kapacitor/services/pushover/pushovertest"
 	"github.com/influxdata/kapacitor/services/sensu"
 	"github.com/influxdata/kapacitor/services/sensu/sensutest"
 	"github.com/influxdata/kapacitor/services/slack"
@@ -6990,6 +6992,84 @@ stream
 				Origin:      "override",
 				Service:     []string{"serviceA", "serviceB"},
 				Value:       "10",
+			},
+		},
+	}
+
+	ts.Close()
+	var got []interface{}
+	for _, g := range ts.Requests() {
+		got = append(got, g)
+	}
+
+	if err := compareListIgnoreOrder(got, exp, nil); err != nil {
+		t.Error(err)
+	}
+}
+
+func TestStream_AlertPushover(t *testing.T) {
+	ts := pushovertest.NewServer()
+	defer ts.Close()
+
+	var script = `
+stream
+	|from()
+		.measurement('cpu')
+		.where(lambda: "host" == 'serverA')
+		.groupBy('host')
+	|window()
+		.period(10s)
+		.every(10s)
+	|count('value')
+	|alert()
+		.id('{{ index .Tags "host" }}')
+		.message('kapacitor/{{ .Name }}/{{ index .Tags "host" }} is {{ .Level }} @{{.Time}}')
+		.info(lambda: "count" > 6.0)
+		.warn(lambda: "count" > 7.0)
+		.crit(lambda: "count" > 8.0)
+		.pushover()
+			.sound('siren')
+			.device('mydev')
+			.title('mytitle')
+			.URL('http://example.com')
+			.URLTitle('myurltitle')
+		.pushover()
+			.title('othertitle')
+			.device('otherdev')
+`
+	tmInit := func(tm *kapacitor.TaskMaster) {
+		c := pushover.NewConfig()
+		c.Enabled = true
+		c.URL = ts.URL
+		c.UserKey = "user"
+		c.Token = "KzGDORePKggMaC0QOYAMyEEuzJnyUi"
+		sl := pushover.NewService(c, logService.NewLogger("[test_pushover] ", log.LstdFlags))
+		tm.PushoverService = sl
+	}
+	testStreamerNoOutput(t, "TestStream_Alert", script, 13*time.Second, tmInit)
+
+	exp := []interface{}{
+		pushovertest.Request{
+			PostData: pushovertest.PostData{
+				Token:    "KzGDORePKggMaC0QOYAMyEEuzJnyUi",
+				UserKey:  "user",
+				Message:  "kapacitor/cpu/serverA is CRITICAL @1971-01-01 00:00:10 +0000 UTC",
+				Device:   "mydev",
+				Sound:    "siren",
+				Title:    "mytitle",
+				URL:      "http://example.com",
+				URLTitle: "myurltitle",
+				Priority: 1,
+			},
+		},
+		pushovertest.Request{
+			PostData: pushovertest.PostData{
+				Token:    "KzGDORePKggMaC0QOYAMyEEuzJnyUi",
+				UserKey:  "user",
+				Message:  "kapacitor/cpu/serverA is CRITICAL @1971-01-01 00:00:10 +0000 UTC",
+				Device:   "otherdev",
+				Title:    "othertitle",
+				Priority: 1,
 			},
 		},
 	}
--- a/pipeline/alert.go
+++ b/pipeline/alert.go
@@ -45,6 +45,7 @@ const defaultDetailsTmpl = "{{ json . }}"
 //    * OpsGenie -- Send alert to OpsGenie.
 //    * VictorOps -- Send alert to VictorOps.
 //    * PagerDuty -- Send alert to PagerDuty.
+//    * Pushover -- Send alert to Pushover.
 //    * Talk -- Post alert message to Talk client.
 //    * Telegram -- Post alert message to Telegram client.
 //
@@ -316,6 +317,10 @@ type AlertNode struct {
 	// Send alert to PagerDuty.
 	// tick:ignore
 	PagerDutyHandlers []*PagerDutyHandler `tick:"PagerDuty"`
+
+	// Send alert to Pushover.
+	// tick:ignore
+	PushoverHandlers []*PushoverHandler `tick:"Pushover"`
 
 	// Send alert to Sensu.
 	// tick:ignore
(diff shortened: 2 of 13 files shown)


integrations/streamer_test.go
```diff
@@ -7006,6 +7008,85 @@ stream
 	}
 }
 
+func TestStream_AlertPushocer(t *testing.T) {
```
nathanielc: Typo here TestStream_AlertPushocer -> TestStream_AlertPushover

pipeline/alert.go
```diff
@@ -963,6 +968,70 @@ type SensuHandler struct {
 	*AlertNode
 }
 
+// Send the alert to Pushover.
+//
```
nathanielc: How does one get a token to Pushover? Should we add steps here on how to do that?

pipeline/alert.go
```diff
@@ -963,6 +968,70 @@ type SensuHandler struct {
...
+	// A title for your supplementary URL, otherwise just URL is shown
+	URLTitle string
+
+	// The name of one of the sounds supported by the device clients to override
+	// the user's default sound choice
+	Sound string
+
+	// A Unix timestamp of your message's date and time to display to the user,
+	// rather than the time your message is received by the Pushover API
+	Timestamp bool
```
nathanielc: I am confused on what this does. Does it display the time in unix format? Or does it instruct Pushover to use the receipt time instead of the time of the alert?

services/alert/service.go
```diff
@@ -1028,6 +1033,14 @@ func (s *Service) createHandlerActionFromSpec(spec HandlerActionSpec) (ha handle
 		}
 		h := s.PagerDutyService.Handler(c, s.logger)
 		ha = newPassThroughHandler(h)
+	case "pushover":
+		c := pushover.HandlerConfig{}
```
nathanielc: Should this be a call to DefaultHandlerConfig instead?

services/pushover/config.go
```diff
@@ -0,0 +1,54 @@
...
+		return errors.New("must specify token")
+	}
+
+	if c.Enabled && c.User == "" {
+		return errors.New("must specify user")
+	}
+
+	if c.Enabled && c.URL == "" {
+		return errors.New("must specify url")
+	}
```
nathanielc: Should we nest these three check under a single `if c.Enabled` statement?

services/pushover/pushovertest/pushovertest.go
```diff
@@ -0,0 +1,85 @@
...
+	User      string `json:"user"`
+	Message   string `json:"message"`
+	Device    string `json:"device"`
+	Title     string `json:"title"`
+	URL       string `json:"url"`
+	URLTitle  string `json:"url_title"`
+	Sound     string `json:"sound"`
+	Timestamp string `json:"timestamp"`
+	Priority  int    `json:"priority"`
+}
```
nathanielc: Are these json tags used? Or are all values URL parameters?

services/pushover/service.go
```diff
@@ -0,0 +1,290 @@
...
+	case alert.Info:
+		// -1 to always send as a quiet notification
+		return -1
+	case alert.Warning:
+		// 1 to display as high-priority and bypass the user's quiet hours,
+		return 1
+	case alert.Critical:
+		// 2 to also require confirmation from the user
+		return 2
+	}
```
nathanielc: This mapping should be documented in the pipeline/alert.go comments. As any comments in that package become the auto generated docs for Kapacitor.

services/pushover/service.go
```diff
@@ -0,0 +1,290 @@
...
+}
+
+type handler struct {
+	s      *Service
+	c      HandlerConfig
+	logger *log.Logger
+}
+
+func (s *Service) DefaultHandlerConfig() HandlerConfig {
+	return HandlerConfig{}
```
nathanielc: If the zero value of the config is the default I don't see a need to define  a special function for it.

pipeline/alert.go
desa: It's coming from the [Pushover API](https://pushover.net/api).

```
timestamp - a Unix timestamp of your message's date and time to display to the user, rather than the time your message is received by our API
```

nathanielc: That makes it sound like a time.Time field not a bool.

desa: Ah this may be due to my own confusion on how everything fits together. I used a bool assign a time when the alert is generated.


https://github.com/influxdata/kapacitor/pull/1162/files/fd448c654675d31dc0269bc9802644053ec2b7d6#diff-bca4b6d3553ed032bed7ba3254d7646bR218

```
	if timestamp {
 +		now := time.Now()
 +		p.Timestamp = &now
 +	}
```

Is this okay or should it be a `time.Time`

services/pushover/pushovertest/pushovertest.go
desa: Whoops. My mistake. This is left-over from when I thought the Pushover API accepted JSON.

services/pushover/config.go
desa: We should.

services/alert/service.go
desa: It probably should be. :)

pipeline/alert.go
nathanielc: A bool makes more sense now.

What does Pushover do with the timestamp? Meaning if we specify it how is it different from if we don't? Also if we do specify the timestamp shouldn't it be the time on the alert event not the current time?

integrations/streamer_test.go
desa: Fixed

pipeline/alert.go
desa: I'm not entirely sure what they do with the timestamp. And yup it should definitely be the event time. Didn't realize that that existed. 👍

services/alert/service.go
desa: Removed the `DefaultHandlerConfig` method since it wasn't really doing anything.

services/pushover/service.go
desa: Added

pipeline/alert.go
nathanielc: My reasoning for asking what Pushover does with the timestamp is because I'd like to minimize the number of config options for the integration. If it seems to make sense to always set the time or to never set the time I'd like to pick one way or the other and remove the option. But its hard to say without knowing what it does.

FWIW I have a pushover account so I can try it out it its not clear from their docs etc.

desa: That makes sense. I'll play with it right now to see what it does.

Would you prefer moving it to the config file, or pulling the option completely?

desa: Just played with it, doesn't seem particularly useful.

services/pushover/service.go
```diff
@@ -75,6 +75,7 @@ func (s *Service) Test(options interface{}) error {
 	if !ok {
 		return fmt.Errorf("unexpected options type %t", options)
 	}
+	var ts *time.Time
```
nathanielc: Why are we using a pointer to the time instead of just the time value?

desa: Just removed the logic around timestamps. It didn't seem particularly useful.

The pointer was a remanent from when I thought their API accepted JSON so that when I tagged it with `omitempty` it wouldn't appear in the JSON I sent.

desa:
@nathanielc anything else? Also, what's the typical git workflow for Kapacitor? Should I squash all the commits beside the first one, or are you okay with keeping the individual commits on a PR?

nathanielc:
@desa Yes, please squash all commits into a single commit. The only reason I do not squash a commit from a merge is because it is the commit that adds a vendored dependency, I like to keep those as isolated commits. Doesn't apply here but since you asked. 

Can you add a CHANGELOG entry?

desa:
Makes sense. Yeah, I'll add a CHANGELOG entry.

desa:
Should I add it to the `1.2.1` section, or make a new section for `1.3` and add it there?

nathanielc:
Add it to the current unreleased section which in this case is 1.2.1.

services/pushover/config.go
```diff
@@ -0,0 +1,55 @@
...
+	"net/url"
+
+	"github.com/pkg/errors"
+)
+
+// DefaultPushoverURL is the default URL for the Pushover API.
+const DefaultPushoverURL = "https://api.pushover.net/1/messages.json"
+
+// Config is the [pushover] configuration as defined in the Kapacitor configuration file.
+type Config struct {
```
nathanielc: Can you add a section to the `etc/kapacitor/kapacitor.conf` file for pushover?

services/pushover/service.go
```diff
@@ -0,0 +1,272 @@
...
+	// A title for your supplementary URL, otherwise just URL is shown
+	URLTitle string `mapstructure:"url_title"`
+
+	// The name of one of the sounds supported by the device clients to override
+	// the user's default sound choice
+	Sound string `mapstructure:"sound"`
+
+	// A Unix timestamp of your message's date and time to display to the user,
+	// rather than the time your message is received by the Pushover API
+	Timestamp bool `mapstructure:"timestamp"`
```
nathanielc: This is now unused correct? If so we should remove it from here and the pipeline/alert.go PushoverHandler struct as well.

nathanielc:
@desa I am testing this out locally and I have run into a few issues. 


* I kept getting a response 400 back saying that my user is invalid. I was not sure what my user was,  then I tried using my user token and it started working. We should probably rename the user config option to user_token or at least put int he comments that it can be the user token.
* I had to create a new app token instead of just using my user_token. Normally when I configure Pushover for apps I only need my user_token and do not have to create a specific app token. From what I can tell it looks like we could register Kapacitor as an official integration app and then the user would not have to do anything but provide the user_token.
* When I was getting 400 errors the log would say `failed to understand Pushover response` could you update the logic that inteprets 400 errors from pushover to grab the error message from the JSON?

desa:
> * I kept getting a response 400 back saying that my user is invalid. I was not sure what my user was,  then I tried using my user token and it started working. We should probably rename the user config option to user_token or at least put in the comments that it can be the user token.

I think they refer to it as USER_KEY should I use `user_key` or `user_token`. 

> * I had to create a new app token instead of just using my user_token. Normally when I configure Pushover for apps I only need my user_token and do not have to create a specific app token. From what I can tell it looks like we could register Kapacitor as an official integration app and then the user would not have to do anything but provide the user_token.

I think it's possible, but if we do that then we're responsible for purchasing additional capacity for their API as needed. Maybe I'm missing something.

https://pushover.net/faq#technical-publicapp

> * When I was getting 400 errors the log would say `failed to understand Pushover response` could you update the logic that inteprets 400 errors from pushover to grab the error message from the JSON?

On it.

desa:
@nathanielc I reached out to Pushover and they suggested that each user of Kapacitor should have their own API key. https://pushover.net/faq#technical-distribution

We can make it easier for people to register by creating a Kapacitor app and then allowing people to clone the application.

Do you have an opinion on whether the user values should be renamed to `user_key` or `user_token`?

nathanielc:
Let's use `user_key` as that is what they call it in their docs.

desa:
@nathanielc should be good to go. Let me know if there's anything else I missed.

pipeline/alert.go
```diff
@@ -963,6 +968,74 @@ type SensuHandler struct {
...
+
+// tick:embedded:AlertNode.Pushover
+type PushoverHandler struct {
+	*AlertNode
+
+	// User/Group key of your user (or you), viewable when logged
+	// into the Pushover dashboard. Often referred to as USER_KEY
+	// in the Pushover documentation.
+	// If empty uses the user from the configuration.
+	UserKey string
```
nathanielc: Since the UserKey is sensitive information should we not allow it to be specified in the TICKscript? Meaning we don't want someones TICKscript to have the user key in plain text. This would restrict it so that you can only use the Pushover integration with a single user. But that seems reasonable and expected since you can only configure a single API token anyways.

services/pushover/pushovertest/pushovertest.go
```diff
@@ -0,0 +1,83 @@
...
+	URLTitle string
+	Sound    string
+	Priority int
+}
+
+func NewPostData(v url.Values) (PostData, error) {
+	p := PostData{}
+
+	p.Token = v.Get("token")
+	p.UserKey = v.Get("user")
```
nathanielc: Does Pushover not support placing the token and user keys in the POST data so they are not logged in the query string?

desa: They should be in the request body. I use [PostForm](https://golang.org/pkg/net/http/#PostForm) when I make the request.

> PostForm issues a POST to the specified URL, with data's keys and values URL-encoded as the request body.

pipeline/alert.go
desa: Ah, this is probably due to my own confusion. I was under the impression that UserKey wasn't sensitive. From the [Pushover documentation](https://pushover.net/api#identifiers)

> Once you have an API token for your application, you'll need the user key and optional device name for each user to which you are pushing notifications.

But thinking about it a bit more makes me think that the user token should be secret.

services/pushover/pushovertest/pushovertest.go
nathanielc: Great, I just assumed the would be in the query string, should have double checked :)

desa:
Should be good to go now. (I think)

services/pushover/config.go
```diff
@@ -0,0 +1,57 @@
...
+)
+
+// Config is the [pushover] configuration as defined in the Kapacitor configuration file.
+type Config struct {
+	// Whether Pushover integration is enabled.
+	Enabled bool `toml:"enabled" override:"enabled"`
+	// The Pushover API token.
+	Token string `toml:"token" override:"token,redact"`
+	// The User/Group that will be alerted.
+	UserKey string `toml:"user_key" override:"user_key,redact"`
```
nathanielc: One last small change. Can this be `user-key` instead? The Kapacitor config option name convention is all over the place but currently the `-` is winning so I have decided that should be the standard going forward.

desa: @nathanielc fixed.

PR #1164, opened 2017-02-02
Fix hang in config API if update failed or takes too long

nathanielc:
This fixes a flaky test. Add defensive measures to handleUpdateSection code, increases timeout in test.

Now if the timeout is reached in the test the test will fail immediately instead of hanging and waiting for the longer test timeout.

- [x] Rebased/mergable
- [x] Tests pass
- [x] CHANGELOG.md updated


easyrasta:
The config API update blocks until timeout reached.
ErrC is never updated with http post API. Only works with tests

PR #1166, opened 2017-02-03
Fix parse dbrp param

oiooj:
- [x] Rebased/mergable
- [x] Tests pass
- [ ] CHANGELOG.md updated

If the database name is `collect.dev.test.com`, and the RP name is `default`.

exec:
```
/usr/bin/kapacitor define cpu_idel_alert -type batch -tick /data/tick/cpu-idel.tick -dbrp "collect.dev.test.com"."default"
```
the result:

```
ID                        Type      Status    Executing Databases and Retention Policies
cpu_idel_alert            batch     disabled  false     ["collect"."dev.test.com.default"]
```

After fixed:

```
ID                        Type      Status    Executing Databases and Retention Policies
cpu_idel_alert            batch     disabled  false     ["collect.dev.test.com"."default"]
```

### DIFF
--- a/cmd/kapacitor/main.go
+++ b/cmd/kapacitor/main.go
@@ -561,7 +561,7 @@ func (d *dbrps) Set(value string) error {
 	if value[0] == '"' {
 		dbrp.Database, n = parseQuotedStr(value)
 	} else {
-		n = strings.IndexRune(value, '.')
+		n = strings.LastIndexByte(value, '.')
 		if n == -1 {
 			return errors.New("does not contain a '.', it must be in the form \"dbname\".\"rpname\" where the quotes are optional.")
 		}


oiooj:
@nathanielc

nathanielc:
@oiooj This fix assumes that the rp does not have any dots either, which may not be the case. But this is an issue and I have submitted a PR #1181 that correctly parses quoted dbrp strings.

```
kapacitor define cpu_idel_alert -type batch -tick /data/tick/cpu-idel.tick -dbrp '"collect.dev.test.com"."default"'
```

Running the above command with the fix the PR should work. Notice the single qoutes around the double quotes. Bash removes the double quotes unless they are single quoted.

Issue #1165, opened 2017-02-02
[alerta] token header prefix


allen13:
Back in December I merged a pull request that changed the alerta Authorization prefix from Bearer to Key because of an issue experienced by @emirozer https://github.com/influxdata/kapacitor/issues/1076.

I was running a custom version of alerta at the time so I assumed that I had changed Bearer to Key myself. On further inspection I realized that alerta officialy started using the Bearer prefix a year ago. https://github.com/guardian/alerta/commit/46e52094013f0c19a664f3aeb7ebdf1bfcbf64f4 which means @emirozer was using an older version of alerta.

I have been running a custom version of kapacitor for myself that uses Bearer so I haven't had any issues, but going forward there is a chance that someone will run into this again. I am proposing that the prefix be configurable. The default value would be Bearer.

I will work on a pull request in the coming days that addresses this. I will be sure to alert @emirozer if the pull request gets accepted.

nathanielc:
@allen13 Thanks for being proactive on this. I'll keep an eye out for the PR.

allen13:
@emirozer Once this commit goes through if you are still on the older version of alerta you can add this to your config:
[alerta]
token-prefix = "Key"

allen13:
Thanks getting this merged in so quickly @nathanielc!

v-zhuravlev:
Sorry, speaking about new and old alerta... I actually managed to get kapacitor+alerta running only after I used token-prefix ="Key". before that it was `Alerta: Token is invalid` error all the time.

And alerta version is 5.0.4 which is the latest.

satterly:
I agree with @v-zhuravlev. Systems integrating with Alerta should use "Key \<api-key\>" in the Authorization header, not "Bearer". "Bearer" is reserved for user logins like the web UI and `alerta` command-line tool. I suggest you change the default back to "Key".

allen13:
Whatever the latest default for alerta is should probably be the default here. This is more of an issue with the alerta project changing back and forth.

satterly:
@allen13 Just to avoid confusion, the Alerta authorization header has never changed. It was always "Key" for API keys and "Bearer" for the web UI.

SebastianCzoch:
Hey, I've made PR because I spend some time why proper tokens are not working with Alerta on latest version of kapacitor (upgrade from 1.2 to 1.4).

Alerta doesn't work with `Bearer`, so it shouldn't be a default value.

emirozer:
[...]

PR #1169, opened 2017-02-03
Make the alerta auth token prefix configurable and default it to Bearer.

allen13:
- [X] Rebased/mergable
- [x] Tests pass
- [X] CHANGELOG.md updated

Addresses this issue: https://github.com/influxdata/kapacitor/issues/1165

### DIFF
--- a/services/alerta/config.go
+++ b/services/alerta/config.go
@@ -15,6 +15,9 @@ type Config struct {
 	InsecureSkipVerify bool `toml:"insecure-skip-verify" override:"insecure-skip-verify"`
 	// The authentication token for this notification, can be overridden per alert.
 	Token string `toml:"token" override:"token,redact"`
+	// The prefix for the Authentication field where the token is stored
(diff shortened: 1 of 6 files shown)


services/alerta/config.go
```diff
@@ -15,6 +15,8 @@ type Config struct {
 	InsecureSkipVerify bool `toml:"insecure-skip-verify" override:"insecure-skip-verify"`
 	// The authentication token for this notification, can be overridden per alert.
 	Token string `toml:"token" override:"token,redact"`
+	// The prefix for the Authentication field where the token is stored
+	TokenPrefix string `toml:"token-prefix" override:"token-prefix"`
```
nathanielc: Can you add a comment here that if you are on an older version of Alerta you may want to set this to `Key`? Along those lines can you update the example config in `etc/kapacitor/kapacitor.conf` with this new option and comments as well.

allen13: Done.

PR #1175, opened 2017-02-06
Add error counters to every node type

desa:
- [x] Rebased/mergable
- [x] Tests pass
- [x] CHANGELOG.md updated

### DIFF
--- a/join.go
+++ b/join.go
@@ -277,6 +277,7 @@ func (j *JoinNode) getGroup(p models.PointInterface, groupErrs chan<- error) *gr
 		go func() {
 			err := group.run()
 			if err != nil {
+				j.incrementErrorCount()
 				j.logger.Println("E! join group error:", err)
 				select {
 				case groupErrs <- err:
@@ -342,6 +343,7 @@ func (g *group) collect(i int, p models.PointInterface) error {
 	sets := g.sets[t]
 	if len(sets) == 0 {
 		set = newJoinset(
+			g.j,
 			g.j.j.StreamName,
 			g.j.fill,
 			g.j.fillValue,
@@ -362,6 +364,7 @@ func (g *group) collect(i int, p models.PointInterface) error {
 	}
 	if set == nil {
 		set = newJoinset(
+			g.j,
 			g.j.j.StreamName,
 			g.j.fill,
 			g.j.fillValue,
@@ -466,6 +469,7 @@ func (g *group) emitJoinedSet(set *joinset) error {
 
 // represents a set of points or batches from the same joined time
 type joinset struct {
+	j         *JoinNode
 	name      string
 	fill      influxql.FillOption
 	fillValue interface{}
@@ -486,6 +490,7 @@ type joinset struct {
 }
 
 func newJoinset(
+	n *JoinNode,
 	name string,
 	fill influxql.FillOption,
 	fillValue interface{},
@@ -497,6 +502,7 @@ func newJoinset(
 ) *joinset {
 	expected := len(prefixes)
 	return &joinset{
+		j:         n,
 		name:      name,
 		fill:      fill,
 		fillValue: fillValue,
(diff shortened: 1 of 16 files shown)


alert/HANDLERS.md
```diff
@@ -208,6 +208,7 @@ Add the following snippet to the `service.go` file.
 // Handle takes an event and posts its message to the Foo service chat room.
 func (h *handler) Handle(event alert.Event) {
 	if err := h.s.Alert(h.c.Room, event.State.Message); err != nil {
+		h.incrementErrorCount()
```
nathanielc: This doesn't look right as the handler is not a node and so will not have the incrementErrorCount method.

node.go
```diff
@@ -59,6 +60,9 @@ type Node interface {
 
 	emittedCount() int64
 
+	// TODO: Better name
+	incrementErrorCount()
```
nathanielc: I actually think the name is fine.

node.go
```diff
@@ -276,6 +282,11 @@ func (n *node) emittedCount() (count int64) {
 	return
 }
 
+// node increment error count increments a nodes error_count stat
+func (n *node) incrementErrorCount() {
+	n.statMap.Add(statErrorCount, 1)
```
nathanielc: To avoid the map look up each time you should keep the `errCountVar` from above on the node struct and then just increment it directly.

node.go
```diff
@@ -20,6 +20,7 @@ import (
 
 const (
 	statAverageExecTime = "avg_exec_time_ns"
+	statErrorCount      = "error_count"
```
nathanielc: I think `errors` is a better name for the metric but I don't have a strong opinion.

alert/HANDLERS.md
desa: Lol. May have went into autopilot and added this. Removed now.

node.go
desa: I like `errors` its consistent with the other nodes that have error counts as well.

desa:
I think I've hit every node that has `E!`. 

The command I'm running to check is

```
git grep -B 3 -n E!
```

desa:
I did have a specific question on what to do in this case here

https://github.com/influxdata/kapacitor/pull/1175/commits/419f612cb3237cb3cb630147d01c7711a870e508#diff-d410d6dce1320d105297cb9419d065e3R603

desa:
@nathanielc anything else that's missing from here? (other than the rebase and changelog addition)

desa:
@nathanielc any thoughts on what I should do here 419f612#diff-d410d6dce1320d105297cb9419d065e3R603

join.go
```diff
@@ -599,6 +600,8 @@ BATCH_POINT:
 			}
 			b, ok := batch.(models.Batch)
 			if !ok {
+				// TODO: determine what to do here. No reference to node is available, but
+				//       a possible error case exists. cant make call to incrementErrorCount
```
nathanielc: Pass the reference node down so it can be called.

desa:
@nathanielc Should be good to go.

join.go
```diff
@@ -570,7 +571,7 @@ func (js *joinset) JoinIntoPoint() (models.Point, bool) {
 }
 
 // join all batches the set into a single batch
-func (js *joinset) JoinIntoBatch() (models.Batch, bool) {
+func (js *joinset) JoinIntoBatch(j *JoinNode) (models.Batch, bool) {
```
nathanielc: I'd rather the JoinNode get passed into the `newJoinset` function and become a field on the joinset type instead of passing it into each function of the joinset type.

k8s_autoscale.go
```diff
@@ -17,7 +17,7 @@ import (
 const (
 	statsK8sIncreaseEventsCount = "increase_events"
 	statsK8sDecreaseEventsCount = "decrease_events"
-	statsK8sErrorsCount         = "errors"
+	statsK8sErrorsCount         = "k8s_errors"
```
nathanielc: Should we just remove the k8s_errors stats and use the one counter?

desa: Yeah, should I do that for the other one as well? I particular the `batch` and `eval`

join.go
desa: Sounds good.

PR #1209, opened 2017-02-17
Refactor Alert Service

nathanielc:
This is a refactor of the Alert Service there are a handful of changes:

* Add the concept of External vs Internal Handlers
* Create explicit types for each of the different responsibilities of the alert service
* Remove the concept of Actions, now a handler only ever has a single action.
* General reorg of the code to make it easy to consume/extend.

TODO
- [x] Rebased/mergable
- [x] Tests pass
- [x] CHANGELOG.md updated
- [x] Add test for Aggregate and Publish handlers
- [x] Safely migrate from old handlers to new handlers.


services/alert/api.go
```diff
@@ -0,0 +1,503 @@
...
+func (s sortedTopics) Less(i int, j int) bool { return s[i].ID < s[j].ID }
+func (s sortedTopics) Swap(i int, j int)      { s[i], s[j] = s[j], s[i] }
+
+func (s *apiServer) handleListTopics(w http.ResponseWriter, r *http.Request) {
+	pattern := r.URL.Query().Get("pattern")
+	if err := validatePattern(pattern); err != nil {
+		httpd.HttpError(w, fmt.Sprint("invalid pattern: ", err.Error()), true, http.StatusBadRequest)
+		return
+	}
+	minLevelStr := r.URL.Query().Get("min-level")
```
desa: Not a really an issue since this isn't a hot code path, but `URL.Query()` generates a new map each time its called.

https://golang.org/src/net/url/url.go?s=27419:27447#L967

services/alert/handlers.go
```diff
@@ -341,30 +347,31 @@ func (h *publishHandler) Handle(event alert.Event) {
...
-	if event.State.Level != event.PreviousState().Level {
-		h.next.Handle(event)
-	}
-}
-
-func (h *stateChangesOnlyHandler) SetNext(n alert.Handler) {
-	h.next = n
-}
-func (h *stateChangesOnlyHandler) Close() {}
+// TODO implement state changes only as a match condition
```
desa: What does this mean?

nathanielc:
> it seems like its not possible to create a topic without using it in a TICKscript. Is this correct?

No quite, Topics are created whenever they are referenced so there is no need to explicitly create them. So if they are referenced in a TICKscript they are created as well as if a handler references a topic it will be created.

services/alert/handlers.go
nathanielc: Ah, yes so with the change to not have handler `actions` anymore there is no way to do stateChangesOnly via a handler.

The proposed solution is to use match criteria on handlers. So something like this

```yaml
id: my_handler
topic: system
kind: slack
   options:
       channel: '#alerts'
match: "changed" == TRUE AND "sometag" == 'bob'
```

Where the `match` property is a lambda expression then when it evals to true then the handler handles the alert event. This way we get the same functionality back but in a more consistent manner.

I was thinking we should add the `match` expression logic in another PR so that its can remain as an independent change. Or should we lump it in here so that we maintain feature parity from the get go? Thoughts?

desa:
> No quite, Topics are created whenever they are referenced so there is no need to explicitly create them. So if they are referenced in a TICKscript they are created as well as if a handler references a topic it will be created.

Ah I see. I spent some time confused about this. Here's the process I just went through.


First I created a task with this tickscript

```
var warn_threshold = 4
var crit_threshold = 10

var period = 1h
var every = 1m

var data = stream
  |from()
    .measurement('system')
  |alert()
   .crit(lambda: TRUE)
   .topic('learning')
```

Then 
```
$ kapacitor
```
I noticed the `define-handler` subcommand and then ran
```
$ kapacitor define-handler -h
```
From here I noticed
```
For example:

	Define a handler using the slack.yaml file:

		$ kapacitor define-handler system slack.yaml
```
So I make a `slack.yaml` file
```
id: example
kind: slack
topic: learning
options:
  username: What
```
and then I run
```
$ kapacitor define-handler slack slack.yaml
```
No errors, so moving on.
```
$ kapacitor show-topic learning
ID: learning
Level: CRITICAL
Collected: 1340
Handlers: [ ]
Events:
Event      Level    Message                Date
system:nil CRITICAL system:nil is CRITICAL 14 Mar 17 15:27 EDT
```
Hmm, then I run
```
$ kapacitor list handlers
ID         Topic    Kind
example   slack slack
```
Hmm, so I rerun
```
$ kapacitor define-handler -h
```
This time I notice
```
Must provide both a topic ID and a path to a handler file.
Usage: kapacitor define-handler <topic> <path to handler file>
```
and I notice this time that you specify the topic id as the first positional argument to the command. Then everything was fine.

This confusion was definitely due to me not reading, but the positional arguments vs flags split is different than what I'm used to with kapacitor. The two other `define` subcommands have

```
$ kapacitor define <task ID> [options]
$ kapacitor define-template <template ID> [options]
```
where positional arguments are used for naming things and flags are used to specify details.

nathanielc:
@desa Thanks for the walk through. Agreed lets change it to be more consistent. 


Would this be more clear?
```
kapacitor define-handler slack.yaml -topic system
```

The difference here is that the ID is in the yaml file. Should we move the ID out of the yaml file so that the ID is specified on the command line?

desa:
@nathanielc 
My only issue with pulling it out of the config file is that the `define-handler` is kind of different than the others.

With `define` or `define-template`
```
kapacitor define example -type stream -dbrp telegraf.autogen -tick example.tick
```
defines a single task and reissuing that command with different flags redefines the task `example`. The same thing isn't the case for `define-handler` though. e.g.

```
kapacitor define-handler my_hanlder -topic system -file slack.yaml
```
and
```
kapacitor define-handler my_hanlder -topic other -file slack.yaml
```
define two different handlers.

I keep mulling over different ways of going about it and I cant think of one I like.

Syntacticly, I like 
```
kapacitor define-handler my_hanlder -topic system -file slack.yaml
```
the most. Just because its very similar to how the other `define` commands work. However, it philosophically bothers me, that changing a flag defines a new thing.

nathanielc:
> However, it philosophically bothers me, that changing a flag defines a new thing.

This is why I made it an positional argument. In other words to ID for a handler is topicID and handlerID together.

desa:
maybe changing the name of the command?
```
kapacitor define-topic-handler <topic> <file>
```

nathanielc:
I like that, its a bit verbose but that is want tab completion is for :)

I did a similar thing in the Go client.

desa:
or what about moving the topic into the handler config

```
id: example
kind: slack
topic: learning
options:
  username: What
```

and just having
```
kapacitor define-handler <file>
```

nathanielc:
I like `topic-handler` as it also helps with the other commands like `list` and `delete` as they become

```
kapacitor list topic-handlers <topic> <handler>
kapacitor delete topic-handlers <topic> <handler>
```

And `show-handler` becomes `show-topic-handler <topic> <handler>`

It creates a more consistent treatment of handlers, such that they are always referenced in the scope of a topic.

desa:
👍

desa:
Final thought. Maybe this is what you were already thinking, but what about something like

```
kapacitor define-topic-handler <topic id> <handler id> -config <file>
```
with
```
kind: slack
options:
  username: What
```

nathanielc:
@desa I have pushed an updated CLI, could you look it over and see how UX is?

desa:
For some reason github wont let me directly comment on where some of the issues are.
https://github.com/influxdata/kapacitor/blob/ec9dc79f793f491734efa4dbb0178a593d57fedc/cmd/kapacitor/main.go#L1833
Should say topic-handlers

https://github.com/influxdata/kapacitor/blob/ec9dc79f793f491734efa4dbb0178a593d57fedc/cmd/kapacitor/main.go#L1849
https://github.com/influxdata/kapacitor/blob/ec9dc79f793f491734efa4dbb0178a593d57fedc/cmd/kapacitor/main.go#L1868
Should say topic-handlers

https://github.com/influxdata/kapacitor/blob/ec9dc79f793f491734efa4dbb0178a593d57fedc/cmd/kapacitor/main.go#L2002
Should say topic-handlers

Just a couple copy issues.

Personally, I think I'd prefer having the handler configuration be specified as a flag, but I don't particularly mind it the way that it currently is.

nathanielc:
> Personally, I think I'd prefer having the handler configuration be specified as a flag, but I don't particularly mind it the way that it currently is.

 I started that way and then reversed it because Go is picky about where flags need to be, so it would always have to be the last or first argument which basically turned it into a positional argument. And it is not an optional argument which a flag implies.

I'll fix the other issues.

nathanielc:
@desa The migration path is ready for testing.

You should be able to

*  upgrade with any v1.2 handler and have all functionality continue to work.
* downgrade back to v1.2 and have everything back to how it was

desa:
@nathanielc Just did a manual test and verified that this process works.

Issue #1132, opened 2017-01-17
Define sensu source in the sensu service


ellis2323:
Hello,

I'm using telegraf to monitor network equipments. I have many hundred of equipments. With kapacitor i'm checking the ping and push alerts on my sensu.
The problem is that you define the source in the kapacitor.conf:
services/sensu/service.go line 136 (method prepareData)

	postData["source"] = c.Source

With this config, all the alerts goes into the same client. So i have one client with one alert per equipements.

To change this behavious (https://github.com/ellis2323/kapacitor/), i have a made a little patch. Not sure that the way to do :) I have added a Hostname in the event.State among the ID, message, details and i put it in the 

	postData["source"] = hostname

My alert in TICKS is like this:
```
stream
  |from()
       .measurement('ping')
  |alert()
        .id('{{ .Name }}.{{ index .Tags "url" }}')
        .hostname('{{ index .Tags "url" }}')
        .info(lambda: "average_response_ms" > 10)
        .warn(lambda: "average_response_ms" > 20)
        .crit(lambda: "average_response_ms" > 30)
        .sensu()
```

With this hack, i have one client alert by equipement. It is more readable
[My new result](http://imgur.com/a/wHVgc)

nathanielc:
@ellis2323 Thanks for creating the patch! See https://help.github.com/articles/creating-a-pull-request/ for how to submit a pull request against this repo and we can discuss getting it merged.

ellis2323:
Not really sure if it's the best solution. I think it could be easier not to use my hostname but use in "source" the ID. Or to put the source expr in sensu('{{ index .Tags "url" }}').

henry-megarry-tr:
This would be really helpful to have, when can we expect to see the PR?

ellis2323:
Hi @henry-megarry-tr,

What do you prefer ? Look the alerta config for a coherency. They have way more parameters

henry-megarry-tr:
@ellis2323 

I would do it the same way that they set up the channel for slack. You can have it set in the config, but you can also change it per alert. I'm not very familiar with go but I think you can see how they do it in kapacitor/services/slack/service.go

mega0022:
I have been trying to do it the same way that slack does it, however it wont replace the '{{ index .Tags "url" }}' part for some reason.

https://github.com/mega0022/kapacitor

@nathanielc 
slack.channel doesn't seem to be able to do that either, is there a way I can get it to or is it impossible?

@ellis2323 
[...]

PR #1214, opened 2017-02-21
Better Sensu alerts

ellis2323:
- [ ] Rebased/mergable
- [ ] Tests pass
- [ ] CHANGELOG.md updated


ellis2323:
The first commit has already been integrated. For the auto_resolve, i'm not sure if it is the most common parameter but we use sensu like this.

alert.go
```diff
@@ -45,16 +45,17 @@ const maxWeight = 1.2
...
+	topic        string
+	anonTopic    string
+	handlers     []alert.Handler
+	levels       []stateful.Expression
+	scopePools   []stateful.ScopePool
+	states       map[models.GroupID]*alertState
+	idTmpl       *text.Template
+	messageTmpl  *text.Template
+	detailsTmpl  *html.Template
+	hostnameTmpl *text.Template
```
nathanielc: Since this is specific to the Sensu handler the template logic should exists in services/sensu/service.go.

Have a look at the Alerta handler as it does something quite similar to this.

pipeline/alert.go
```diff
@@ -355,6 +361,7 @@ func newAlertNode(wants EdgeType) *AlertNode {
 		Id:        defaultIDTmpl,
 		Message:   defaultMessageTmpl,
 		Details:   defaultDetailsTmpl,
+		Hostname:  defaultHostnameTmpl,
```
nathanielc: Again instead of making this a global option to all alerts add it the the SensuHandler type which can be found later on in this file.

services/sensu/service.go
```diff
@@ -133,9 +136,10 @@ func (s *Service) prepareData(name, output string, level alert.Level) (*net.TCPA
 
 	postData := make(map[string]interface{})
 	postData["name"] = name
-	postData["source"] = c.Source
+	postData["source"] = hostname
```
nathanielc: Would `source` be a better name for this option instead of `hostname`. While most of the time the source is a hostname the use of the name hostname initially confused me because Kapacitor can only have a single hostname and it doesn't change. But this is not talking about Kapacitor's hostname but rather the hostname of the source.

services/sensu/service.go
```diff
@@ -133,9 +136,10 @@ func (s *Service) prepareData(name, output string, level alert.Level) (*net.TCPA
...
 	postData["output"] = output
 	postData["status"] = status
+	postData["auto_resolve"] = false
```
nathanielc: What does this change do?

ellis2323: If a node has a crit alert then a ok alert, the node is always in crit. In our case, this allow to find Flapping on network equipment.

alert.go
ellis2323: ok

pipeline/alert.go
ellis2323: ok

services/sensu/service.go
ellis2323: ok

services/sensu/service.go
nathanielc: In that case we should create a new property on the SensuHandler called noAutoResolve(), as I think most users will want the auto resolve behavior.

For reference auto_resolve docs https://sensuapp.org/docs/0.23/reference/checks.html

ellis2323: i'm sure you're right. In most cases, auto_resolve must be true.

nathanielc:
Closing, issue #1132 is now fixed in #1314

PR #1216, opened 2017-02-21
add StateDuration node

### LINKED ISSUE #1211: [feature request] Need way to alert on time in certain state
This is a feature request to solve an issue previously discussed on the google group: https://groups.google.com/d/topic/influxdb/1CdjyXrhbjY/discussion
I've been trying to solve this problem for months and have yet to find a solution.

The request is to be able to alert when something has been in a certain state for too long. That google group posting gives one example, but I have others, which are more relevant to other people.
For example:

Using the telegraf ntpq probe, to monitor the ntp daemon on our systems, we want to alert when the host has not had any active peer for a certain amount of time. Warn at 5 minutes, critical at 1 hour.

This logic used to determine this is: when 0 points are seen within the interval where `state_prefix == '*'`.

Another way to look at this is: when the time duration since the last point where `state_prefix == '*'` is greater than the threshold.
However this has one significant different from the above logic in that this approach needs to also handle if `state_prefix == '*'` was never seen, and therefore there is no starting point to calculate duration. But this logic does feel like it would have more potential use cases.

----

One proposal to solve this would be a new node which takes a lambda parameter, and emits a point containing the duration for which the lambda has evaluated as true.

Usage in my scenario would then be:

```
data
  | state_duration(lambda: "state_prefix" != '*').unit(1m)
  | default('state_duration',-1)
  | alert()
    .warn(lambda: "state_duration" >= 5)
    .crit(lambda: "state_duration" >= 60)
```

phemmer:
This is a proposal for a solution to #1211 
It provides a node which evaluates an expression, and adds a field which tracks how long the expression has evaluated as true.

Example usage which alerts when a host has not had any primary ntp peers for over 5 minutes looks like:
```
stream
  |from().measurement('ntpq')|default().tag('state_prefix',' ')
  |groupBy('host')
  |stateDuration(lambda: "state_prefix" != '*').unit(1m)
  |alert()
    .warn(lambda: "state_duration" > 5)
    .crit(lambda: "state_duration" > 60)
```

If this is an acceptable solution to the problem, I can finish up the PR by adding tests, docs, etc.

Closes #1211 #1225 

- [x] Rebased/mergable
- [x] Tests pass
- [x] CHANGELOG.md updated

### DIFF
--- a/state_tracking.go
+++ b/state_tracking.go
@@ -1,0 +1,190 @@
+package kapacitor
+
+import (
+	"fmt"
+	"log"
+	"time"
+
+	"github.com/influxdata/kapacitor/models"
+	"github.com/influxdata/kapacitor/pipeline"
+	"github.com/influxdata/kapacitor/tick/ast"
+	"github.com/influxdata/kapacitor/tick/stateful"
+)
+
+type stateTracker interface {
+	track(p models.BatchPoint, inState bool) interface{}
+	reset()
+}
+
+type stateTrackingGroup struct {
+	stateful.Expression
+	stateful.ScopePool
+	tracker stateTracker
+}
+
+type StateTrackingNode struct {
+	node
+	lambda *ast.LambdaNode
+	as     string
+
+	newTracker func() stateTracker
+	groups     map[models.GroupID]*stateTrackingGroup
+}
+
+func (stn *StateTrackingNode) group(g models.GroupID) (*stateTrackingGroup, error) {
+	stg := stn.groups[g]
+	if stg == nil {
+		stg = &stateTrackingGroup{}
+
+		var err error
+		stg.Expression, err = stateful.NewExpression(stn.lambda.Expression)
+		if err != nil {
+			return nil, fmt.Errorf("Failed to compile expression: %v", err)
+		}
+
+		stg.ScopePool = stateful.NewScopePool(stateful.FindReferenceVariables(stn.lambda.Expression))
+
+		stg.tracker = stn.newTracker()
+
+		stn.groups[g] = stg
+	}
+	return stg, nil
+}
+
+func (stn *StateTrackingNode) runStateTracking(_ []byte) error {
+	switch stn.Provides() {
+	case pipeline.StreamEdge:
+		for p, ok := stn.ins[0].NextPoint(); ok; p, ok = stn.ins[0].NextPoint() {
+			stn.timer.Start()
+			stg, err := stn.group(p.Group)
+			if err != nil {
+				return err
+			}
+
+			pass, err := EvalPredicate(stg.Expression, stg.ScopePool, p.Time, p.Fields, p.Tags)
+			if err != nil {
+				stn.incrementErrorCount()
+				stn.logger.Println("E! error while evaluating expression:", err)
+				stn.timer.Stop()
+				continue
+			}
+
+			p.Fields = p.Fields.Copy()
+			p.Fields[stn.as] = stg.tracker.track(models.BatchPointFromPoint(p), pass)
+
+			stn.timer.Stop()
+			for _, child := range stn.outs {
+				err := child.CollectPoint(p)
+				if err != nil {
+					return err
+				}
+			}
+		}
+	case pipeline.BatchEdge:
+		for b, ok := stn.ins[0].NextBatch(); ok; b, ok = stn.ins[0].NextBatch() {
+			stn.timer.Start()
+
+			stg, err := stn.group(b.Group)
+			if err != nil {
+				return err
+			}
+			stg.tracker.reset()
+
+			for i := 0; i < len(b.Points); {
+				p := &b.Points[i]
+				pass, err := EvalPredicate(stg.Expression, stg.ScopePool, p.Time, p.Fields, p.Tags)
+				if err != nil {
+					stn.incrementErrorCount()
+					stn.logger.Println("E! error while evaluating epression:", err)
+					b.Points = append(b.Points[:i], b.Points[i+1:]...)
+					continue
+				}
+				i++
+
+				p.Fields = p.Fields.Copy()
+				p.Fields[stn.as] = stg.tracker.track(*p, pass)
+			}
+
+			stn.timer.Stop()
+			for _, child := range stn.outs {
+				err := child.CollectBatch(b)
+				if err != nil {
+					return err
+				}
+			}
+		}
+	}
+	return nil
+}
+
+type stateDurationTracker struct {
+	sd *pipeline.StateDurationNode
+
+	startTime time.Time
+}
+
+func (sdt *stateDurationTracker) reset() {
+	sdt.startTime = time.Time{}
+}
+
+func (sdt *stateDurationTracker) track(p models.BatchPoint, inState bool) interface{} {
+	if !inState {
+		sdt.startTime = time.Time{}
+		return float64(-1)
+	}
+
+	if sdt.startTime.IsZero() {
+		sdt.startTime = p.Time
+	}
+	return float64(p.Time.Sub(sdt.startTime)) / float64(sdt.sd.Unit)
+}
+
+func newStateDurationNode(et *ExecutingTask, sd *pipeline.StateDurationNode, l *log.Logger) (*StateTrackingNode, error) {
+	if sd.Lambda == nil {
+		return nil, fmt.Errorf("nil expression passed to StateDurationNode")
+	}
+	stn := &StateTrackingNode{
+		node:   node{Node: sd, et: et, logger: l},
+		lambda: sd.Lambda,
+		as:     sd.As,
+
+		groups:     make(map[models.GroupID]*stateTrackingGroup),
+		newTracker: func() stateTracker { return &stateDurationTracker{sd: sd} },
+	}
+	stn.node.runF = stn.runStateTracking
+	return stn, nil
+}
+
+type stateCountTracker struct {
+	count int64
+}
+
+func (sct *stateCountTracker) reset() {
+	sct.count = 0
+}
+
+func (sct *stateCountTracker) track(p models.BatchPoint, inState bool) interface{} {
+	if !inState {
+		sct.count = 0
+		return int64(-1)
+	}
+
+	sct.count++
+	return sct.count
+}
+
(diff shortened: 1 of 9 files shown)


state_duration.go
```diff
@@ -0,0 +1,138 @@
...
+			scopePool := sdn.scopePool(p.Group)
+
+			pass, err := EvalPredicate(expr, scopePool, p.Time, p.Fields, p.Tags)
+			if err != nil {
+				sdn.logger.Println("E! error while evaluating expression:", err)
+				continue
+			}
+
+			if pass {
+				if sdn.stateStart.IsZero() {
```
nathanielc: Should a stateStartTime be recorded for each group? This way if you group by host the output is the duration that each host has been in the state. 

If you want the duration that any host has been in the state then do not groupBy host.

state_duration.go
```diff
@@ -0,0 +1,138 @@
...
+				err := child.CollectPoint(p)
+				if err != nil {
+					return err
+				}
+			}
+		}
+	case pipeline.BatchEdge:
+		for b, ok := sdn.ins[0].NextBatch(); ok; b, ok = sdn.ins[0].NextBatch() {
+			sdn.timer.Start()
+			sdn.stateStart = time.Time{}
```
nathanielc: Does it make sense for the startTime to be constrained to fit within a single batch? 

My initial thought would be to treat batch just like stream an basically ignore the batch boundaries since the state could span a boundary. 

Did you need the behavior for batch to constrain the start time to within the batch?

state_duration.go
phemmer: Yes, it should track each group separately. I was under the mistaken idea that kapacitor would create a separate node object for each grouping, but after looking at other examples, it looks like I have to track by the `p.Group`.  
Will adjust, thanks :-)

state_duration.go
phemmer: I reset it within each batch as this is what other nodes seemed to be doing. This seems like the correct behavior as well, as with things like `|window().period(1m).every(30s)`, the batches will overlap, and result in really whacked duration values.

nathanielc: ok, that makes sense.

state_duration.go
nathanielc: Yeah, I considered that design at some point and may come back to it later but for now each node is responsible for dealing with groups as dictated by the node.

phemmer:
What should be the behavior when we hit a point where the expression doesn't evaluate (such as referencing a field which is null)? The options I see
1. Do nothing and pass the point through unchanged.
2. Treat it as if the expression evaluated as false (reset duration).
3. Drop the point.

I don't like option 2. I'm torn between options 1 & 3.

nathanielc:
I would go with option 3 as that is how the eval node behaves. In addition to dropping the point the error is logged and a counter is incremented indicating there was an error.

nathanielc:
Also as a heads up this PR https://github.com/influxdata/kapacitor/pull/1175 is about to be merged which changes how the error should be counted. So I would recommend rebasing your work off that branch for a bit or just wait till that PR is merged. I expect the PR will be merged either tonight or tomorrow sometime.

desa:
Just merged  #1175

phemmer:
Adjusted for comments, and added tests & docs.

pipeline/state_duration.go
```diff
@@ -0,0 +1,58 @@
...
+)
+
+// Compute the duration of a given state.
+// The state is defined via a lambda expression. For each consecutive point for
+// which the expression evaluates as true, the state duration will be
+// incremented by the duration between points. When a point evaluates as false,
+// the state duration is reset to 0.
+//
+// The state duration will be added as an additional field to each point. If the
+// expression evaluates as false, the field will not be added. If the expression
```
nathanielc: > If the expression evaluates as false, the field will not be added.

What is the purpose of this behavior? I expected it to just set the field with a value of 0 whenever the expression was false.

phemmer: The first point where the expression evaluates as true is 0. After all on the first point, there was no previous point, so it has been evaluating as true for 0s.
You might then argue we use a default of -1 instead, but even this can be incorrect as someone could do `order by time desc` and now the durations are negative instead of positive.
The only way to be absolutely certain the statement evaluated as false is to just not add the field at all.

integrations/data/TestBatch_StateDuration.0.brpl
```diff
@@ -0,0 +1,2 @@
+{"name":"cpu","tags":{"host":"serverA"},"points":[{"time":"1971-01-01T00:00:00Z","fields":{"value":97.1}},{"time":"1971-01-01T00:00:01Z","fields":{"value":96.6}},{"time":"1971-01-01T00:00:02Z","fields":{"value":83.6}},{"time":"1971-01-01T00:00:03Z","fields":{"value":99.1}}]}
+{"name":"cpu","tags":{"host":"serverB"},"points":[{"time":"1971-01-01T00:00:00Z","fields":{"value":47}},{"time":"1971-01-01T00:00:01Z","fields":{"value":95.1}},{"time":"1971-01-01T00:00:02Z","fields":{"value":null}},{"time":"1971-01-01T00:00:03Z","fields":{"value":96.1}}]}
```
nathanielc: If it helps I updated the parser a while back to not care about newlines, so you can break this data up and format it like normal JSON. Not a needed change but useful to know.

pipeline/state_duration.go
phemmer: We could make it default to `-1`, as this will satisfy the 99.9999% of use cases, and then just provide a `.default()` attribute to change it to whatever the user desires.

nathanielc: OK, so why is it important to distinguish between the first point that evaluates to true and other points that evaluate to false? My understanding was that the user cares about the duration of the event and so the user wouldn't trigger anything on the first point anyways. 

Said a different way under what circumstances would the user want to trigger an alert when the stateDuration is 0? Seems like the alert condition would always be "state_duration" > 0 . But even if not you could always reevaluate the same state later if it was important. 

For example:

```
stream
  |from().measurement('ntpq')|default().tag('state_prefix',' ')
  |groupBy('host')
  |stateDuration(lambda: "state_prefix" != '*').unit(1m)
  |default().field('state_duration',-1)
  |alert()
    // Trigger info on first occurrence.
    .info(lambda: "state_duration" == 0 AND "state_prefix" != '*') 
    // Warn after 5 minutes.
    .warn(lambda: "state_duration" > 5)
    // Critical after 1 hour.
    .crit(lambda: "state_duration" > 60)
```

From my perspective it seems that the node would be easier to reason about if it always set the field with either a 0 or the current duration, but I want to understand your perspective.

phemmer: Maybe someone wants alert when the duration is less than X. Silly example, but maybe a beer factory wants to turn on an indicator light when a keg is full, but hasn't sat long enough to finish fermenting (light on means "fermentation in progress"). If it's 0 when expression evaluates as false (keg is empty), the light would be on.

Or maybe someone wants to have a critical state when the expression is false, warning state when it's less than X, and OK state when it's greater than X.

[Defensive programming](https://en.wikipedia.org/wiki/Defensive_programming). I just want it to be able to handle all use cases, even the ones I can't think of.

nathanielc: Sounds good, lets default the value to -1 like you suggest. That way a value of >= 0 indicates that the state condition is true and can always be interpreted as the duration of the state. By using -1 as the default a value is always set so a `|default` node is not required and then it can easily be changed on via an eval node.

Thanks for working through this with me.

phemmer:
Updated to set the resulting duration to `-1` if expression evaluates as false.

How do we want to handle #1225? It's still under discussion, but do we want to implement any of the proposals in here?

desa:
@phemmer I think include it at a part of this PR.

nathanielc:
@desa @phemmer

Quote from @desa in the other issue

>My only concern is if I'm a user, looking for this functionality, would I think to use stateDuration? I think not. 

As I result I think having a separate node that is similar in functionality but mirrors the behavior is best. 

It keeps it simple so that each node has a single purpose. And since they could mirror each other in property names etc then it makes it so that learning one teaches you how to use the other.

phemmer:
What should it be called?

desa:
I think `stateCount` works.

phemmer:
I'm not fond of the name `stateCount` as it's not indicative of the fact that it's a consecutive count, instead of a total count. But meh, naming is easy to change so I've updated the PR to include the feature with that name.

Also instead of a massive copy/paste of code with a tiny amount of different lines, I redesigned them to share the same core code. There are unfortunately a dozen different ways of doing this, so the implementation I have up now is just what I thought would be easiest to understand & maintain. But open to other ideas.

state_tracking.go
```diff
@@ -0,0 +1,181 @@
...
+	"log"
+	"time"
+
+	"github.com/influxdata/kapacitor/models"
+	"github.com/influxdata/kapacitor/pipeline"
+	"github.com/influxdata/kapacitor/tick/ast"
+	"github.com/influxdata/kapacitor/tick/stateful"
+)
+
+type stateTrackingGrouper interface {
```
nathanielc: I like how they are combined and share the common implementation. To me the name grouper was confusing at first since I thought it was responsible for grouping the data not tracking each group. Maybe this should be called `stateTracker` and the method is called `track` instead of `pointResult`? That way the `tracker` `track`s which is how Go interfaces should read. Then `newGrouper` would be `newTracker` etc.

Thoughts? Not super important as like you said naming is easy to change later.

state_tracking.go
```diff
@@ -0,0 +1,181 @@
...
+			pass, err := EvalPredicate(stg.Expression, stg.ScopePool, p.Time, p.Fields, p.Tags)
+			if err != nil {
+				stn.incrementErrorCount()
+				stn.logger.Println("E! error while evaluating expression:", err)
+				continue
+			}
+
+			p.Fields = p.Fields.Copy()
+			p.Fields[stn.as] = stg.grouper.pointResult(models.BatchPointFromPoint(p), pass)
+			fmt.Printf("VALUES: %#v\n", p.Fields)
```
nathanielc: This line should be removed and I think there are a few others like it.

phemmer:
@nathanielc Requested change have been made.

PR #1221, opened 2017-02-23
WIP: Add cardinality stat to each node type.

desa:
This PR introduces a new stat `cardinality` that tracks the number of groups that exist for a given node for each relevant node type. 

- [x] Add new tests for cardinality tracking
- [x] Fill out k8s cardinality test

- [x] Rebased/mergable
- [x] Tests pass
- [x] CHANGELOG.md updated

### DIFF
--- a/alert.go
+++ b/alert.go
@@ -58,12 +58,15 @@ type AlertNode struct {
 	messageTmpl *text.Template
 	detailsTmpl *html.Template
 
+	cardinalityMu sync.RWMutex
+
 	alertsTriggered *expvar.Int
 	oksTriggered    *expvar.Int
 	infosTriggered  *expvar.Int
 	warnsTriggered  *expvar.Int
 	critsTriggered  *expvar.Int
 	eventsDropped   *expvar.Int
+	nodeCardinality *expvar.IntFuncGauge
 
 	bufPool sync.Pool
 
@@ -436,6 +439,12 @@ func newAlertNode(et *ExecutingTask, n *pipeline.AlertNode, l *log.Logger) (an *
 		n.History = 2
 	}
 	an.states = make(map[models.GroupID]*alertState)
+	an.nodeCardinality = expvar.NewIntFuncGauge(func() int {
+		an.cardinalityMu.RLock()
+		l := len(an.states)
+		an.cardinalityMu.RUnlock()
+		return l
+	})
 
 	// Configure flapping
 	if n.UseFlapping {
@@ -479,6 +488,9 @@ func (a *AlertNode) runAlert([]byte) error {
 	a.eventsDropped = &expvar.Int{}
 	a.statMap.Set(statsCritsTriggered, a.critsTriggered)
 
+	// a.nodeCardinality is assigned in newAlertNode
+	a.statMap.Set(statsCardinalityGauge, a.nodeCardinality)
+
 	switch a.Wants() {
 	case pipeline.StreamEdge:
 		for p, ok := a.ins[0].NextPoint(); ok; p, ok = a.ins[0].NextPoint() {
@@ -488,7 +500,10 @@ func (a *AlertNode) runAlert([]byte) error {
 				return err
 			}
 			var currentLevel alert.Level
-			if state, ok := a.states[p.Group]; ok {
+			a.cardinalityMu.RLock()
+			state, ok := a.states[p.Group]
+			a.cardinalityMu.RUnlock()
+			if ok {
 				currentLevel = state.currentLevel()
 			} else {
 				// Check for previous state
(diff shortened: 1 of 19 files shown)


expvar/expvar.go
```diff
@@ -46,6 +46,26 @@ func (v *Int) IntValue() int64 {
 	return atomic.LoadInt64(&v.i)
 }
 
+// IntGauge is a 64-bit integer variable that satisfies the expvar.Var interface.
+type IntFuncGauge struct {
+	ValueF func() int
```
nathanielc: Should this be `func() int64` so that the function itself can handle the casting?

expvar/expvar.go
```diff
@@ -46,6 +46,26 @@ func (v *Int) IntValue() int64 {
 	return atomic.LoadInt64(&v.i)
 }
 
+// IntGauge is a 64-bit integer variable that satisfies the expvar.Var interface.
+type IntFuncGauge struct {
```
nathanielc: Can you update the comment for the name `IntFuncGague`?

desa:
My only hesitation for putting it on the node is that its not relevant for all nodes. I suppose having a `cardinality=0` for those nodes isn't that bad though.

expvar/expvar.go
desa: yup

desa:
@nathanielc assuming that latest commit goes green, it should be in a reviewable state.

There's a bit of ugliness around setting the `n.nodeCardinality.ValueF` because of a race.

node.go
```diff
@@ -83,8 +84,10 @@ type node struct {
 	timer      timer.Timer
 	statsKey   string
 	statMap    *kexpvar.Map
+	statMu     sync.Mutex
 
-	nodeErrors *kexpvar.Int
+	nodeErrors      *kexpvar.Int
+	nodeCardinality *kexpvar.IntFuncGauge
```
desa: That works and avoids the locking nonsense. I'll do that.

desa: @nathanielc fixed.

alert.go
```diff
@@ -558,7 +571,10 @@ func (a *AlertNode) runAlert([]byte) error {
 			var highestPoint *models.BatchPoint
 
 			var currentLevel alert.Level
-			if state, ok := a.states[b.Group]; ok {
+			a.cardinalityMu.RLock()
```
nathanielc: Since this is common and references a field on the alert node I think it would be best to wrap this into a small function.

combine.go
```diff
@@ -60,6 +65,14 @@ func (t timeList) Less(i, j int) bool { return t[i].Before(t[j]) }
 func (t timeList) Swap(i, j int)      { t[i], t[j] = t[j], t[i] }
 
 func (n *CombineNode) runCombine([]byte) error {
+	valueF := func() int64 {
+		n.cardinalityMu.RLock()
```
nathanielc: Should we name this expressionsByGroupMu? I have typically named mutexes using the convention <name of thing being locked>Mu so that it is obvious, or just `mu` if it is global to the entire struct.

Thoughts? Maybe we leave the name alone but above in the struct place the cardinalityMu next to the things being locked so its clear.

sample.go
```diff
@@ -73,10 +85,14 @@ func (s *SampleNode) shouldKeep(group models.GroupID, t time.Time) bool {
 		keepTime := t.Truncate(s.duration)
 		return t.Equal(keepTime)
 	} else {
+		s.cardinalityMu.RLock()
```
nathanielc: Seems like we should just get the write lock for the whole section here.

combine.go
desa: I like the idea of naming it after the thing its locking. The reason I named it `cardinalityMu` was so that it would be universal across each of the nodes.

sample.go
desa: Good call.

alert.go
desa: I think this pattern is pretty common throughout, should I add it to the other nodes that have this pattern as well?

alert.go
```diff
@@ -580,7 +591,8 @@ func (a *AlertNode) runAlert([]byte) error {
 			var highestPoint *models.BatchPoint
 
 			var currentLevel alert.Level
-			if state, ok := a.states[b.Group]; ok {
+			state, ok := a.getAlertState(b.Group)
+			if ok {
```
nathanielc: This can go back inside the if statement now.

alert.go
```diff
@@ -488,7 +498,8 @@ func (a *AlertNode) runAlert([]byte) error {
 				return err
 			}
 			var currentLevel alert.Level
-			if state, ok := a.states[p.Group]; ok {
+			state, ok := a.getAlertState(p.Group)
```
nathanielc: And this one too.

desa: Fixed

PR #1238, opened 2017-03-02
Use ProxyFromEnvironment for all HTTP Transport where needed

### LINKED ISSUE #841: Conditional proxy support
Our environment requires the use of a proxy for outbound internet access. Thus we need proxy support for alert destinations like PagerDuty.
Go does support the `http_proxy` and `https_proxy` environment variables, however these appear to control ALL http calls, which is not what we want as we do not want to use the proxy server for access to influxdb.

So we need some way of controlling what uses the proxy.
2 solutions I can think of:

1) Add a parameter to alert handlers which might need a proxy. E.G. `alert().pagerDuty.proxy('http://foo:bar@proxyhost:3128')`.

or

```
[pagerduty]
    enabled = true
    proxy = "http://foo:bar@proxyhost:3128"
```

2) Add a per-site proxy list to the config. E.G.

```
[proxy]
    events.pagerduty.com = "http://foo:bar@proxyhost:3128"
```

^if toml is even capable of such a config

desa:
Most services and clients throughout Kapacitor use the Default HTTP
Client which uses the Default HTTP Transport, which uses HTTP proxies as
directed by the $HTTP_PROXY and $NO_PROXY (or $http_proxy and $no_proxy)
environment variables.

For the services that use a custom Transport, we use
http.ProxyFromEnvironment.

The downside to this is that all future implementers will need to copy
this pattern

Fixes #841 (though its not conditional proxy as the issue has requested)

- [x] Rebased/mergable
- [x] Tests pass
- [x] CHANGELOG.md updated

### DIFF
--- a/client/v1/client.go
+++ b/client/v1/client.go
@@ -149,6 +149,7 @@ func New(conf Config) (*Client, error) {
 	}
 
 	tr := &http.Transport{
+		Proxy: http.ProxyFromEnvironment,
 		TLSClientConfig: &tls.Config{
 			InsecureSkipVerify: conf.InsecureSkipVerify,
 		},
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@
     The behavior of the node changes slightly in order to provide a consistent fix to the bug.
     The breaking change is that now, the time of the points returned are from the right hand or current point time, instead of the left hand or previous point time.
 - [#1353](https://github.com/influxdata/kapacitor/issues/1353): Fix panic in scraping TargetManager.
+- [#1238](https://github.com/influxdata/kapacitor/pull/1238): Use ProxyFromEnvironment for all outgoing HTTP traffic.
 
 ## v1.3.0-beta2 [2017-05-01]
 
--- a/influxdb/client.go
+++ b/influxdb/client.go
@@ -123,7 +123,9 @@ func NewHTTPClient(conf Config) (*HTTPClient, error) {
 		return nil, errors.Wrap(err, "invalid URLs")
 	}
 	if conf.Transport == nil {
-		conf.Transport = &http.Transport{}
+		conf.Transport = &http.Transport{
+			Proxy: http.ProxyFromEnvironment,
+		}
 	}
 	c := &HTTPClient{
 		config: conf,
--- a/services/alerta/service.go
+++ b/services/alerta/service.go
@@ -38,6 +38,7 @@ func NewService(c Config, l *log.Logger) *Service {
 	s.configValue.Store(c)
 	s.clientValue.Store(&http.Client{
 		Transport: &http.Transport{
+			Proxy:           http.ProxyFromEnvironment,
 			TLSClientConfig: &tls.Config{InsecureSkipVerify: c.InsecureSkipVerify},
 		},
 	})
@@ -115,6 +116,7 @@ func (s *Service) Update(newConfig []interface{}) error {
 		s.configValue.Store(c)
 		s.clientValue.Store(&http.Client{
 			Transport: &http.Transport{
+				Proxy:           http.ProxyFromEnvironment,
 				TLSClientConfig: &tls.Config{InsecureSkipVerify: c.InsecureSkipVerify},
 			},
 		})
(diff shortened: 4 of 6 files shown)


client/v1/client.go
```diff
@@ -145,10 +145,9 @@ func New(conf Config) (*Client, error) {
 		}
 	}
 
-	tr := &http.Transport{
-		TLSClientConfig: &tls.Config{
-			InsecureSkipVerify: conf.InsecureSkipVerify,
-		},
+	tr := http.DefaultTransport.(*http.Transport)
+	tr.TLSClientConfig = &tls.Config{
```
nathanielc: Since the DefaultTransport is a pointer aren't we modifying it globally? I think that will create a race condition as well.

desa: Ah good catch.

What do you think of doing
```go
tr := *http.DefaultTransport.(*http.Transport)
...
	return &Client{
		url:       u,
		userAgent: conf.UserAgent,
		httpClient: &http.Client{
			Timeout:   conf.Timeout,
			Transport: &tr,
		},
   }
```

desa:
As for testing this, I'm not sure what we should do. On a cursory look, I cant find anything on this. Maybe I'm not looking for the right thing?

client/v1/client.go
desa: Turns out this approach wont work either. `go vet` is yelling at me.

desa:
```
assignment copies lock value to tr: net/http.Transport contains sync.Mutex
```

desa:
@nathanielc At this point I'm not sure of what we should do. I could add a `ProxyFromEnvironment` to each of the `Transport` structs, but I really don't like that. As mentioned previously, there's not really a good way to enforce that we keep to this pattern in the future. The only reasonable thing I can think of currently would be to create an kapacitor `http` package that we use throughout. At least that way we can notice that someone is using the `default` http package and not the `khttp` package. That may also be a bad idea, I'm not sure.

desa:
@nathanielc I haven't tried it manually. I'll be sure to do that before this is merged.

desa:
@nathanielc just confirmed that this works and updated the changelog. Also as a side note, I came across https://github.com/elazarl/goproxy which was really useful for setting up a basic proxy.

phemmer:
> Fixes #841 (though its not conditional proxy as the issue has requested)

Can you clarify how I would use this to address the issue of needing to use a proxy for PagerDuty, but not for InfluxDB?

desa:
@phemmer it doesn't solve the conditional proxy issue. This PR added basic proxy support for all services that make outgoing http requests. The reason it was associated with #841 was that it spurred the initial discussion about fixing general proxy support.

phemmer:
Can I get #841 re-opened then?

desa:
@phemmer yeah no problem.

joaquin386:
I have the Kapacitor 1.3.0 version .... in which version will this be included?

desa:
@joaquin386 it should be functional in 1.3.0

joaquin386:
@desa the https_proxy & http_proxy & no_proxy enviroment variables (#841) right?

joaquin386:
put it on the file /etc/default/kapacitor 
https_proxy="http://proxy:8000"
http_proxy="http://proxy:8000"
no_proxy="localhost"
But seems not to make any difference. It is not picking up the config for the env variables.

desa:
@joaquin386  I tested it locally on my laptop, with kapacitor running in the foreground. I'll see if I can verify this with it running as a service with systemd

joaquin386:
2017/05/30 22:01:17 E! failed to send event to Slack Post https://hooks.slack.com/services/X/X: dial tcp 52.XXX.XX.XX:443: getsockopt: connection timed out

Even in foregorund I can not set it up:
#export https_proxy=http://proxy:8000
#export http_proxy=http://proxy:8000
#export no_proxy=localhost,localhost.localdomain
#kapacitord -config kapacitor.conf

desa:
@joaquin386 found the issue. Not sure how I missed it in the first place.

The slack handler is missing `ProxyFromEnvironment` https://github.com/influxdata/kapacitor/blob/master/services/slack/service.go#L39-L41

Could you test one of the other services to ensure that they are working correctly.

joaquin386:
PagerDuty is working fine. I started testing Slack and it didn't work.
So I suppose the fix needs to be push to the Slack part.
One bug less :)

desa:
@joaquin386 👍  I'll get a fix out right now https://github.com/influxdata/kapacitor/pull/1415

PR #1242, opened 2017-03-06
Revert "Revert "WIP: Add cardinality stat to each node type.""

desa:
Reverts influxdata/kapacitor#1240


nathanielc:
Adding a bit of clarification for this commit...

Changes were merged but before the merge we forgot to rebase and squash the changes into a single commit. Immediately thereafter a revert commit was made to undo the changes. But a revert commit only reverts the changes not the history. Since we rather keep history intact we just want the changes back, the easiest way to do that is to revert the revert commit.

If on the other hand the original merge was in error because the content of the merge was bad we would expect a new PR and merge with fresh changes not just a revert-revert commit. 

To reiterate in in this case the original changes are good and we just want them back.

Issue #117, opened 2015-12-22
Add headers to the POST notifications


panda87:
It would be great if we will be able to add headers to the POST alert handler call, since we can add auth and other tokens to the request and make it more generic for many 3rd party services.

Thanks
D.

nathanielc:
Something like this:

```
.alert()
   .post('http://example.com/path/to/something')
      .header('Authorization', 'Basic QWxhZGRpbjpvcGVuIHNlc2FtZQ==')
```

This creates a new problem in that you probably don't want to expose the auth in plain text in the TICKscript. As a result we probably need a method for grabbing values from the config or environment:

Maybe?
```
.alert()
   .post('http://example.com/path/to/something')
      .header('Authorization', config('example.com-password'))
```

with corresponding config:

```
[tickscript]
   example.com-password = 'Basic QWxhZGRpbjpvcGVuIHNlc2FtZQ=='
```

Or env based without config:

```
.alert()
   .post('http://example.com/path/to/something')
      .header('Authorization', env('EXAMPLE_COM_PASSWORD'))
```

Follow up question: Would being able to set header as static strings (independent of whether they are directly in the TICKscript, or come from config/env) be enough flexibility to do what you need? Or would you need to be able to specify the header value as a function of the data?

panda87:
I personally prefer the ENV option since it's allow me to work close with Vault, and grab the secrets from there directly to the ENV var of the system.

fcastello:
Is this already implemented?

nathanielc:
@fcastello nope

earthnut:
@nathanielc  is this in the latest version ?

nhaugo:
Lets support both the env and config options.

desa:
After discussion, the option that we've decided on going with the following

```
|alert()
  .post()
    .endpoint('example')
```

 Which requires the configuration

```toml
[[post]]
endpoint = "example"
url = "http://example.com"
headers = { Authorization = "...", others = "..." }
```

This way values may be set through the configuration file, environment variables (via the config), and the API.

desa:
@panda87 @fcastello We just recently implemented this feature. It would be great if you could give it a try. We've added a section to the configuration file like so

```
 [[httppost]]
    endpoint = "example"
    url = "http://example.com"
    headers = { Example = "result" }
    basic-auth = { username = "my-user", password = "my-pass" }
```

Then to use it in a tickscript you simply

```
...
|alert()
 ...
 .post()
[...]

PR #1254, opened 2017-03-10
WIP: Add httppost service (add headers to POST alerts)

desa:
This PR allows users to create POST request `endpoints` for an alert node where arbitrary HTTP headers may be set. It works as a fix for https://github.com/influxdata/kapacitor/issues/117.

Todo:
- [x] Add comments
- [x] Cleanup 
- [x] Test how ENV variables work with toml maps

- [x] Rebased/mergable
- [x] Tests pass
- [x] CHANGELOG.md updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased [unreleased]
 
+### Features
+
+- [#117](https://github.com/influxdata/kapacitor/issues/117): Add headers to alert POST requests.
+
 ### Bugfixes
 
 - [#1294](https://github.com/influxdata/kapacitor/issues/1294): Fix bug where batch queries would be missing all fields after the first nil field.
--- a/alert.go
+++ b/alert.go
@@ -17,6 +17,7 @@ import (
 	"github.com/influxdata/kapacitor/pipeline"
 	alertservice "github.com/influxdata/kapacitor/services/alert"
 	"github.com/influxdata/kapacitor/services/hipchat"
+	"github.com/influxdata/kapacitor/services/httppost"
 	"github.com/influxdata/kapacitor/services/opsgenie"
 	"github.com/influxdata/kapacitor/services/pagerduty"
 	"github.com/influxdata/kapacitor/services/pushover"
@@ -128,15 +129,6 @@ func newAlertNode(et *ExecutingTask, n *pipeline.AlertNode, l *log.Logger) (an *
 	}).Parse(n.Details)
 	if err != nil {
 		return nil, err
-	}
-
-	// Construct alert handlers
-	for _, post := range n.PostHandlers {
-		c := alertservice.PostHandlerConfig{
-			URL: post.URL,
-		}
-		h := alertservice.NewPostHandler(c, l)
-		an.handlers = append(an.handlers, h)
 	}
 
 	for _, tcp := range n.TcpHandlers {
@@ -360,6 +352,16 @@ func newAlertNode(et *ExecutingTask, n *pipeline.AlertNode, l *log.Logger) (an *
 			c.Sound = p.Sound
 		}
 		h := et.tm.PushoverService.Handler(c, l)
+		an.handlers = append(an.handlers, h)
+	}
+
+	for _, p := range n.HTTPPostHandlers {
+		c := httppost.HandlerConfig{
+			URL:      p.URL,
+			Endpoint: p.Endpoint,
+			Headers:  p.Headers,
+		}
+		h := et.tm.HTTPPostService.Handler(c, l)
 		an.handlers = append(an.handlers, h)
 	}
 
--- a/alert/types.go
+++ b/alert/types.go
@@ -17,6 +17,18 @@ type Event struct {
 	previousState EventState
 }
 
+func (e Event) AlertData() Data {
+	return Data{
+		ID:       e.State.ID,
+		Message:  e.State.Message,
+		Details:  e.State.Details,
+		Time:     e.State.Time,
+		Duration: e.State.Duration,
+		Level:    e.State.Level,
+		Data:     e.Data.Result,
+	}
+}
+
 func (e Event) PreviousState() EventState {
 	return e.previousState
 }
@@ -153,4 +165,16 @@ func ParseLevel(s string) (l Level, err error) {
 type TopicState struct {
 	Level     Level
 	Collected int64
+}
+
+// Data is a structure that contains relevant data about an alert event.
+// The structure is intended to be JSON encoded, providing a consistent data format.
+type Data struct {
+	ID       string        `json:"id"`
+	Message  string        `json:"message"`
+	Details  string        `json:"details"`
+	Time     time.Time     `json:"time"`
+	Duration time.Duration `json:"duration"`
+	Level    Level         `json:"level"`
+	Data     models.Result `json:"data"`
 }
--- a/etc/kapacitor/kapacitor.conf
+++ b/etc/kapacitor/kapacitor.conf
@@ -275,6 +275,20 @@ default-retention-policy = ""
   # The URL for the Pushover API.
   url = "https://api.pushover.net/1/messages.json"
 
+##########################################
+# Configure Alert POST request Endpoints
+
+# As ENV variables:
+# KAPACITOR_HTTPPOST_0_ENDPOINT = "example"
+# KAPACITOR_HTTPPOST_0_URL = "http://example.com"
+# KAPACITOR_HTTPPOST_0_HEADERS_Example = "header"
+
+# [[httppost]]
+#   endpoint = "example"
+#   url = "http://example.com"
+#   headers = { Example = "your-key" }
+#   basic-auth = { username = "my-user", password = "my-pass" }
+
 [slack]
   # Configure Slack.
   enabled = false
--- a/http_post.go
+++ b/http_post.go
@@ -2,6 +2,7 @@ package kapacitor
 
 import (
 	"encoding/json"
+	"fmt"
 	"log"
 	"net/http"
 	"sync"
@@ -9,24 +10,42 @@ import (
 	"github.com/influxdata/kapacitor/bufpool"
 	"github.com/influxdata/kapacitor/models"
 	"github.com/influxdata/kapacitor/pipeline"
+	"github.com/influxdata/kapacitor/services/httppost"
 )
 
 type HTTPPostNode struct {
 	node
-	c   *pipeline.HTTPPostNode
-	url string
-	mu  sync.RWMutex
-	bp  *bufpool.Pool
+	c        *pipeline.HTTPPostNode
+	endpoint *httppost.Endpoint
+	mu       sync.RWMutex
+	bp       *bufpool.Pool
 }
 
 // Create a new  HTTPPostNode which submits received items via POST to an HTTP endpoint
 func newHTTPPostNode(et *ExecutingTask, n *pipeline.HTTPPostNode, l *log.Logger) (*HTTPPostNode, error) {
+
 	hn := &HTTPPostNode{
 		node: node{Node: n, et: et, logger: l},
 		c:    n,
 		bp:   bufpool.New(),
-		url:  n.Url,
 	}
+
+	// Should only ever be 0 or 1 from validation of n
+	if len(n.URLs) == 1 {
+		e := httppost.NewEndpoint(n.URLs[0], nil, httppost.BasicAuth{})
+		hn.endpoint = e
+	}
+
+	// Should only ever be 0 or 1 from validation of n
+	if len(n.HTTPPostEndpoints) == 1 {
+		endpointName := n.HTTPPostEndpoints[0].Endpoint
+		e, ok := et.tm.HTTPPostService.Endpoint(endpointName)
+		if !ok {
+			return nil, fmt.Errorf("endpoint '%s' does not exist", endpointName)
+		}
+		hn.endpoint = e
+	}
+
 	hn.node.runF = hn.runPost
 	return hn, nil
 }
@@ -72,14 +91,27 @@ func (h *HTTPPostNode) postRow(group models.GroupID, row *models.Row) {
 	defer h.bp.Put(body)
 	err := json.NewEncoder(body).Encode(result)
 	if err != nil {
+		h.incrementErrorCount()
+		h.logger.Printf("E! failed to marshal row data json: %v", err)
+		return
+	}
+	req, err := h.endpoint.NewHTTPRequest(body)
+	if err != nil {
+		h.incrementErrorCount()
 		h.logger.Printf("E! failed to marshal row data json: %v", err)
 		return
 	}
 
-	resp, err := http.Post(h.url, "application/json", body)
+	req.Header.Set("Content-Type", "application/json")
+	for k, v := range h.c.Headers {
+		req.Header.Set(k, v)
+	}
+	resp, err := http.DefaultClient.Do(req)
 	if err != nil {
+		h.incrementErrorCount()
 		h.logger.Printf("E! failed to POST row data: %v", err)
 		return
 	}
 	resp.Body.Close()
+
 }
--- a/integrations/streamer_test.go
+++ b/integrations/streamer_test.go
@@ -33,6 +33,8 @@ import (
 	"github.com/influxdata/kapacitor/services/alerta/alertatest"
 	"github.com/influxdata/kapacitor/services/hipchat"
 	"github.com/influxdata/kapacitor/services/hipchat/hipchattest"
+	"github.com/influxdata/kapacitor/services/httppost"
+	"github.com/influxdata/kapacitor/services/httppost/httpposttest"
 	k8s "github.com/influxdata/kapacitor/services/k8s/client"
 	"github.com/influxdata/kapacitor/services/logging/loggingtest"
 	"github.com/influxdata/kapacitor/services/opsgenie"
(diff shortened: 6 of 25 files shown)


alert.go
```diff
@@ -129,13 +130,13 @@ func newAlertNode(et *ExecutingTask, n *pipeline.AlertNode, l *log.Logger) (an *
...
-		h := alertservice.NewPostHandler(c, l)
-		an.handlers = append(an.handlers, h)
-	}
+	//for _, post := range n.PostHandlers {
+	//	c := alertservice.PostHandlerConfig{
+	//		URL: post.URL,
+	//	}
+	//	h := alertservice.NewPostHandler(c, l)
+	//	an.handlers = append(an.handlers, h)
+	//}
```
desa: Remove

alert.go
```diff
@@ -355,6 +356,18 @@ func newAlertNode(et *ExecutingTask, n *pipeline.AlertNode, l *log.Logger) (an *
...
 	}
 
+	for _, p := range n.PostHandlers {
+		c := alertpost.HandlerConfig{}
+		if p.URL != "" {
+			c.URL = p.URL
+		}
+		if p.Endpoint != "" {
+			c.Endpoint = p.Endpoint
+		}
```
desa: Not necessary.

Just do
```go
c.URL = p.URL
c.Endpoint = p.Endpoint
```

etc/kapacitor/kapacitor.conf
```diff
@@ -275,6 +275,14 @@ default-retention-policy = ""
...
   url = "https://api.pushover.net/1/messages.json"
 
+##########################################
+# Configure Alert POST request Endpoints
+
+# [[alertpost]]
+#   endpoint = "example"
+#   url = "http://example.com"
+#   headers = { Authorization = "your-key" }
+
```
desa: How does this work with environment variables?

integrations/streamer_test.go
```diff
@@ -9523,6 +9597,7 @@ func testStreamer(
 	tm.HTTPDService = httpdService
 	tm.TaskStore = taskStore{}
 	tm.DeadmanService = deadman{}
+	tm.PostService = alertpost.NewService(nil, logService.NewLogger("[alertpost] ", log.LstdFlags))
```
desa: Change to

```go
tm.AlertPostService = alertpost.NewService(nil, logService.NewLogger("[alertpost] ", log.LstdFlags))
```

pipeline/alert.go
```diff
@@ -485,6 +507,9 @@ type PostHandler struct {
 	// The POST URL.
 	// tick:ignore
 	URL string
+
+	// Name of the endpoint to be used, as defined in the configuration file
```
desa: as is defined by the configuration.

server/server.go
```diff
@@ -457,6 +459,18 @@ func (s *Server) appendPushoverService() {
 	s.AppendService("pushover", srv)
 }
 
+func (s *Server) appendPostService() {
```
desa: `appendAlertPostService`

services/alert/service.go
```diff
@@ -108,6 +109,9 @@ type Service struct {
 	PushoverService interface {
 		Handler(pushover.HandlerConfig, *log.Logger) alert.Handler
 	}
+	PostService interface {
```
desa: `AlertPostService`

services/alertpost/service.go
```diff
@@ -0,0 +1,155 @@
...
+
+func (s *Service) Test(options interface{}) error {
+	return nil
+}
+
+type testOptions struct{}
+
+func (s *Service) TestOptions() interface{} {
+	return &testOptions{}
+}
```
desa: Can this be removed?

task_master.go
```diff
@@ -105,6 +106,9 @@ type TaskMaster struct {
 	PushoverService interface {
 		Handler(pushover.HandlerConfig, *log.Logger) alert.Handler
 	}
+	PostService interface {
```
desa: `AlertPostService`

desa:
@nathanielc this should be in a reviewable state.

etc/kapacitor/kapacitor.conf
nathanielc: Good question, have we tried it? I think there is support for maps in our env code but we should test it. Even add a test to the server package for it.

services/alertpost/service.go
nathanielc: No, it needs to implement the interface. Also would having and endpoint option make sense so you can test each configured endpoint?

services/alertpost/service.go
```diff
@@ -0,0 +1,155 @@
...
+
+	err = json.NewEncoder(body).Encode(ad)
+	if err != nil {
+		h.logger.Printf("E! failed to marshal alert data json: %v", err)
+		return
+	}
+
+	// Create the HTTP request
+	var req *http.Request
+	if h.url != "" {
```
nathanielc: Should we validate in pipeline that you do not have both endpoint and url specified?

services/alertpost/service.go
```diff
@@ -0,0 +1,155 @@
...
+// The structure is intended to be JSON encoded, providing a consistent data format.
+type AlertData struct {
+	ID       string        `json:"id"`
+	Message  string        `json:"message"`
+	Details  string        `json:"details"`
+	Time     time.Time     `json:"time"`
+	Duration time.Duration `json:"duration"`
+	Level    alert.Level   `json:"level"`
+	Data     models.Result `json:"data"`
+}
```
nathanielc: Hmm, should we move the AlertData struct out of the alert/service package and into the alert package? The purpose of the struct is to provide a consistent view of how alert data is exposed so I'd rather there only be one definition of the type.

desa: Yeah, that sounds like a good idea.

services/alertpost/service.go
desa: How would we validate this in pipeline?

desa:
@nathanielc this should be in a reviewable state.

http_post.go
```diff
@@ -2,21 +2,24 @@ package kapacitor
...
 )
 
 type HTTPPostNode struct {
 	node
-	c   *pipeline.HTTPPostNode
-	url string
-	mu  sync.RWMutex
-	bp  *bufpool.Pool
+	c        *pipeline.HTTPPostNode
+	handlers []*httppost.Handler
```
nathanielc: It feels odd that the HTTPPost node is dealing with alert handlers. Maybe we should rename the type  `httppost.Handler` to `httppost.Endpoint` since all it really is is an endpoint. This change also helps clarify the purpose of the type since currently the Handler type is not used for its `Handle` method, in this context.

http_post.go
```diff
@@ -27,6 +30,30 @@ func newHTTPPostNode(et *ExecutingTask, n *pipeline.HTTPPostNode, l *log.Logger)
 		bp:   bufpool.New(),
 		url:  n.Url,
 	}
+
+	if n.Url != "" {
+		c := httppost.HandlerConfig{}
+		c.URL = n.Url
+		h := et.tm.HTTPPostService.Handler(c, l)
```
nathanielc: The type change is also useful here since you can return the concrete type so you do not have to type assert it. Or event add a new method for getting Endpoints.

http_post.go
```diff
@@ -68,18 +95,40 @@ func (h *HTTPPostNode) postRow(group models.GroupID, row *models.Row) {
...
-	defer h.bp.Put(body)
-	err := json.NewEncoder(body).Encode(result)
-	if err != nil {
-		h.logger.Printf("E! failed to marshal row data json: %v", err)
-		return
-	}
+	// Technically we don't need the wait group, but
+	// having it hear will help contain any leaky go
+	// routines
+	var wg sync.WaitGroup
```
nathanielc: We should probably defer wg.Wait() here instead of at the end of the function.

http_post.go
```diff
@@ -68,18 +95,40 @@ func (h *HTTPPostNode) postRow(group models.GroupID, row *models.Row) {
...
+				h.logger.Printf("E! failed to marshal row data json: %v", err)
+				return
+			}
 
-	resp, err := http.Post(h.url, "application/json", body)
-	if err != nil {
-		h.logger.Printf("E! failed to POST row data: %v", err)
-		return
+			req.Header.Set("Content-Type", "application/json")
+			resp, err := http.DefaultClient.Do(req)
```
nathanielc: Does this break the HTTP proxy support we added? Or had that not been merged yet?

alert/types.go
```diff
@@ -154,3 +154,15 @@ type TopicState struct {
 	Level     Level
 	Collected int64
 }
+
+// AlertData is a structure that contains relevant data about an alert event.
+// The structure is intended to be JSON encoded, providing a consistent data format.
+type AlertData struct {
```
nathanielc: The move was necessary since we moved the post handler out of the alertservice package correct?

pipeline/http_post.go
```diff
@@ -14,13 +14,48 @@ package pipeline
 type HTTPPostNode struct {
 	chainnode
 
+	HTTPPostHandlers []*HTTPPostHandler `tick:"Endpoint"`
```
nathanielc: This needs a `tick:ignore`

pipeline/http_post.go
```diff
@@ -14,13 +14,48 @@ package pipeline
...
 	// tick:ignore
 	Url string
 }
 
-func newHTTPPostNode(wants EdgeType, url string) *HTTPPostNode {
+func newHTTPPostNode(wants EdgeType, urls ...string) *HTTPPostNode {
+	var url string
+	// TODO: What to do if len(urls) > 1?
+	if len(urls) != 0 {
+		url = urls[0]
```
nathanielc: We should add a validate method to this node and we can return the error there.

pipeline/node.go
```diff
@@ -327,8 +327,8 @@ func (n *chainnode) HttpOut(endpoint string) *HTTPOutNode {
 }
 
 // Creates an HTTP Post node that POSTS received data to the provided HTTP endpoint.
-func (n *chainnode) HttpPost(url string) *HTTPPostNode {
-	h := newHTTPPostNode(n.provides, url)
+func (n *chainnode) HttpPost(url ...string) *HTTPPostNode {
```
nathanielc: We should document that the URL  is optional if the endpoint is specified, hence the variadic signature.

server/config.go
```diff
@@ -257,6 +263,72 @@ func (c *Config) ApplyEnvOverrides() error {
...
+
+		}
+		mapValue.SetMapIndex(key, reflect.ValueOf(floatValue))
+	case reflect.Struct:
+		if err := c.applyEnvOverridesToStruct(prefix, s); err != nil {
+			return err
+		}
+	}
+	return nil
+}
```
nathanielc: Do we have a test for this code?

services/httppost/service.go
```diff
@@ -0,0 +1,200 @@
...
+	s.mu.RUnlock()
+	return
+}
+
+func (s *Service) Update(newConfigs []interface{}) error {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	for _, nc := range newConfigs {
+		if c, ok := nc.(Config); ok {
+			s.endpoints[c.Endpoint] = c
```
nathanielc: This does look like it is going to work.

The config updates need to effect live code, meaning if you change the URL for an endpoint then the existing alert handlers and httpPost nodes need to update to the new URL. This should be easily accomplished by locking the `Handler|Endpoint` type NewHTTPRequest method etc.

As it currently stands the configs are read once so these changes never propagate to running code.

Let me know if I can help clarify this more.

services/httppost/service.go
```diff
@@ -0,0 +1,200 @@
...
+	ad := alertDataFromEvent(event)
+
+	err = json.NewEncoder(body).Encode(ad)
+	if err != nil {
+		h.logger.Printf("E! failed to marshal alert data json: %v", err)
+		return
+	}
+
+	// Create the HTTP request
+	var req *http.Request
```
nathanielc: Why not use the NewHTTPRequest method here?

http_post.go
```diff
@@ -2,21 +2,24 @@ package kapacitor
...
 
 type HTTPPostNode struct {
 	node
-	c   *pipeline.HTTPPostNode
-	url string
-	mu  sync.RWMutex
-	bp  *bufpool.Pool
+	c        *pipeline.HTTPPostNode
+	handlers []*httppost.Handler
+	url      string
```
nathanielc: I believe this field is now unused.

alert/types.go
nathanielc: Also should we rename the type to `Data` to remove the stutter?

desa:
@nathanielc 

> My suggestion is the have the httppost service expose an Endpoint type that the httpPost node consumes instead of the handler interface.

I'll fix that right now.jk

> My second request is about how live config updates propagate. In the current state changes to endpoints do not effect running tasks and they should. We need a way for the running alert handlers and httpPost nodes to get these live updates. I think the type separating above will help with this, since we can lock the NewHTTPRequest method and then modify the endpoints instead of having to hand out new ones.

I think this currently works.
Using this tickscript
```
var data = stream
    |from()
        .measurement('cpu')
    |window()
        .period(5s)
        .every(5s)
    |httpPost()
        .endpoint('hello')
```

with configuration
```
...
[[httppost]]
    endpoint = "hello"
    url = "http://localhost:7070/"
    headers = { Authorization = "your-key" }
```

and a telegraf reporting every second.

when I first enable the task, the data is written to a mock server I have on `7070`. If I then run
```
 curl -XPOST http://localhost:9092/kapacitor/v1/config/httppost/hello -d '{ "set": { "url": "http://localhost:5050"} }'
```
the data no longer written to `7070` and is written to `5050`.

services/httppost/service.go
desa: I could be crazy, but I think this works.

http_post.go
desa: It hasn't been merged yet, but I also think that it won't break. The default client, I'm pretty sure, does proxy from environment.

http_post.go
desa: Fixed.

http_post.go
desa: Fixed

http_post.go
desa: Fixed.

http_post.go
desa: Fixed

pipeline/node.go
desa: Where should this be documented?

server/config.go
desa: Nope, I should add that. Is there a place where the logic for array env elements lives?

services/httppost/service.go
desa: Fized

alert/types.go
desa: Yup. Stutter removed.

pipeline/http_post.go
desa: fixed

pipeline/http_post.go
```diff
@@ -14,48 +21,61 @@ package pipeline
...
-	}
 	return &HTTPPostNode{
 		chainnode: newBasicChainNode("http_post", wants, wants),
-		Url:       url,
+		urls:      urls,
 	}
 }
 
+// tick:ignore
+func (p *HTTPPostNode) Validate() error {
```
nathanielc: This should be named `validate` in order to be automatically called in the parsing execution steps.

pipeline/node.go
desa: Added it in the description of HTTPPostNode.

pipeline/http_post.go
```diff
@@ -14,48 +21,61 @@ package pipeline
...
+	if len(p.urls) > 2 {
+		return fmt.Errorf("httpPost expects 0 or 1 arguments, got %v", len(p.urls))
+	}
+	return nil
+}
+
+func (p *HTTPPostNode) URL() (u string, err error) {
```
nathanielc: While this is a nice clean up we can't do this as the public API of this struct is callable via TICKscript. We have to treat pipeline struct types as config structs with no behavior.

services/httppost/service.go
```diff
@@ -27,25 +27,55 @@ func NewService(c Configs, l *log.Logger) *Service {
...
+
+func (e *Endpoint) NewHTTPRequest(body io.Reader) (req *http.Request, err error) {
+	if e.url != "" {
+		req, err = http.NewRequest("POST", e.url, body)
+		if err != nil {
+			return nil, fmt.Errorf("failed to create POST request: %v", err)
+		}
+		return
+	}
+	c, ok := e.s.endpoint(e.name)
```
nathanielc: This is the bit I missed before, we reread the config on every NewHTTPRequest, which is why it is working.

services/httppost/service.go
```diff
@@ -27,25 +27,55 @@ func NewService(c Configs, l *log.Logger) *Service {
...
+	if !ok {
+		return nil, fmt.Errorf("endpoint does not exist: %v", e.name)
+	}
+
+	return c.NewRequest(body)
+}
+
+func (s *Service) Endpoint(c HandlerConfig) *Endpoint {
```
nathanielc: Why do we need to create a HandlerConfig to get an Endpoint?

services/httppost/service.go
```diff
@@ -27,25 +27,55 @@ func NewService(c Configs, l *log.Logger) *Service {
 	return s
 }
 
+// Only one of name and url should be non-empty
+type Endpoint struct {
+	s      *Service
+	name   string
+	url    string
```
nathanielc: As a preference I would like to see Endpoint as having a url field and not know its name. Then for the Endpoints created with just a URL use the URL as the name of the endpoint in the map. 

This way all endpoints are the same and just reference a URL and a set of headers.

Also, where are the headers? Are they in the config struct still?

services/httppost/service.go
nathanielc: Although my suggestion above invalidate this way of doing it.

pipeline/http_post.go
desa: ah dang :( I'll come up with another solution.

nathanielc: Simplest thing is to define this method inside `http_post.go` like 


```
func  nodeURL(p *pipeline.HTTPPostNode) (u string, err error) {
}
```

it needs a better name but you get the idea.

services/httppost/service.go
desa: > This way all endpoints are the same and just reference a URL and a set of headers.

I like this idea, but I think we might need the name for the case when an endpoint hasn't been defined, but is used in a tickscript. When the task in initialized and the endpoint doesn't currently exist and then we create the endpoint. When the configuration is updated how do we know which `httppost.Endpoints` need to have their values updated? I could be over thinking this.

> Also, where are the headers? Are they in the config struct still?

Yeah they're only in the config struct.

services/httppost/httpposttest/server.go
```diff
@@ -0,0 +1,53 @@
...
+		s.data = append(s.data, req)
+	}))
+	s.ts = ts
+	s.URL = ts.URL
+	return s
+}
+
+type Request struct {
+	MatchingHeaders bool
+	Data            alert.Data
```
nathanielc: This seems incorrect now. The posted data could be either alert data or just a raw row of data. Should we try and capture both or  alternately we could just capture the entire body and let the consumer of this test type decode it.

services/httppost/service.go
```diff
@@ -0,0 +1,253 @@
...
+		if c, ok := nc.(Config); ok {
+			if err := c.Validate(); err != nil {
+				return err
+			}
+			e, ok := s.endpoints[c.Endpoint]
+			if !ok {
+				s.endpoints[c.Endpoint] = NewEndpoint(c.URL, c.Headers)
+				continue
+			}
+			e.SetURL(c.URL)
```
nathanielc: I think we want these Set calls to be atomic together. Meaning there is a slight chance that the endpoint is used right after the URL is modified but before the Headers are which could cause issues. Maybe we should just have an Update method that takes a Config? Lots of ways to do it, but all updates should happen with one lock/unlock of the mutex.

services/httppost/service.go
```diff
@@ -0,0 +1,253 @@
...
+	req.Header.Set("Content-Type", "application/json")
+	resp, err := http.DefaultClient.Do(req)
+	if err != nil {
+		h.logger.Printf("E! failed to POST alert data: %v", err)
+		return
+	}
+	resp.Body.Close()
+}
+
+func alertDataFromEvent(event alert.Event) alert.Data {
```
nathanielc: Doesn't this function already exist in another package?

desa: yup it is, I'll export and use it in this package.

nathanielc: Great, you can move it to the alert package now too. Maybe even just make it a method of the Event type, but what ever looks right to you.

CHANGELOG.md
```diff
@@ -69,6 +69,7 @@ kapacitor define-handler system aggregate_by_1m.yaml
 - [#507](https://github.com/influxdata/kapacitor/issues/507): Add API endpoint for performing Kapacitor database backups.
 - [#1132](https://github.com/influxdata/kapacitor/issues/1132): Adding source for sensu alert as parameter
 - [#1346](https://github.com/influxdata/kapacitor/pull/1346): Add discovery and scraping services.
+- [#117](https://github.com/influxdata/kapacitor/issues/117): Add headers to alert POST requests.
```
nathanielc: Can you move this up to the next section now that beta1 was cut?

desa: Which section? the unreleased section?

Issue #1203, opened 2017-02-16
[feature request] Operator for verifying whether a value is present


rossmcdonald:
Add a function that is able to be referenced in a lambda expression that returns a boolean based on whether the field/tag key is present. For example:

```
|eval(lambda: is_present("myfield"))
```

Which would return TRUE if `myfield` is a valid identifier and FALSE otherwise.

/cc @kostasb

desa:
Going  to close this as it was delivered a while ago.

PR #1284, opened 2017-03-28
WIP: Add type signature to built-in functions

desa:
This PR introduces a type system for builtin functions. Was built to make https://github.com/influxdata/kapacitor/issues/1203 easier to implement.

Todo:
- [x] Squash Commits

Todo Type System:
- [x] Add Changelog entry
- [x] Remove execution state from `NodeEvaluator.Type` function signature

Todo Missing and isPresent:
- [x] Add Changelog entry
- [x] Fix problem with is `stateDuration` and `stateCount`
- [x] General Cleanup
- [x] Better errors
- [x] Add tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@
 ### Features
 
 - [#1299](https://github.com/influxdata/kapacitor/pull/1299): Allowing sensu handler to be specified 
+- [#1284](https://github.com/influxdata/kapacitor/pull/1284): Add type signatures to Kapacitor functions.
+- [#1203](https://github.com/influxdata/kapacitor/issues/1203): Add `isPresent` operator for verifying whether a value is present (part of [#1284](https://github.com/influxdata/kapacitor/pull/1284)).
 
 ### Bugfixes
 
--- a/expr.go
+++ b/expr.go
@@ -5,6 +5,7 @@ import (
 	"time"
 
 	"github.com/influxdata/kapacitor/models"
+	"github.com/influxdata/kapacitor/tick/ast"
 	"github.com/influxdata/kapacitor/tick/stateful"
 )
 
@@ -17,11 +18,12 @@ func EvalPredicate(se stateful.Expression, scopePool stateful.ScopePool, now tim
 		return false, err
 	}
 
-	b, err := se.EvalBool(vars)
-	if err != nil {
+	// for function signature check
+	if _, err := se.Type(vars); err != nil {
 		return false, err
 	}
-	return b, nil
+
+	return se.EvalBool(vars)
 }
 
 // fillScope - given a scope and reference variables, we fill the exact variables from the now, fields and tags.
@@ -50,8 +52,9 @@ func fillScope(vars *stateful.Scope, referenceVariables []string, now time.Time,
 		}
 		if !isFieldExists && !isTagExists {
 			if !vars.Has(refVariableName) {
-				return fmt.Errorf("no field or tag exists for %s", refVariableName)
+				vars.Set(refVariableName, ast.MissingValue)
 			}
+
 		}
 	}
 
--- a/services/alert/handlers.go
+++ b/services/alert/handlers.go
@@ -481,6 +481,21 @@ func (h *matchHandler) Handle(event alert.Event) {
 	}
 }
 
+var changedFuncSignature = map[stateful.Domain]ast.ValueType{}
+var levelFuncSignature = map[stateful.Domain]ast.ValueType{}
+var nameFuncSignature = map[stateful.Domain]ast.ValueType{}
+var taskNameFuncSignature = map[stateful.Domain]ast.ValueType{}
+var durationFuncSignature = map[stateful.Domain]ast.ValueType{}
+
+func init() {
+	d := stateful.Domain{}
+	changedFuncSignature[d] = ast.TBool
+	levelFuncSignature[d] = ast.TInt
+	nameFuncSignature[d] = ast.TString
+	taskNameFuncSignature[d] = ast.TString
+	durationFuncSignature[d] = ast.TDuration
+}
+
 func (h *matchHandler) match(event alert.Event) (bool, error) {
 	// Populate scope
 	h.scope.Reset()
@@ -488,47 +503,62 @@ func (h *matchHandler) match(event alert.Event) (bool, error) {
 	h.logger.Printf("D! match %+v", h)
 
 	if h.usesChanged {
-		h.scope.SetDynamicMethod(changedFunc, func(self interface{}, args ...interface{}) (interface{}, error) {
-			if len(args) != 0 {
-				return nil, fmt.Errorf("%s takes no arguments", changedFunc)
-			}
-			return event.State.Level != event.PreviousState().Level, nil
+		h.scope.SetDynamicFunc(changedFunc, &stateful.DynamicFunc{
+			F: func(args ...interface{}) (interface{}, error) {
+				if len(args) != 0 {
+					return nil, fmt.Errorf("%s takes no arguments", changedFunc)
+				}
+				return event.State.Level != event.PreviousState().Level, nil
+			},
+			Sig: changedFuncSignature,
 		})
 	}
 
 	if h.usesLevel {
-		h.scope.SetDynamicMethod(levelFunc, func(self interface{}, args ...interface{}) (interface{}, error) {
-			if len(args) != 0 {
-				return nil, fmt.Errorf("%s takes no arguments", levelFunc)
-			}
-			return int64(event.State.Level), nil
+		h.scope.SetDynamicFunc(levelFunc, &stateful.DynamicFunc{
+			F: func(args ...interface{}) (interface{}, error) {
+				if len(args) != 0 {
+					return nil, fmt.Errorf("%s takes no arguments", levelFunc)
+				}
+				return int64(event.State.Level), nil
+			},
+			Sig: levelFuncSignature,
 		})
 	}
 
 	if h.usesName {
-		h.scope.SetDynamicMethod(nameFunc, func(self interface{}, args ...interface{}) (interface{}, error) {
-			if len(args) != 0 {
-				return nil, fmt.Errorf("%s takes no arguments", nameFunc)
-			}
-			return event.Data.Name, nil
+		h.scope.SetDynamicFunc(nameFunc, &stateful.DynamicFunc{
+			F: func(args ...interface{}) (interface{}, error) {
+				if len(args) != 0 {
+					return nil, fmt.Errorf("%s takes no arguments", nameFunc)
+				}
+				return event.Data.Name, nil
+			},
+			Sig: nameFuncSignature,
 		})
 	}
 
 	if h.usesTaskName {
-		h.scope.SetDynamicMethod(taskNameFunc, func(self interface{}, args ...interface{}) (interface{}, error) {
-			if len(args) != 0 {
-				return nil, fmt.Errorf("%s takes no arguments", taskNameFunc)
-			}
-			return event.Data.TaskName, nil
+		h.scope.SetDynamicFunc(taskNameFunc, &stateful.DynamicFunc{
+			F: func(args ...interface{}) (interface{}, error) {
+				if len(args) != 0 {
+					return nil, fmt.Errorf("%s takes no arguments", taskNameFunc)
+				}
+				return event.Data.TaskName, nil
+			},
+			Sig: taskNameFuncSignature,
 		})
 	}
 
 	if h.usesDuration {
-		h.scope.SetDynamicMethod(durationFunc, func(self interface{}, args ...interface{}) (interface{}, error) {
-			if len(args) != 0 {
-				return nil, fmt.Errorf("%s takes no arguments", durationFunc)
-			}
-			return event.State.Duration, nil
+		h.scope.SetDynamicFunc(durationFunc, &stateful.DynamicFunc{
+			F: func(args ...interface{}) (interface{}, error) {
+				if len(args) != 0 {
+					return nil, fmt.Errorf("%s takes no arguments", durationFunc)
+				}
+				return event.State.Duration, nil
+			},
+			Sig: durationFuncSignature,
 		})
 	}
 
--- a/tick/stateful/eval_function_node.go
+++ b/tick/stateful/eval_function_node.go
@@ -32,15 +32,31 @@ func NewEvalFunctionNode(funcNode *ast.FunctionNode) (*EvalFunctionNode, error) 
 }
 
 func (n *EvalFunctionNode) Type(scope ReadOnlyScope, executionState ExecutionState) (ast.ValueType, error) {
-	// PERF: today we are evaluating the function, it will be much faster if will type info the function it self
-	result, err := n.callFunction(scope.(*Scope), executionState)
-	if err != nil {
-		return ast.InvalidType, err
+	f := executionState.Funcs[n.funcName]
+	if f == nil {
+		return ast.InvalidType, fmt.Errorf("undefined function: %q", n.funcName)
+	}
+	signature := f.Signature()
+
+	domain := Domain{}
+	if len(n.argsEvaluators) > len(domain) {
 	}
 
-	// We can't cache here the result (although it's very tempting ;))
-	// because can't trust function to return always the same consistent type
-	return ast.TypeOf(result), nil
+	for i, argEvaluator := range n.argsEvaluators {
+		t, err := argEvaluator.Type(scope, executionState)
+		if err != nil {
+			return ast.InvalidType, fmt.Errorf("Failed to handle %v argument: %v", i+1, err)
+		}
+		domain[i] = t
+	}
+
+	retType, ok := signature[domain]
+	if !ok {
+		// TODO: Cleanup error returned here
+		return ast.InvalidType, fmt.Errorf("Wrong function signature")
+	}
+
+	return retType, nil
 }
 
 func (n *EvalFunctionNode) IsDynamic() bool {
@@ -174,7 +190,7 @@ func (n *EvalFunctionNode) EvalBool(scope *Scope, executionState ExecutionState)
 // eval - generic evaluation until we have reflection/introspection capabillities so we can know the type of args
 // and return type, we can remove this entirely
 func eval(n NodeEvaluator, scope *Scope, executionState ExecutionState) (interface{}, error) {
-	retType, err := n.Type(scope, CreateExecutionState())
+	retType, err := n.Type(scope, executionState)
 	if err != nil {
 		return nil, err
 	}
--- a/tick/stateful/eval_function_node.go
+++ b/tick/stateful/eval_function_node.go
@@ -3,6 +3,7 @@ package stateful
 import (
 	"fmt"
 	"regexp"
+	"strings"
 	"time"
 
 	"github.com/influxdata/kapacitor/tick/ast"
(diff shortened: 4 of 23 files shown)


tick/stateful/eval_function_node.go
```diff
@@ -187,6 +216,12 @@ func eval(n NodeEvaluator, scope *Scope, executionState ExecutionState) (interfa
 		return n.EvalTime(scope, executionState)
 	case ast.TDuration:
 		return n.EvalDuration(scope, executionState)
+	case ast.TMissing:
```
nathanielc: Do we need this? I can't think of a reason that we would want the result of an entire expression to evaluate to the type TMissing. Or am I misunderstanding the use of TMissing?

tick/stateful/eval_int_node.go
```diff
@@ -30,15 +30,23 @@ func (n *EvalIntNode) EvalString(scope *Scope, executionState ExecutionState) (s
...
+
 func (n *EvalIntNode) EvalTime(scope *Scope, executionState ExecutionState) (time.Time, error) {
 	return time.Time{}, ErrTypeGuardFailed{RequestedType: ast.TTime, ActualType: ast.TInt}
 }
+
 func (n *EvalIntNode) EvalDuration(scope *Scope, executionState ExecutionState) (time.Duration, error) {
 	return 0, ErrTypeGuardFailed{RequestedType: ast.TDuration, ActualType: ast.TInt}
 }
+
+func (n *EvalIntNode) EvalMissing(scope *Scope, executionState ExecutionState) (*ast.Missing, error) {
```
nathanielc: With the above comment it doesn't seem like we need the `EvalMissing` method.

tick/stateful/expr.go
```diff
@@ -115,6 +120,12 @@ func (se *expression) Eval(scope *Scope) (interface{}, error) {
 			return nil, err
 		}
 		return result, err
+	case ast.TMissing:
```
nathanielc: Same here I don't thing we need this case.

tick/stateful/functions.go
```diff
@@ -12,14 +12,24 @@ import (
...
+	"github.com/influxdata/kapacitor/tick/ast"
+)
+
+const (
+	maxArgs = 4
 )
 
 var ErrNotFloat = errors.New("value is not a float")
 
+var funcSignatures = map[string]map[Domain]ast.ValueType{}
```
nathanielc: Is this left over from something?

tick/stateful/functions.go
```diff
@@ -541,6 +740,24 @@ func (boolean) Call(args ...interface{}) (v interface{}, err error) {
 	return
 }
 
+var booleanFuncSignature = map[Domain]ast.ValueType{}
+
+func initializeBooleanFuncSignature() {
```
nathanielc: Should we rename these functions to `init` that way it is clear when they are called, since  it is valid to have more than one `init` function.

tick/stateful/functions.go
```diff
@@ -573,6 +790,24 @@ func (integer) Call(args ...interface{}) (v interface{}, err error) {
...
+func initializeIntegerFuncSignature() {
+	d := Domain{}
+	d[0] = ast.TBool
+	integerFuncSignature[d] = ast.TInt
+	d[0] = ast.TString
+	integerFuncSignature[d] = ast.TInt
+	d[0] = ast.TInt
+	integerFuncSignature[d] = ast.TInt
+	d[0] = ast.TFloat
+	integerFuncSignature[d] = ast.TInt
```
nathanielc: Reusing the same domain in this function works right? I would expect so but wanted to confirm.

tick/stateful/functions.go
```diff
@@ -925,3 +1301,69 @@ func (ifFunc) Call(args ...interface{}) (interface{}, error) {
...
+	types := []ast.ValueType{
+		ast.TFloat,
+		ast.TInt,
+		ast.TString,
+		ast.TBool,
+		ast.TRegex,
+		ast.TTime,
+		ast.TDuration,
+		ast.TLambda,
+		ast.TList,
```
nathanielc: I don't think TLambda and TList make sense here.

tick/stateful/functions.go
```diff
@@ -925,3 +1301,69 @@ func (ifFunc) Call(args ...interface{}) (interface{}, error) {
...
+
+func (isPresent) Reset() {
+
+}
+
+func (isPresent) Call(args ...interface{}) (v interface{}, err error) {
+	if len(args) != 1 {
+		return false, errors.New("isMissing expects exactly one argument")
+	}
+	if _, isMissing := args[0].(*ast.Missing); !isMissing {
```
nathanielc: You can remove the if statement here and just return `!isMissing, nil`

tick/stateful/node_evaluator.go
```diff
@@ -34,7 +34,9 @@ type NodeEvaluator interface {
 	EvalRegex(scope *Scope, executionState ExecutionState) (*regexp.Regexp, error)
 	EvalTime(scope *Scope, executionState ExecutionState) (time.Time, error)
 	EvalDuration(scope *Scope, executionState ExecutionState) (time.Duration, error)
+	EvalMissing(scope *Scope, executionState ExecutionState) (*ast.Missing, error)
```
nathanielc: Again here, I don't think we need this? Are we calling it anywhere I did't see calls to it anywhere.

tick/stateful/functions.go
desa: yup.

tick/stateful/functions.go
desa: I didn't know that. That's awesome. I'll do that now.

tick/stateful/functions.go
desa: Yup :)

tick/stateful/eval_function_node.go
desa: There could be another way of doing it, this is just what seemed most clear to me.

Here's my reasoning (in a some what deductive order):

Typically, in tickscript if a field or tag doesn't exist, when the scope is field, an error is returned. So the first thing I did was change this so that instead, a missing value was set.
```
		if !isFieldExists && !isTagExists {
			if !vars.Has(refVariableName) {
				var missingValue interface{} = &ast.Missing{}
				vars.Set(refVariableName, missingValue)
				//return fmt.Errorf("no field or tag exists for %s", refVariableName)
			}
		}
```
https://github.com/influxdata/kapacitor/blob/c55df2732113e3e82ea72f4dd62367b4b3fd2499/expr.go#L54-L55

The function `isPresent` must be passed some value to check presence against (the lack of a value isn't something you can check). This value is a reference and must have a value distinct from all other types of values (so that `isPresent` can properly distinguish between present and non-present values). 

At some point, this reference must be evaluated to a value as it must be passed to the `isPresent` function.
```
	for i, argEvaluator := range n.argsEvaluators {
		value, err := eval(argEvaluator, scope, executionState)
		if err != nil {
			return nil, fmt.Errorf("Failed to handle %v argument: %v", i+1, err)
		}

		args = append(args, value)
	}
```
https://github.com/influxdata/kapacitor/blob/c55df2732113e3e82ea72f4dd62367b4b3fd2499/tick/stateful/eval_function_node.go#L71

Since we have to evaluate the reference to a missing value, we need an `EvalMissing` function on `NodeEvaluator` (because of the function signature of `eval`).
```
func eval(n NodeEvaluator, scope *Scope, executionState ExecutionState)
```
https://github.com/influxdata/kapacitor/blob/c55df2732113e3e82ea72f4dd62367b4b3fd2499/tick/stateful/eval_function_node.go#L198

nathanielc: In summary we need EvalMissing so that a ReferenceNode can correctly return it, makes sense. There is a hacky way around that but you have done the work needed so lets leave it as it.

desa: 👍

expr.go
```diff
@@ -50,7 +51,9 @@ func fillScope(vars *stateful.Scope, referenceVariables []string, now time.Time,
 		}
 		if !isFieldExists && !isTagExists {
 			if !vars.Has(refVariableName) {
-				return fmt.Errorf("no field or tag exists for %s", refVariableName)
+				var missingValue interface{} = &ast.Missing{}
```
nathanielc: Should the ast package export a VarMissing var so that we do not have to allocate a new one each time?

```
 vars.Set(refVariableName, ast.MissingValue)
```

tick/stateful/functions.go
desa: good catch.

expr.go
desa: Yup.

tick/stateful/node_evaluator.go
```diff
@@ -34,7 +34,9 @@ type NodeEvaluator interface {
...
 
+	// TODO: Remove executionState from Type functions type signature
```
nathanielc: I am curious to see how this works out. I am really hoping it doesn't break something unexpected.

desa: The reason for this TODO is that I naively tried and it broke a bunch of things. So I set it off temporarily.

desa:
@nathanielc One thing I'm not sure what to do about is [task_master.go#L393-L394](https://github.com/influxdata/kapacitor/blob/d0b41996febb833e65c1d4a41d1f68ad5461261d/task_master.go#L393-L394).

nathanielc:
Hmm, I think we need to separate out dynamic methods vs dynamic functions. Where a method is always a function bound to an object and a function is global. 

Right now the DynamicMethod bits are used for two purposes.

1. Allow UDFs to be dynamically part of TICKscript a la `@myUDF()...` the `myUDF` is a method call on the parent node.
2. Allow arbitrary global functions to be added to the scope, this is only used in the new alert matching logic for things like `changed()` and `level()`. 

Since those are really very different and the first case doesn't ever show up in a lambda expression I think they should be handled separately. I think the best way forward is to add DynamicFunctions and leave DynamicMethods alone. This code https://github.com/influxdata/kapacitor/blob/master/tick/stateful/eval_function_node.go#L65-L69 was added with the recent alert handler changes, which re-purposed DynamicMethods. In hind sight that wasn't a good change. Lets revert that change and have both DynamicFunctions and DynamicMethods. Then only Functions need declare their signature since methods are bound to an object and cannot be part of evaluating stateful expressions.

Does that help?

desa:
I think that makes sense. I'll get on that now.

desa:
@nathanielc So I'll still need a way to access the `DynamicFunction`. Should I add another field to the `Scope` struct for that? So scope will look like

```go
type Scope struct {
    variables map[string]interface{}

    dynamicMethods map[string]DynamicMethod

    dynamicFunctions map[string]DynamicFunction
}
```

nathanielc:
@desa Yes, that sounds good.

desa:
@nathanielc kk made suggested changes.

tick/stateful/scope.go
```diff
@@ -19,14 +31,16 @@ var empty = new(interface{})
 type Scope struct {
 	variables map[string]interface{}
 
-	dynamicMethods map[string]*DynamicMethod
+	dynamicMethods map[string]DynamicMethod
+	dynamicFuncs   map[string]*DynamicFunc
```
desa: So that function signature of 
```
func (s *Scope) DynamicFunc(name string) *DynamicFunc {
	return s.dynamicFuncs[name]
}
```
doesn't have to be
```
func (s *Scope) DynamicFunc(name string) (*DynamicFunc, bool)
```
https://github.com/influxdata/kapacitor/blob/md-type-system/tick/stateful/scope.go#L90

So we can do a nil check in [Type] and [callFunction](https://github.com/influxdata/kapacitor/blob/md-type-system/tick/stateful/eval_function_node.go#L87).

Happy to change it to return the `bool` if you think thats better.

nathanielc: Ah, makes sense, either is fine with me.

tick/stateful/functions.go
desa: still need to do this.

tick/stateful/scope.go
```diff
@@ -7,9 +7,21 @@ import (
 	"github.com/influxdata/kapacitor/tick/ast"
 )
 
-type DynamicMethod struct {
-	F         func(self interface{}, args ...interface{}) (interface{}, error)
-	Signature map[Domain]ast.ValueType
+type DynamicMethod func(self interface{}, args ...interface{}) (interface{}, error)
+
+type DynamicFunc struct {
+	F   func(self interface{}, args ...interface{}) (interface{}, error)
```
desa: fixed

services/alert/handlers.go
```diff
@@ -506,54 +506,84 @@ func (h *matchHandler) Handle(event alert.Event) {
...
 
 	if h.usesChanged {
-		h.scope.SetDynamicMethod(changedFunc, func(self interface{}, args ...interface{}) (interface{}, error) {
-			if len(args) != 0 {
-				return nil, fmt.Errorf("%s takes no arguments", changedFunc)
-			}
-			return event.State.Level != event.PreviousState().Level, nil
+		h.scope.SetDynamicFunc(changedFunc, &stateful.DynamicFunc{
+			F: func(self interface{}, args ...interface{}) (interface{}, error) {
+				if len(args) != 0 {
```
desa: While it would be a violation of the signature, there's nothing that prevents a user from deciding to call `F` themselves without first checking the signature. Especially since `F` is exported. Thoughts?

tick/stateful/eval_function_node.go
```diff
@@ -32,15 +33,35 @@ func NewEvalFunctionNode(funcNode *ast.FunctionNode) (*EvalFunctionNode, error)
...
+		if df == nil {
+			return ast.InvalidType, fmt.Errorf("undefined function: %q", n.funcName)
+		}
+		f = df
+	}
+	signature := f.Signature()
+
+	domain := Domain{}
+	if len(n.argsEvaluators) > len(domain) {
+	}
```
desa: TBH, no idea. I think I meant for there to be an error if someone supplied more arguments than any of our functions took, but it looks like I never got around to filling that in. I'll do that now.

tick/stateful/eval_function_node.go
```diff
@@ -171,10 +192,22 @@ func (n *EvalFunctionNode) EvalBool(scope *Scope, executionState ExecutionState)
...
+		return missingValue, nil
+	}
+	return nil, ErrTypeGuardFailed{RequestedType: ast.TMissing, ActualType: ast.TypeOf(refValue)}
+}
+
 // eval - generic evaluation until we have reflection/introspection capabillities so we can know the type of args
 // and return type, we can remove this entirely
 func eval(n NodeEvaluator, scope *Scope, executionState ExecutionState) (interface{}, error) {
-	retType, err := n.Type(scope, CreateExecutionState())
+	retType, err := n.Type(scope, executionState)
```
desa: Nothing should actually use Execution state, but that might not be true. I haven't tried to remove it again. I'll switch back to using  `CreateExecutionState()` until we rip out execution state from `Type`.

desa:
@nathanielc pending a review of the most recent couple commits and final squash I think things should be good to go.

tick/stateful/eval_function_node.go
```diff
@@ -31,16 +32,36 @@ func NewEvalFunctionNode(funcNode *ast.FunctionNode) (*EvalFunctionNode, error)
...
+		f = df
 	}
+	signature := f.Signature()
 
-	// We can't cache here the result (although it's very tempting ;))
-	// because can't trust function to return always the same consistent type
-	return ast.TypeOf(result), nil
+	domain := Domain{}
+	if gotLen, expLen := len(n.argsEvaluators), len(domain); gotLen > expLen {
+		return ast.InvalidType, fmt.Errorf("functions do not accept more than %v arguements: received %v", expLen, gotLen)
```
nathanielc: Would it be better to return an ErrWrongFuncSignature here instead so the user has an idea of how to fix the problem?

tick/stateful/eval_function_node.go
```diff
@@ -63,8 +84,8 @@ func (n *EvalFunctionNode) callFunction(scope *Scope, executionState ExecutionSt
 
 	// Look for function on scope
 	if f == nil {
-		if dm := scope.DynamicMethod(n.funcName); dm != nil {
-			f = dynamicMethodFunc{dm: dm}
+		if df := scope.DynamicFunc(n.funcName); df != nil {
+			f = df
 		}
 	}
```
nathanielc: Since this logic is repeated above should we make a small helper method for it? Something like `lookupFunction(state, scope)`?

Not super important, just thought of it while looking through.

tick/stateful/functions.go
```diff
@@ -12,21 +12,76 @@ import (
 
 	"github.com/dustin/go-humanize"
 	"github.com/influxdata/influxdb/influxql"
+	"github.com/influxdata/kapacitor/tick/ast"
 )
 
+const (
+	maxArgs = 4
```
nathanielc: Can you add a comment here that this about the max args for any function? That way it is clear it needs to be increased if a new function is added.

tick/stateful/functions.go
```diff
@@ -12,21 +12,76 @@ import (
...
+
+	return ds
 }
 
 // Lookup for functions
 type Funcs map[string]Func
 
 var statelessFuncs Funcs
 
+var GlobalExecutionSate ExecutionState
```
nathanielc: As far as I can tell this is just used to lookup the built in functions as opposed to the dynamic functions. And only the map of functions is used. Maybe this would be better as

```
var builtinFunc Func
```

Also I think it should be private since it is not used outside the package.

tick/stateful/eval_reference_node.go
```diff
@@ -128,3 +128,17 @@ func (n *EvalReferenceNode) EvalBool(scope *Scope, executionState ExecutionState
...
+
+func (n *EvalReferenceNode) EvalMissing(scope *Scope, executionState ExecutionState) (*ast.Missing, error) {
+	refValue, err := n.getReferenceValue(scope)
+	if err != nil {
+		return nil, err
+	}
+
+	if missingVal, isMissing := refValue.(*ast.Missing); isMissing {
+		//  This error gets checked in the eval method of a function node
+		return missingVal, fmt.Errorf("missing value: \"%v\"", n.Node.Reference)
```
nathanielc: > This error gets checked in the eval method of a function node

I couldn't find where this was happening?

nathanielc:
I just tested this locally and got this error:

```
E! error while evaluating expression: invalid comparison operator > for type missing
```

with this TICKscript 

```
stream
    |from()
        .measurement('missing')
    |where(lambda: "m1" > 6)
    |where(lambda: isPresent("m2"))
    |eval(lambda: "m2" + "m1")
        .as('m3')
    |log()
```

and sending this point:

```
missing m2=8
```

Can we make the error about the fact that the field `m1` is missing not about the operator type error?

tick/stateful/eval_function_node.go
desa: Yup. Good idea.

tick/stateful/eval_reference_node.go
desa: Here https://github.com/influxdata/kapacitor/blob/1545c7a278653ed8e7f4642a7fbc994e9c51e8c1/tick/stateful/eval_function_node.go#L230

tick/stateful/functions.go
desa: Yeah. Good point.

desa:
> E! error while evaluating expression: invalid comparison operator > for type missing

Should be an easy fix

desa:
@nathanielc made changes you suggested.

nathanielc:
@desa Now I am getting the error:

```
E! error while evaluating expression: left value is missing value
```

Can it include the name of the missing field/tag?

desa:
ah yes. I'll add that. Sorry for the oversight.

desa:
@nathanielc  should be good to go error message says

```
[miss:where2] 2017/04/20 15:02:56 E! error while evaluating expression: left value "m1" is missing value
```

now.

nathanielc:
@desa There are still a few cases where errors can occur when the type is missing that doesn't include what the name of the field was that is missing. 

For example:

```
stream
    |from()
        .measurement('m')
    |where(lambda: "t")
    |log()
```

Gives the error:

```
E! error while evaluating expression: TypeGuard: expression returned unexpected type missing, expected boolean
```


Since TypeGuard is an error type maybe we can simply have it check for the Missing Type and return a better error message?

desa:
@nathanielc I'll fix type guard to add missing variable name. I'll go through and do a scan for anywhere that might have that issue

desa:
@nathanielc I avoided the type guard error and opted to just check for missing types in the reference nodes. Figured it was better than changing type guard.

I did another comb through. I think there might be some way to get an unexpected error from an unary node, but I'm having a hard time reaching that case. I can write some additional check in just in case. Thoughts?

desa:
in particular I'm able to produce

```
[another:eval2] 2017/05/02 14:22:12 E! TypeGuard: expression returned unexpected type invalid type, expected missing
```

with this tickscript

```
stream
    |from()
        .measurement('m')
    |eval(lambda: -"t")
        .as('isk')
    |log()
```

desa:
fixed. should be good to go assuming I haven't missed anything

CHANGELOG.md
```diff
@@ -81,6 +81,8 @@ kapacitor define-handler system aggregate_by_1m.yaml
 - [#507](https://github.com/influxdata/kapacitor/issues/507): Add API endpoint for performing Kapacitor database backups.
 - [#1132](https://github.com/influxdata/kapacitor/issues/1132): Adding source for sensu alert as parameter
 - [#1346](https://github.com/influxdata/kapacitor/pull/1346): Add discovery and scraping services.
+- [#1284](https://github.com/influxdata/kapacitor/pull/1284): Add type signatures to Kapacitor functions.
+- [#1203](https://github.com/influxdata/kapacitor/issues/1203): Add `isPresent` operator for verifying whether a value is present (part of [#1284](https://github.com/influxdata/kapacitor/pull/1284)).
```
nathanielc: Can you update the CHANGELOG here for the new unreleased section?

desa:
@nathanielc updated changelog

PR #1285, opened 2017-03-30
Example TICKScript improvements

Crapworks:
Hey Folks,

I've discovered some minor issues with your examples and wanted to share my experiences:

1. The cpu example is only alerting when `usage_user` exceeds the thresholds. Problems like high IOWait or System usage are not alerted. It should check the threshold against everything that is **not idle**.

2. The disk example checks the threshold agains the `mean` of **all disks available**. This means if one disk is full but another one almost empty, there is no alert. It should group by path and then alert with the path as well.

Cheers,
Christian


nathanielc:
@Crapworks Thanks!

PR #1286, opened 2017-03-30
Start with a blank URL; Chronograf provides a placeholder

lukevmorris:
- [x] Rebased/mergable
- [x] Tests pass
- [x] CHANGELOG.md updated

### DIFF
--- a/etc/kapacitor/kapacitor.conf
+++ b/etc/kapacitor/kapacitor.conf
@@ -321,7 +321,7 @@ default-retention-policy = ""
   enabled = false
   # The HipChat API URL. Replace subdomain with your
   # HipChat subdomain.
-  url = "https://subdomain.hipchat.com/v2/room"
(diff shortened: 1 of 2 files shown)


etc/kapacitor/kapacitor.conf
```diff
@@ -321,7 +321,7 @@ default-retention-policy = ""
   enabled = false
   # The HipChat API URL. Replace subdomain with your
   # HipChat subdomain.
-  url = "https://subdomain.hipchat.com/v2/room"
+  url = ""
```
nathanielc: Yeah, that should not have been there as it is not a valid configuration value. Instead of removing it can we leave it commented out so people can see the structure the URL they need to supply?

lukevmorris: Sure thing

lukevmorris:
@nathanielc Updated!

nathanielc:
Thanks!

PR #1296, opened 2017-04-03
Add repair logic for broken indexes

### LINKED ISSUE #507: Add backup endpoint to API for storage data.
BoltDB makes performing backups of the entire db easy. We need to expose that ability through the API.

### LINKED ISSUE #1295: Deleting all recordings corrupts list of recordings
I think this was triggered by issuing `kapacitor delete recordings *` because initially that didn't work either and the issue appeared immediately afterwards.

Now when I issue `kapacitor list recordings` I get:

`failed to list recordings with pattern "": no key exists`

I can see the following appear in `kapacitor.log`:

`[httpd] 127.0.0.1 - - [03/Apr/2017:08:44:48 +1000] "GET /kapacitor/v1/recordings?fields=type&fields=size&fields=date&fields=status&limit=100&offset=0&pattern= HTTP/1.1" 400 92 "-" "KapacitorClient" fa11ae1f-17f5-11e7-8332-000000000000 16340`

Strangely enough, I still seem to be able to record and replay recordings just fine.

nathanielc:
Fixes #1295 

Also Fixes #507 since using any of these operations is potentially destructive we needed a backup endpoint.

- [x] Rebased/mergable
- [x] Tests pass
- [x] CHANGELOG.md updated
- [x]  Migrate the repair process to be an API call instead of a startup action.

### DIFF
--- a/client/API.md
+++ b/client/API.md
@@ -8,6 +8,7 @@
 * [Replays](#replays)
 * [Alerts](#alerts)
 * [Configuration](#configuration)
+* [Storage](#storage)
 * [Testing Services](#testing-services)
 * [Miscellaneous](#miscellaneous)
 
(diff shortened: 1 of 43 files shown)


services/replay/service.go
```diff
@@ -120,6 +120,12 @@ func (s *Service) Open() error {
 		return err
 	}
 
+	//TODO: Should we expose this via the API instead of doing it during every startup?
+	// This could be expensive and should not need to be run frequently.
+	if err := s.repairRecordingIndex(); err != nil {
```
nathanielc: @desa What do you think we should do here?

desa: I'd say have it be a part of the API and not to run it on startup.

desa: Sorry that I missed this earlier.

nathanielc:
@desa Could you give this another look now that I have exposed the rebuild process through the API?

nathanielc:
@desa Actually if you want to hold off, I just realized I need to make a change.

nathanielc:
@desa OK this is ready for review. Summary of changes since your first review:

* Added API endpoint to create database Backups
* Added API endpoints to perform actions on various underlying stores.

client/API.md
```diff
@@ -2116,6 +2117,62 @@ POST /kapacitor/v1/config/influxdb/remote
...
+Always make a backup of the database before performing any of these operations.
+
+### Backing up the Storage
+
+Making a GET request to `/kapacitor/v1/storage/backup` will return a dump of the Kapacitor database.
+To restore from a backup replace the `kapacitor.db` file with the contents of the backup request.
+
+```
+# Create a backup.
+curl http://localhost:9092/kapacitor/v1/storage/backup > kapacitor.db
```
desa: maybe we should have a `kapacitor backup` sub-command?

desa:
Just Finished reading. From what I can tell LGTM.

client/API.md
nathanielc: Agreed, I'll put this together. I don't want a subcommand for the rebuild action though, I figure those are best only in the API so that you have to read and be aware of how they work.

desa: Should we update the example to use the backup subcommand?

nathanielc: I think the example here using the API is fine because it is part of the API docs.

desa: kk

PR #1299, opened 2017-04-04
allowing sensu handler to be specified

davidloutsch:
- [x] Rebased/mergable
- [x] Tests pass
- [x] CHANGELOG.md updated

### DIFF
--- a/pipeline/alert.go
+++ b/pipeline/alert.go
@@ -949,6 +949,7 @@ func (a *AlertaHandler) Services(service ...string) *AlertaHandler {
 //      enabled = true
 //      url = "http://sensu:3030"
 //      source = "Kapacitor"
+//      handlers = ["sns","slack"]
 //
 // Example:
 //    stream
@@ -956,6 +957,14 @@ func (a *AlertaHandler) Services(service ...string) *AlertaHandler {
 //             .sensu()
 //
 // Send alerts to Sensu client.
+//
+// Example:
+//    stream
+//         |alert()
+//             .sensu()
+//             .handlers('sns','slack')
+//
+// Send alerts to Sensu specifying the handlers
 //
 // tick:property
 func (a *AlertNode) Sensu() *SensuHandler {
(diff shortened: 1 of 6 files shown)


pipeline/alert.go
```diff
@@ -949,6 +949,7 @@ func (a *AlertaHandler) Services(service ...string) *AlertaHandler {
 //      enabled = true
 //      url = "http://sensu:3030"
 //      source = "Kapacitor"
+//      handlers = "sns"
```
nathanielc: How would you list multiple handlers?

nathanielc: Would you want different handlers per alert or global to all alerts?

davidloutsch: Good questions. I checked in some code that would allow for a list of handlers. I only needed one, but lists work as well.
I could understand a need for having different handlers per alert, but we only need a single global one at the moment.

nathanielc:
Can you fix this failing test?
```
--- FAIL: TestServer_UpdateConfig (0.37s)
	server_test.go:6510: unexpected defaults for sensu/: sensu: section element 0 are not equal: extra option "Handlers" with value <nil>
	server_test.go:6525: unexpected update result 0 for sensu/: sensu: section element 0 are not equal: extra option "Handlers" with value <nil>
```

davidloutsch:
Sure thing. Let me know if you need anything else.

nathanielc:
> I could understand a need for having different handlers per alert, but we only need a single global one at the moment.

I think it would be more complete to add the list of handlers to the each instance of a sensu handler. That way you can specify them globally and per alert. It shouldn't be too much more work and here is an open PR doing something very similar  https://github.com/influxdata/kapacitor/pull/1314 . 

Do you think you could take a crack at adding Handlers on a per alert basis?

services/sensu/config.go
```diff
@@ -11,6 +11,8 @@ type Config struct {
 	Addr string `toml:"addr" override:"addr"`
 	// The JIT sensu source name of the alert.
 	Source string `toml:"source" override:"source"`
+	// The sensu handler to use
+	Handlers []string `toml:"handlers"`
```
nathanielc: Adding the `override` tag here will make it so that the tests above use the lower case name to be consistent.

henry-megarry-tr:
Hey Nathaniel, I have added the changes to allow handlers on a per alert basis. Let me know if you need anything else.

services/sensu/service.go
```diff
@@ -55,18 +55,20 @@ func (s *Service) Update(newConfig []interface{}) error {
...
 func (s *Service) TestOptions() interface{} {
 	return &testOptions{
-		Name:   "testName",
-		Source: "Kapacitor",
-		Output: "testOutput",
-		Level:  alert.Critical,
+		Name:     "testName",
+		Source:   "Kapacitor",
+		Output:   "testOutput",
+		Handlers: []string{"testHandlers"},
```
nathanielc: By default I think it would be best to have an empty list of Handlers. Otherwise a default call to test the sensu service will probably fail.

PR #1314, opened 2017-04-11
adding source for sensu alert as parameter

henry-megarry-tr:
- [x] Rebased/mergable
- [x] Tests pass
- [x] CHANGELOG.md updated

### DIFF
--- a/server/server_test.go
+++ b/server/server_test.go
@@ -6000,7 +6000,7 @@ func TestServer_UpdateConfig(t *testing.T) {
 						Set: map[string]interface{}{
 							"addr":    "sensu.local:3000",
 							"enabled": true,
-							"source":  "",
+							"source":  "Kapacitor",
 						},
 					},
 					expSection: client.ConfigSection{
@@ -6010,7 +6010,7 @@ func TestServer_UpdateConfig(t *testing.T) {
 							Options: map[string]interface{}{
 								"addr":    "sensu.local:3000",
 								"enabled": true,
-								"source":  "",
+								"source":  "Kapacitor",
 							},
 							Redacted: nil,
 						}},
@@ -6020,7 +6020,7 @@ func TestServer_UpdateConfig(t *testing.T) {
 						Options: map[string]interface{}{
 							"addr":    "sensu.local:3000",
 							"enabled": true,
-							"source":  "",
+							"source":  "Kapacitor",
 						},
 						Redacted: nil,
 					},
@@ -6616,6 +6616,7 @@ func TestServer_ListServiceTests(t *testing.T) {
 				Options: client.ServiceTestOptions{
 					"name":   "testName",
 					"output": "testOutput",
+					"source": "Kapacitor",
 					"level":  "CRITICAL",
 				},
 			},
@@ -6717,6 +6718,7 @@ func TestServer_ListServiceTests_WithPattern(t *testing.T) {
 				Options: client.ServiceTestOptions{
 					"name":   "testName",
 					"output": "testOutput",
+					"source": "Kapacitor",
 					"level":  "CRITICAL",
 				},
 			},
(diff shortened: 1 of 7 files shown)


server/server_test.go
```diff
@@ -7751,6 +7756,7 @@ stream
 		.message('message')
 		.details('details')
 		.crit(lambda: TRUE)
+		.sensu()
```
nathanielc: What is this change doing? The alert should be sent to sensu via the topic handler, not explicitly in the TICKscript.

services/sensu/service.go
```diff
@@ -145,21 +153,47 @@ func (s *Service) prepareData(name, output string, level alert.Level) (*net.TCPA
...
 func (h *handler) Handle(event alert.Event) {
+	td := event.TemplateData()
+	var buf bytes.Buffer
+	err := h.sourceTmpl.Execute(&buf, td)
+	if err != nil {
+		h.logger.Printf("E! failed to evaluate Sensu source template %s: %v", h.c.Source, err)
+		return
+	}
+	sourceStr := buf.String()
+	buf.Reset()
```
nathanielc: No need to reset the buffer here as it isn't being used later.

server/server_test.go
henry-megarry-tr: The sensu TestServer_AlertHandlers test was failing without that being set. 
`        server_test.go:7784: sensu: unexpected sensu request:
                exp
                [{Name:id Source:Kapacitor Output:message Status:2}]
                got
                []
`
 Is there something I can do to make it work without that?

nathanielc: This bit of code here should be registering a sensu handler for the topic `test`. So it should get an alert event, this hasn't changed in your code so I am not sure why you would be getting that error. But it should definitely be passing without the `.sensu()` bit as it was before.

henry-megarry-tr: It looks like I am getting this error in services/sensu/sensutest/sensutest.go line 53

"accept tcp 127.0.0.1:48578: use of closed network connection"

Not sure what I changed that would cause this to happen. Any suggestions on how to fix the issue?

nathanielc: Hmm, nothing is jumping out at me either. Maybe later I'll pull down the branch and take a look.

henry-megarry-tr: some more digging found that the Sensu service Handler() method is being called however the Handle() method is not.  It is called if the .sensu() is in the tick script. How does the topic handler know to use the sensu service?

services/alert/service.go
```diff
@@ -769,7 +770,15 @@ func (s *Service) createHandlerFromSpec(spec HandlerSpec) (handler, error) {
 		}
 		h = NewPublishHandler(c, s.logger)
 	case "sensu":
-		h = s.SensuService.Handler(s.logger)
+		c := sensu.HandlerConfig{}
+		err = decodeOptions(spec.Options, &c)
+		if err != nil {
+			return handler{}, err
+		}
+		h, err := s.SensuService.Handler(c, s.logger)
```
nathanielc: This is where a topic handler calls into the SensuService for the sensu handler. And it looks correct.

nathanielc: I found the bug it is super subtle, the above line should read

```
h, err = s.SensuService.Handler(c, s.logger)
```

this is because we want to assign to the var h Handler from the function scope not our local scope. I should redesign this so that isn't so easy to mess up, but for now it will fix your test.

henry-megarry-tr: sweet thanks for the help! will get this in right away!

server/server_test.go
```diff
@@ -7391,6 +7393,9 @@ func TestServer_AlertHandlers(t *testing.T) {
 		{
 			handler: client.TopicHandler{
 				Kind: "sensu",
+				Options: map[string]interface{}{
+					"Source": "Kapacitor",
```
nathanielc: The key should be lower case to follow convention.

nathanielc:
@henry-megarry-tr The changes look good. :)

Just a few house keeping items to finish up

*  Add an entry to the CHANGELOG, in the feature section of the `unreleased` section.
*  Rebase and squash your commits into a single commit.
*  Sign the CLA so I can merge the code.

Thanks for this great PR! :tada:

nathanielc:
@henry-megarry-tr Thanks, merged.

Closed #1132

PR #1318, opened 2017-04-12
Fix bug where aggregate operations could not change type

### LINKED ISSUE #1271: InfluxQL default type should be settable by user
The particular problem I'm having is that when count() is chained after where(), which is filtering a window()...

    stream
    | window().every(1s).period(60s)
    | where(...)
    | count('timestamp')

If the stream starts with data that does not fit the where criteria, count seems to initialize to an InfluxQL node for type float64.  If the stream starts with data that does fit the where criteria, counts seems to initialize to an InfluxQL node for type int64.

The problem is if the count node starts as a float64 type, then when data meeting the where criteria start arriving, the timestamp type is int64, therefore kapacitor throws the following error

    2017/03/23 17:14:13 E! failed to aggregate batch: field timestamp has wrong type: got int64 exp float64

I believe there is no way to coerce the data type for an InfluxQL node, and coercing timestamp to float64 would be the wrong thing to do.

This seems to be corroborated by https://github.com/influxdata/kapacitor/blob/master/influxql.go#L180 which seems to be setting the type of reducer used by batch InfluxQL nodes to float64 by default when there are no points in the batch.  Changing this line to int64(0) allows the above scenario to work without issue.

It seems like there should be a user settable property that specifies the default type rather than having a hard coded float64 default.

nathanielc:
Fixes #1271 

- [x] Rebased/mergable
- [x] Tests pass
- [x] CHANGELOG.md updated

### DIFF
--- a/integrations/streamer_test.go
+++ b/integrations/streamer_test.go
@@ -701,6 +701,7 @@ stream
 
 	testStreamerWithOutput(t, "TestStream_Window_Count", script, 2*time.Second, er, false, nil)
 }
+
 func TestStream_Window_Count_Overlapping(t *testing.T) {
 
 	var script = `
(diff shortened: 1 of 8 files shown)


integrations/streamer_test.go
```diff
@@ -1323,6 +1324,43 @@ stream
...
+	er := models.Result{
+		Series: models.Rows{
+			{
+				Name:    "m",
+				Tags:    nil,
+				Columns: []string{"time", "count"},
+				Values: [][]interface{}{
+					{
+						time.Date(1971, 1, 1, 0, 0, 20, 0, time.UTC),
+						1.0,
```
desa: What would have happened previously? It would have been a int?

desa: Or does it have to do with the type of `value` changing.

desa: Or is it something else entirely. I think I might be confused about what this fixes.

desa: This being this PR.

nathanielc: Previously it would have errored and returned no value.

nathanielc: To be more clear, the result here will always be an int since count returns an int. That isn't the type that is changing. Rather the type of the aggregate function used to perform the count (yes it is typed) wasn't changing. So in this test the first window is empty and so we choose a data type of float, but on the second window when there are actual values that are ints, we needed to change the aggregator function to be the int type function. Previous to this fix it would have logged an error about the types and returned not data for the count operation.

desa: got it

desa: that makes sense

PR #1322, opened 2017-04-13
TLS configuration for Slack service for Mattermost compatibility

seuf:
Hi,
Mattermost is compatible with slack messages json format so I've patched the slack services to add TLS configuration parameters.
Now alerts can be sent to self hosted Mattermost with custom certificates.

- [X] Rebased/mergable
- [x] Tests pass
- [x] CHANGELOG.md updated


server/server.go
```diff
@@ -475,7 +475,10 @@ func (s *Server) appendSensuService() {
 func (s *Server) appendSlackService() {
 	c := s.config.Slack
 	l := s.LogService.NewLogger("[slack] ", log.LstdFlags)
-	srv := slack.NewService(c, l)
+	srv, err := slack.NewService(c, l)
+	if err != nil {
+		s.Logger.Println("W! Slack init :", err)
```
nathanielc: Go ahead and return the error here and check it in NewServer and return it from there. That will cause the server to print the error message and shutdown, which is the behavior we want.

services/slack/service.go
```diff
@@ -17,14 +19,21 @@ import (
...
 	logger      *log.Logger
+	client      *http.Client
 }
 
-func NewService(c Config, l *log.Logger) *Service {
+func NewService(c Config, l *log.Logger) (*Service, error) {
+	tlsConfig, err := getTLSConfig(c.SSLCA, c.SSLCert, c.SSLKey, c.InsecureSkipVerify)
+	if err != nil {
+		return nil, err
+	}
```
nathanielc: Can you add a log warn here if the config specifies using InsecureSkipVerify?

services/slack/service.go
```diff
@@ -17,14 +19,21 @@ import (
...
 	s := &Service{
 		logger: l,
+		client: &http.Client{Transport: &http.Transport{TLSClientConfig: tlsConfig}},
```
nathanielc: The configuration can be changed at runtime via the API, there should be an Update method on the service which is in charge of applying new config changes. Can you update that method to allow for creating a new client when the TLS config changes. See the Alerta service for an example.

seuf:
I've updated the pr to add all your requested changes. Is it good to you ?

services/slack/service.go
```diff
@@ -99,7 +119,8 @@ func (s *Service) Alert(channel, message, username, iconEmoji string, level aler
 	if err != nil {
 		return err
 	}
-	resp, err := http.Post(url, "application/json", post)
+	s.logger.Println("D! Posting event to Slack", post)
+	resp, err := s.client.Post(url, "application/json", post)
```
nathanielc: This line races with the lines in Update and New that assign to the client. Either add a mutex to the Sensu service or use the sync.Atomic type to synchronizes access to the client.

Other than that this looks good! Thanks

seuf: Hum. I'm not a Go dev. This looks like a bit tricky for me.. O_ô 
I will try to do this next tuesday when i go back to work.

nathanielc: @seuf No worries, I can help you out too. The Alerta service does something similar if that helps.

seuf: OK, I have updated the PR to use atomic value for the http client like Alerta service.

services/slack/service.go
```diff
@@ -99,7 +127,9 @@ func (s *Service) Alert(channel, message, username, iconEmoji string, level aler
 	if err != nil {
 		return err
 	}
-	resp, err := http.Post(url, "application/json", post)
+	s.logger.Println("D! Posting event to Slack", post)
```
nathanielc: Can you remove this debug statement?

seuf:
done

nathanielc:
@seuf Thanks! :tada:

PR #1333, opened 2017-04-21
fix derivative to add result to current point, not previous

### LINKED ISSUE #1329: derivative node returns value from the future
The `derivative()` node computes the derivative between a point, and the one that follows it, but adds the calculated derivative value to the first point, not the second.
This is counterintuitive as this is not how a derivative is defined, nor is it how InfluxDB's derivative behaves.
Additionally the behavior results in the `derivative()` node buffering a point. If it behaved as expected, buffering would not be necessary.

```
foo v=1i 1000000000
foo v=2i 2000000000
foo v=4i 3000000000
foo v=8i 4000000000
foo v=16i 5000000000
```
```
stream|from().measurement('foo')|derivative('v').as('derivative')|log()
```

Expected:
```
"Fields":{"v":1},"Time":"1970-01-01T00:00:01Z"
"Fields":{"v":2,"derivative":1},"Time":"1970-01-01T00:00:02Z"
"Fields":{"v":4,"derivative":2},"Time":"1970-01-01T00:00:03Z"
"Fields":{"v":8,"derivative":4},"Time":"1970-01-01T00:00:04Z"
"Fields":{"v":16,"derivative":8},"Time":"1970-01-01T00:00:05Z"
```

Actual:
```
"Fields":{"v":1,"derivative":1},"Time":"1970-01-01T00:00:01Z"
"Fields":{"v":2,"derivative":2},"Time":"1970-01-01T00:00:02Z"
"Fields":{"v":4,"derivative":4},"Time":"1970-01-01T00:00:03Z"
"Fields":{"v":8,"derivative":8},"Time":"1970-01-01T00:00:04Z"
```

phemmer:
Fixes #1329 

- [X] Rebased/mergable
- [x] Tests pass
- [ ] CHANGELOG.md updated


integrations/data/TestStream_Derivative.srpl
```diff
@@ -6,7 +6,7 @@ rpname
 packets value=1001 0000000002
 dbname
 rpname
-packets value=1002 0000000003
+packets value2=0 0000000003
```
phemmer: This, and the other brpl changes was to test for the issue in [this comment](https://github.com/influxdata/kapacitor/issues/1329#issuecomment-295558255)

nathanielc:
@desa I would love your feedback on the changes here. I have looked through the code and it looks good. My biggest concern is the breaking behavior changes. See the issue #1329 for more details.

desa:
@nathanielc I agree with this change. While it is breaking, it is the behavior that I think people expect.

Issue #1151, opened 2017-01-24
Is there a MQTT event handler or a way to pass Kapacitor alert events to MQTT? 


rapport:
Hi guys, I am trying like to publish my alerts messages onto an MQTT broker. Is there anyway to do this directly from kapacitor currently?

I can't find anything obvious in the alert handler list and am starting look through to see if there is a way to adapt one of the handlers that are available. I had hoped for something like:
MQTT – Publish alert message on MQTT broker.

nathanielc:
@rapport Not currently, but is something we would definitely like to add. Are you conformable with Go? Is this something you think you could contribute? I can help walk through all the bits and pieces and there are lots of good examples of how to do this. See #1107 as an example

rapport:
Thanks for providing this response. I am OK with GO. Is there documentation regards how to create an alert handler? Would it be possible to adapt the influxdb MQTT output or the Telegraf MQTT input or would it be necessary to write from scratch?

nathanielc:
@rapport Here are some docs https://github.com/influxdata/kapacitor/blob/master/alert/HANDLERS.md. An MQTT package could easily be used, and is recommended. There is some boiler plate code that needs to be added to link it all together, but that is mostly copy paste from the linked doc.

FlavioF:
Hi
@rapport, @nathanielc any progress on this?

rapport:
Hi, I am sorry but no progress from me. I got most of what I wanted working using various workarounds and haven't had a chance to look at developing a new handler. I used the MQTT Warn (https://github.com/jpmens/mqttwarn) and the Paho library running in JavaScript and can also take a post from Kacapitor. However, I think MQTT should be a first rate citizen in kacapitor and was very happy to find MQTT input and output in the database and telegraf!

Koulwa:
Any progress on this? MQTT would be mission critical for using Kapacitor

PR #1345, opened 2017-04-27
Add an MQTT Alert Handler

timraymond:
- [x] Rebased/mergable
- [x] Tests pass
- [ ] CHANGELOG.md updated

Connect #1151 

### DIFF
--- a/Gopkg.toml
+++ b/Gopkg.toml
@@ -33,6 +33,10 @@ required = ["github.com/benbjohnson/tmpl","github.com/golang/protobuf/protoc-gen
   branch = "master"
   name = "github.com/shurcooL/markdownfmt"
 
+[[constraint]]
+  name = "github.com/eclipse/paho.mqtt.golang"
+  version = "~1.0.0"
+
 # Force the Azure projects to be a specific older version that Prometheus needs
 [[override]]
   name = "github.com/Azure/azure-sdk-for-go"
--- a/pipeline/alert.go
+++ b/pipeline/alert.go
@@ -1017,17 +1017,29 @@ func (a *AlertNode) Mqtt(topic string) *MQTTHandler {
 	return m
 }
 
+// tick:embedded:AlertNode.Mqtt
 type MQTTHandler struct {
 	*AlertNode
+
+	// BrokerName is the name of the configured MQTT broker to use when publishing the alert.
+	// If empty defaults to the configured default broker.
+	BrokerName string
 
 	// The topic where alerts will be dispatched to
 	Topic string
 
-	// The QoS that will be used to deliver the alerts
-	QoS int64
+	// The Qos that will be used to deliver the alerts
+	//
+	// Valid values are:
+	//
+	//    * 0 - At most once delivery
+	//    * 1 - At least once delivery
+	//    * 2 - Exactly once delivery
+	//
+	Qos int64
 
 	// Retained indicates whether this alert should be delivered to
-	// clients that were not connected to the broker at the time of the alert
+	// clients that were not connected to the broker at the time of the alert.
 	Retained bool
 }
 
--- a/pipeline/alert.go
+++ b/pipeline/alert.go
@@ -49,6 +49,7 @@ const defaultDetailsTmpl = "{{ json . }}"
 //    * Pushover -- Send alert to Pushover.
 //    * Talk -- Post alert message to Talk client.
 //    * Telegram -- Post alert message to Telegram client.
+//    * MQTT -- Post alert message to MQTT.
 //
 // See below for more details on configuring each handler.
 //
@@ -350,6 +351,10 @@ type AlertNode struct {
 	// Send alert to Talk.
 	// tick:ignore
 	TalkHandlers []*TalkHandler `tick:"Talk"`
+
+	// Send alert to MQTT
+	// tick:ignore
+	MQTTHandlers []*MQTTHandler `tick:"Mqtt"`
 
 	// Send alert using SNMPtraps.
 	// tick:ignore
@@ -999,6 +1004,31 @@ type AlertaHandler struct {
 func (a *AlertaHandler) Services(service ...string) *AlertaHandler {
 	a.Service = service
 	return a
+}
+
+// Send alert to an MQTT broker
+// tick:property
+func (a *AlertNode) Mqtt(topic string) *MQTTHandler {
+	m := &MQTTHandler{
+		AlertNode: a,
+		Topic:     topic,
+	}
+	a.MQTTHandlers = append(a.MQTTHandlers, m)
+	return m
+}
+
+type MQTTHandler struct {
+	*AlertNode
+
+	// The topic where alerts will be dispatched to
+	Topic string
+
+	// The QoS that will be used to deliver the alerts
+	QoS int64
+
+	// Retained indicates whether this alert should be delivered to
+	// clients that were not connected to the broker at the time of the alert
+	Retained bool
 }
 
 // Send the alert to Sensu.
--- a/services/mqtt/client.go
+++ b/services/mqtt/client.go
@@ -1,0 +1,105 @@
+package mqtt
+
+import (
+	"errors"
+	"log"
+	"time"
+
+	pahomqtt "github.com/eclipse/paho.mqtt.golang"
+)
+
+// Client describes an immutable MQTT client, designed to accommodate the
+// incongruencies between real clients and mock clients.
+type Client interface {
+	Connect() error
+	Disconnect()
+	Publish(string, byte, bool, string) error
+}
+
+// NewClient produces a disconnected MQTT client
+var NewClient = func(c Config) Client {
+	return &PahoClient{
+		host:     c.Broker(),
+		port:     c.Port,
+		username: c.Username,
+		password: c.Password,
+		clientID: c.ClientID,
+	}
+}
+
+type PahoClient struct {
+	host string
+	port uint16
+
+	username string
+	password string
+
+	clientID string
+
+	client pahomqtt.Client
+}
+
+var _ Client = &PahoClient{}
+
+// DefaultQuiesceTimeout is the duration the client will wait for outstanding
+// messages to be published before forcing a disconnection
+const DefaultQuiesceTimeout time.Duration = 250 * time.Millisecond
+
+func (p *PahoClient) Connect() error {
+	log.Printf("Current config: %#v\n", p)
+	opts := pahomqtt.NewClientOptions()
+	opts.AddBroker(p.host)
+	opts.SetClientID(p.clientID)
+	opts.SetUsername(p.username)
+	opts.SetPassword(p.password)
+
+	// Using a clean session forces the broker to dispose of client session
+	// information after disconnecting. Retention of this is useful for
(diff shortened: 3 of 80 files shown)


pipeline/alert.go
```diff
@@ -350,6 +351,10 @@ type AlertNode struct {
 	// tick:ignore
 	TalkHandlers []*TalkHandler `tick:"Talk"`
 
+	// Send alert to MQTT
+	// tick:ignore
+	MQTTHandlers []*MQTTHandler `tick:"MQTT"`
```
nathanielc: This will read like this in TICKscript:

```
|alert()
    .mQTT()
```

Which is really odd so I think we should change this and the method name below to `Mqtt` so that in TICKscript it is all lower case.

pipeline/alert.go
```diff
@@ -942,6 +947,23 @@ func (a *AlertaHandler) Services(service ...string) *AlertaHandler {
 	return a
 }
 
+// Send alert to an MQTT broker
+// tick:property
+func (a *AlertNode) MQTT() *MQTTHandler {
```
nathanielc: Can we accept the topic string here in this method? That way it reads like 

```
|alert()
    .mqtt('topic goes here')
```

Since there is no other config for the MQTT handler this keeps it simple.

services/mqtt/config.go
```diff
@@ -0,0 +1,45 @@
...
+
+type Config struct {
+	// Enabled indicates whether the service should be enabled
+	Enabled bool `toml:"enabled" override:"enabled"`
+	// URL of the MQTT Broker
+	Host string `toml:"host" override:"host"`
+	// Port of the MQTT Broker
+	Port uint16 `toml:"port" override:"port"`
+
+	ClientID string `toml:"client_id" override:"client_id"`
```
nathanielc: Kapacitor uses dashes instead of underscores. Can you update this one and the others in this config struct?

services/mqtt/config.go
```diff
@@ -0,0 +1,45 @@
...
+	Password string `toml:"password" override:"password"`
+
+	DefaultTopic string   `toml:"default_topic" override:"default_topic"`
+	DefaultQOS   QOSLevel `toml:"default_qos" override:"default_qos"`
+}
+
+// Broker formats the configured Host and Port as tcp://host:port, suitable for
+// consumption by the Paho MQTT Client
+func (c Config) Broker() string {
+	return fmt.Sprintf("tcp://%s:%d", c.Host, c.Port)
```
nathanielc: What if the host is an IPv6 address with colon etc? I believe you can use the URL type here to safely create the Broker string without worrying about edge cases.

services/mqtt/service.go
```diff
@@ -0,0 +1,151 @@
...
+	"fmt"
+	"log"
+	"sync/atomic"
+	"time"
+
+	pahomqtt "github.com/eclipse/paho.mqtt.golang"
+	"github.com/influxdata/kapacitor/alert"
+)
+
+type QOSLevel byte
```
nathanielc: Is QOSLevel a Qualty of Service Level? If so should we name it QoSLevel?

services/mqtt/service.go
```diff
@@ -0,0 +1,151 @@
...
+
+const (
+	// best effort delivery. "fire and forget"
+	ATMOSTONCE QOSLevel = iota
```
nathanielc: No need for constants to be all upper case in Go, go ahead and use Upper Camel Case here.

services/mqtt/service.go
```diff
@@ -0,0 +1,151 @@
...
+	// guarantees delivery only once. Safest and slowest.
+	EXACTLYONCE
+)
+
+// DEFAULT_QUIESCE_TIMEOUT is the duration the client will wait for outstanding
+// messages to be published before forcing a disconnection
+const DEFAULT_QUIESCE_TIMEOUT time.Duration = 250 * time.Millisecond
+
+type Service struct {
+	Logger *log.Logger
```
nathanielc: Logger should be private.

services/mqtt/service.go
```diff
@@ -0,0 +1,151 @@
...
+}
+
+func (s *Service) Update(newConfig []interface{}) error {
+	if l := len(newConfig); l != 1 {
+		return fmt.Errorf("expected only one new config object, got %d", l)
+	}
+	if c, ok := newConfig[0].(Config); !ok {
+		return fmt.Errorf("expected config object to be of type %T, got %T", c, newConfig[0])
+	} else {
+		s.configValue.Store(c)
```
nathanielc: How do the changes to the config get propagated to the client?

services/mqtt/service.go
```diff
@@ -0,0 +1,151 @@
...
+	return &handler{
+		s:      s,
+		c:      c,
+		logger: l,
+	}
+}
+
+type HandlerConfig struct {
+	Topic string
+	QOS   QOSLevel
```
nathanielc: Again QoS is probably a better name here.

services/mqtt/service.go
```diff
@@ -0,0 +1,151 @@
...
+}
+
+func (s *Service) Close() error {
+	s.client.Disconnect(uint(DEFAULT_QUIESCE_TIMEOUT / time.Millisecond))
+	s.Logger.Println("I! MQTT Client Disconnected")
+	return nil
+}
+
+func (s *Service) Alert(qos QOSLevel, topic, message string) error {
+	s.client.Publish(topic, byte(qos), false, message) // should retained be configureable?
```
nathanielc: What does retained do?

vendor.list
```diff
@@ -4,6 +4,7 @@ github.com/cenkalti/backoff
 github.com/davecgh/go-spew
 github.com/dgrijalva/jwt-go v3.0.0
 github.com/dustin/go-humanize
+github.com/eclipse/paho.mqtt.golang
```
nathanielc: Can you rebase your changes? This file is no longer in use in master. Kapacitor has switched over to using the new golang dep tool. Info on how to use it is in the CONTRIBUTING.md file.

services/mqtt/service.go
timraymond: Ah good catch. I meant to research this some more. Publishers have no guarantee that messages will be received by a subscriber to a topic, since there might not be any clients listening at the time the message was published. In that case, the message is dropped by the broker. Retained (which is the 3rd param in `Publish`) sets a flag on the message that causes the broker to hold onto it and deliver it immediately to clients that subsequently subscribe to that topic. The broker holds onto at most one message per topic, so they're useful in describing state ("light switch is on"). Given this, I think the answer to my question is "yes," and it should probably be configureable from within the TICK script.

This entire series of posts has been invaluable: [MQTT Essentials Part 8: Retained Messages](http://www.hivemq.com/blog/mqtt-essentials-part-8-retained-messages)

services/mqtt/service.go
timraymond: lol! That would help, wouldn't it?

timraymond:
@nathanielc okay, I think this is ready for another look (I see that I need to rebase again though). I moved `Topic` into the argument for `mqtt()` at your suggestion. However, I needed to add two more parameters for `QoS` and `Retained`, since the more I read, the more it made sense to control those on an alert by alert basis.

Regarding testing, I've come up with three options:

I read through the spec a bit, and I _think_ I can put together a minimal broker that handles connections and published messages for a QoS of 0, which is what we'd need for a minimal test.

Alternatively, we could wrap the client in an interface and substitute a mock. It seems like that should be possible, but I'm not quite sure how to do the substitution with the existing test machinery that's in place. I think this option is the best, since we're not testing that the MQTT client is implemented correctly.

If we don't want to go that route, we could install Mosquitto in the test containers... but that requires everyone that works on Kapacitor to have a broker running if they want to run the tests, which is kind of 💩 .

Gopkg.toml
```diff
@@ -1,10 +1,66 @@
...
+# version = "1.0.0"
+# branch = "master"
+# revision = "abc123"
+#
+## Optional: specifying an alternate source location as an override will
+## enforce that the alternate location is used for that project, regardless of
+## what source location any dependent projects specify.
+# source = "https://github.com/myfork/package.git"
+
+
```
nathanielc: Can you remove these comments?

services/mqtt/config.go
```diff
@@ -0,0 +1,53 @@
...
+	// Enabled indicates whether the service should be enabled
+	Enabled bool `toml:"enabled" override:"enabled"`
+	// URL of the MQTT Broker
+	Host string `toml:"host" override:"host"`
+	// Port of the MQTT Broker
+	Port uint16 `toml:"port" override:"port"`
+
+	ClientID string `toml:"client-id" override:"client-id"`
+	Username string `toml:"username" override:"username"`
+	Password string `toml:"password" override:"password"`
```
nathanielc: If you update this `override:"password"` to `override:"password,redact"` then the config API will redact the password in all relevant API calls.

services/mqtt/service.go
```diff
@@ -0,0 +1,186 @@
...
+	"sync/atomic"
+	"time"
+
+	pahomqtt "github.com/eclipse/paho.mqtt.golang"
+	"github.com/influxdata/kapacitor/alert"
+)
+
+// QoSLevel indicates the quality of service for messages delivered to a
+// broker.
+type QoSLevel byte
```
nathanielc: Can you implement the TextUnmarshaler interface on this type so its toml representation is not numeric?

services/mqtt/service.go
```diff
@@ -0,0 +1,186 @@
...
+	opts.SetPassword(c.Password)
+
+	// Using a clean session forces the broker to dispose of client session
+	// information after disconnecting. Retention of this is useful for
+	// constrained clients.  Since Kapacitor is only publishing, it has no
+	// storage requirements and can reduce load on the broker by using a clean
+	// session.
+	opts.SetCleanSession(true)
+
+	s.client = pahomqtt.NewClient(opts)
```
nathanielc: Since `s.client` is being modified here you will need to lock access to it or use the atomic.Value type like you do for the config struct.

timraymond:
@nathanielc This should be ready for another look, if you get a chance

nathanielc:
@timraymond  I am not quite done (still tracking down one bug), but I have added a commit that makes it so you have multiple mqtt's configured. I am also adding tests which are not quite passing yet but I figured I push this up before I disappear for the weekend. Take a look when you have a sec, no rush.

services/mqtt/client.go
```diff
@@ -1,71 +1,63 @@
...
+	"github.com/influxdata/kapacitor/tlsconfig"
 )
 
 // Client describes an immutable MQTT client, designed to accommodate the
 // incongruencies between real clients and mock clients.
 type Client interface {
 	Connect() error
 	Disconnect()
-	Publish(string, byte, bool, string) error
+	Publish(topic string, qos QoSLevel, retained bool, message []byte) error
```
nathanielc: It just documents the arguments better. The implementation can still use whatever names it wants.

nathanielc:
@timraymond All tests are passing and I have rebased and squashed commits. If it looks good to you then we can merge it.

nathanielc:
@timraymond Can you add a CHANGELOG entry before merging?

PR #1350, opened 2017-05-01
Update files and static to files-discovery and static-discovery

goller:

### DIFF
--- a/services/scraper/service.go
+++ b/services/scraper/service.go
@@ -3,6 +3,7 @@ package scraper
 import (
 	"fmt"
 	"log"
+	"math"
 	"sync"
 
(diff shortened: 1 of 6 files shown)


services/scraper/service.go
```diff
@@ -118,6 +119,12 @@ func (s *Service) scrape() {
 
 // Append tranforms prometheus samples and inserts data into the tasks pipeline
 func (s *Service) Append(sample *model.Sample) error {
+	value := float64(sample.Value)
+	// Remove all NaN values
+	if math.IsNaN(value) {
```
nathanielc: Can you add a TODO here to count this event?

PR #1355, opened 2017-05-03
Update python agent code to be python 2 and 3 compatible

nathanielc:
Add python3 support to the python UDF agent code.

- [ ] Rebased/mergable
- [ ] Tests pass
- [ ] CHANGELOG.md updated
- [ ] Add python3 based tests to the server_test.go test suite.

### DIFF
--- a/udf/agent/py/kapacitor/udf/agent.py
+++ b/udf/agent/py/kapacitor/udf/agent.py
@@ -1,12 +1,17 @@
 # Kapacitor UDF Agent implementation in Python
 #
 # Requires protobuf v3
-#   pip install protobuf==3.0.0b2
+#   pip install protobuf==3.0.0
 
 import sys
-import udf_pb2
+from . import udf_pb2
 from threading import Lock, Thread
-from Queue import Queue
+
+try:
+    from queue import Queue
+except ImportError:
+    from Queue import Queue
+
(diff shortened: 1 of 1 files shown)


rca:
Trying this now; first thing to get this working was to use `protobuf==3.5.1`

rca:
I ran into a number of encoding issues that are somewhat sorted out with PR #1750.  I am able to receive streaming data into the UDF, but I'm still not able to successfully send data points out of the UDF.

lalat-das:
Applied both the patch i.e. 
1. pointed by Nathanielc  ( it describes the import specific changes)
2. Then applied "rca"(Roberto)'s patch which describes about using .encode() and making stdin & stdout as .buffer type.

I am able to send and recieve point but as long as the point is 80 columns length. 

If the line protocol complaint point goes beyond 80 column, then EOF exception is raised while reading the size of data in read_loop function.
size = decodeUvarint32(self._in)

 This EOF exception pulls the thread out of the infinite loop then the wait() call ends and the agent process no more able to receive any further points.

Notsure may be because the stin is filled with two consecutive EOF may be because of rca patchmake input as sys.stdin.buffer. I am just speculating it.

omdv:
Hello, is this going to be merged?

Issue #1319, opened 2017-04-12
Refactor Edge type to use Message Passing semantics


nathanielc:
Edges between nodes of the task DAG in Kapacitor are typed, meaning that they send either stream or batch data. The current implementation uses method calls on the Edge type in order to move data around. This was a simple first implementation but it has a few short comings that are causing problems.

* Nodes can only communicate  via data, but there is a need to also send meta data along the edges. For example a barrier message or delete group message are two use cases that we currently have.
* Batch data arrives in a single result, meaning that all nodes have to buffer the entire batch in RAM to perform the operation even if the operation does need to do so (i.e. count doesn't need the entire batch in RAM at once).
* All node implementations have to handle groups manually. 

The proposal is to switch to a [Message Passing Interface](https://en.wikipedia.org/wiki/Message_Passing_Interface) edge type. This has a few advantages:

* Metadata can easily be communicated in order with the data.
* Nodes can choose to buffer the batch data as necessary.
* It becomes easier to wrap higher level types around an edge type such that you can have a grouped edge, thus freeing nodes to do the important work and offloading groupby semantics.
* Easier to abstract edge communication over a network.

There are a few disadvantages of MPI:

* Every message must be inspected for its type which add overhead. (we expect this to be negligible).
* More complexity in consuming data from an edge because we plan to introduce meta data that needs to be processed in order with the real data.

There are several issues that we can tackle after this refactor is complete:

*  #1255 - help reduce memory usage
* #1039 - be able to implement a smart fill operation
* #1157, #598 - be able to expire groups and reduce cardinality

nathanielc:
This was completed a long time ago :)

PR #1371, opened 2017-05-10
Add missing working_cardinality stats on state* nodes

### LINKED ISSUE #1370: New StateDuration and StateCount nodes are missing the working_cardinality
The work to add working_cardinality to all nodes happened concurrently with the work to add the stateDuration and stateCount nodes. As a result the working_cardinality did not get added to the state nodes. It needs to be added.

nathanielc:
Fixes #1370

- [x] Rebased/mergable
- [x] Tests pass
- [x] CHANGELOG.md updated

### DIFF
--- a/state_tracking.go
+++ b/state_tracking.go
@@ -3,8 +3,10 @@ package kapacitor
 import (
 	"fmt"
 	"log"
+	"sync"
 	"time"
 
+	"github.com/influxdata/kapacitor/expvar"
 	"github.com/influxdata/kapacitor/models"
 	"github.com/influxdata/kapacitor/pipeline"
 	"github.com/influxdata/kapacitor/tick/ast"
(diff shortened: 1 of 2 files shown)


state_tracking.go
```diff
@@ -28,30 +30,53 @@ type StateTrackingNode struct {
...
+	groupsMu sync.RWMutex
+	groups   map[models.GroupID]*stateTrackingGroup
 }
 
 func (stn *StateTrackingNode) group(g models.GroupID) (*stateTrackingGroup, error) {
+	stn.groupsMu.RLock()
 	stg := stn.groups[g]
-	if stg == nil {
-		stg = &stateTrackingGroup{}
+	stn.groupsMu.RUnlock()
```
phemmer: Just for my own knowledge, how would the node be evaluated in parallel? When I looked through the code, I didn't see anything which would result in a node's `runF` being called multiple times in parallel. Multiple different nodes, sure, but I didn't see anything for the same node.

nathanielc: The `valueF` function is being passed into the NewIntFuncGauge, which will call and subsequently read the value from another goroutine.

phemmer: Ah, yes. (because of line 48) I was looking at this from the angle thinking that multiple `runF`s would happen in parallel, not that the `valueF` could be called asynchronously.

Thanks.

state_tracking.go
```diff
@@ -28,30 +30,53 @@ type StateTrackingNode struct {
...
-			return nil, fmt.Errorf("Failed to compile expression: %v", err)
-		}
+	if stg == nil {
+		// Grab the write lock
+		stn.groupsMu.Lock()
+		defer stn.groupsMu.Unlock()
+
+		// Check again now that we have the write lock
+		stg = stn.groups[g]
+		if stg == nil {
```
phemmer: Sorry, I know this is merged, but this is just bugging me.
Having this code implies that you can be processing 2 different points for the same node, and the same group at the same time. This isn't the case now, and I don't think it ever will be the case, because otherwise you'd have race conditions where the node might not emit points in the same order in which they came in.

nathanielc: @phemmer You are correct, its just an automatic pattern I follow when using read write mutexes to check the read after acquiring the write lock. It defensive programming. But you are correct since the only place a write lock is acquired is on line 45 then there is technically no need to perform the extra read and nil comparison. But since it is a small cost to pay and is only paid once per new group, I'd rather leave the defensive code in place. That way if later someone sees there is a RWMutex being used they can just grab the write lock and do what they need without having to read all existing locking code.

Thanks for the questions, they are very welcome.

nathanielc: @phemmer On a slightly related note, with the work to change everything to be message passing internally (see #1319), then we can also refactor much of the code to have grouping semantics outside of each node. This way most of the locking logic etc can be moved to a common place. Which I believe is what you expected to see when you first dug into the code. In other words we are headed in that direction and sooner rather than later, so that we don't build up too much momentum in the current state.

Issue #1379, opened 2017-05-16
eval & others nodes add fields to wrong point when in batch


phemmer:
Input data:
```
test v=1i 0s
test v=2i 1s
```
Script:
```
var data = stream|from().measurement('test')
	|window().period(1s).every(0s)
data
	|eval(lambda: "v" + 1).keep().as('v_plus_one')
	|log()
data
	|eval(lambda: "v" + 2).keep().as('v_plus_two')
	|log()
```
Output:
```
[test.tick:log4] 2017/05/16 10:52:22 I!  {"name":"test","tmax":"1970-01-01T00:00:00Z","points":[{"time":"1970-01-01T00:00:00Z","fields":{"v":1,"v_plus_two":3},"tags":{}}]}
[test.tick:log4] 2017/05/16 10:52:22 I!  {"name":"test","tmax":"1970-01-01T00:00:01Z","points":[{"time":"1970-01-01T00:00:01Z","fields":{"v":2,"v_plus_one":3,"v_plus_two":4},"tags":{}}]}
[test.tick:log6] 2017/05/16 10:52:22 I!  {"name":"test","tmax":"1970-01-01T00:00:00Z","points":[{"time":"1970-01-01T00:00:00Z","fields":{"v":1,"v_plus_two":3},"tags":{}}]}
[test.tick:log6] 2017/05/16 10:52:22 I!  {"name":"test","tmax":"1970-01-01T00:00:01Z","points":[{"time":"1970-01-01T00:00:01Z","fields":{"v":2,"v_plus_one":3,"v_plus_two":4},"tags":{}}]}
```
Notice that some of the points have both `v_plus_one` and `v_plus_two`, and that none of the points have just `v_plus_one`.
*You may need to run this a few times as it is a race condition*

I noticed this while fixing the map race reported in #1369. The design pattern in the `stateDuration`, `eval`, `default` node, and others results in a race as multiple nodes all end up iterating over the same `Batch` object, and thus add data to the wrong point.

desa:
@phemmer Good find. Any thoughts on what the expected behavior should be?

phemmer:
```
[test.tick:log4] 2017/05/16 10:52:22 I!  {"name":"test","tmax":"1970-01-01T00:00:00Z","points":[{"time":"1970-01-01T00:00:00Z","fields":{"v":1,"v_plus_one":2},"tags":{}}]}
[...]

Issue #1369, opened 2017-05-10
kapacitor kept restarting with mutiple stateDuration (kapacitor 1.3.0~beta2)


wshi5985:
when we use one stateDuration in data stream, it works fine, but if check multiple event's stateDuration with same data stream,  kapacitor kept  restarting. 

```
$ grep "Kapacitor starting" /logs/kapacitor/kapacitor.log |tail
[run] 2017/05/09 13:05:24 I! Kapacitor starting, version 1.3.0~beta2, branch master, commit 46de6ebb7b51                 7b6127e5122396552148e1b8e2f8
[run] 2017/05/09 13:17:25 I! Kapacitor starting, version 1.3.0~beta2, branch master, commit 46de6ebb7b51                 7b6127e5122396552148e1b8e2f8
[run] 2017/05/09 13:32:26 I! Kapacitor starting, version 1.3.0~beta2, branch master, commit 46de6ebb7b51                 7b6127e5122396552148e1b8e2f8
[run] 2017/05/09 13:57:27 I! Kapacitor starting, version 1.3.0~beta2, branch master, commit 46de6ebb7b51                 7b6127e5122396552148e1b8e2f8
[run] 2017/05/09 14:16:22 I! Kapacitor starting, version 1.3.0~beta2, branch master, commit 46de6ebb7b51  
```

```
var response_data = stream
    |from()
        .database('test')
        .measurement(measurement)
        .groupBy('domain','monitor_name')
    |window()
        .period(5m)
        .every(1m)

response_data
    |stateDuration(lambda: int("result_status") != 1 )
        .unit(1m)
        .as('resultstatus_events_duration')
    |stateDuration(lambda: int("httpcode_status") != 1 )
        .unit(1m)
        .as('httpcode_events_duration')
    |stateDuration(lambda: float("response_time") > float("responsetime_threshold") )
        .unit(1m)
        .as('responsetime_events_duration')
```

phemmer:
That's really odd that it just exits without displaying any error. If you run it by hand, does it display anything different? Does `dmesg` show anything from the kernel saying the kernel is killing the process?

nathanielc:
@wshi5985 I think I have been able to reproduce the issue. Running the task above with about 1000 writes per second, I can see a steady increase in RAM usage of Kapacitor. My guess is that the Linux oom killer is killing the Kapacitor process.

[...]

PR #1380, opened 2017-05-16
fix stateDuration & stateCount adding fields to wrong point in batch

phemmer:
This fixes the map write race condition reported in https://github.com/influxdata/kapacitor/issues/1369#issuecomment-301778741 as well as the point sharing issue reported in #1379, but only for stateDuration & stateCount. If this is the solution we want to go with (copying the batch), I can implement on the other nodes.

- [X] Rebased/mergable
- [X] Tests pass
- [ ] CHANGELOG.md updated

### DIFF
--- a/integrations/batcher_test.go
+++ b/integrations/batcher_test.go
@@ -2363,15 +2363,18 @@ batch
 
 func TestBatch_StateDuration(t *testing.T) {
 	var script = `
-batch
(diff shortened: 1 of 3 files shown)


nathanielc:
@phemmer Thanks, I'll merge and add a CHANGELOG entry.

PR #1385, opened 2017-05-17
Fix panic in InfluxQL node with missing field

desa:
- [x] Rebased/mergable
- [x] Tests pass
- [x] CHANGELOG.md updated

### DIFF
--- a/influxql.go
+++ b/influxql.go
@@ -89,7 +89,7 @@ func (n *InfluxQLNode) runStreamInfluxQL() error {
 		mu.RLock()
 		context := contexts[p.Group]
 		mu.RUnlock()
-		// Fisrt point in window
+		// First point in window
 		if context == nil {
 			// Create new context
 			c := baseReduceContext{
@@ -103,7 +103,16 @@ func (n *InfluxQLNode) runStreamInfluxQL() error {
 				pointTimes: n.n.PointTimes || n.isStreamTransformation,
 			}
 
-			k := reflect.TypeOf(p.Fields[c.field]).Kind()
+			f, exists := p.Fields[c.field]
+			if !exists {
+				n.incrementErrorCount()
+				n.logger.Printf("E! field %s missing from point, skipping point", c.field)
+				p, ok = n.ins[0].NextPoint()
+				n.timer.Stop()
+				continue
+			}
+
+			k := reflect.TypeOf(f).Kind()
 			kindChanged := k != kind
 			kind = k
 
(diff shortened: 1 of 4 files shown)


desa:
@nathanielc This should be good to go.

influxql.go
```diff
@@ -187,7 +187,13 @@ func (n *InfluxQLNode) runBatchInfluxQL() error {
 				kind = reflect.Float64
 			}
 		} else {
-			k := reflect.TypeOf(b.Points[0].Fields[c.field]).Kind()
+			f, ok := b.Points[0].Fields[c.field]
```
nathanielc: Are there other places we read from the field map?

desa: Whoops there are indeed.

integrations/data/TestStream_InfluxQLNodeMissingValue.srpl
```diff
@@ -0,0 +1,39 @@
...
+packets value=1007 0000000008
+dbname
+rpname
+packets value=1008 0000000009
+dbname
+rpname
+packets value=1009 0000000010
+dbname
+rpname
+packets is_missing_value=1011 0000000011
```
desa: @nathanielc 
Notice how in the test `TestStream_InfluxQLNodeMissingValue_Stream` the data's timestamp is off by one. I'm not sure if that is intentional behavior or not.

```go
 +	er := models.Result{
 +		Series: models.Rows{
 +			{
 +				Name:    "packets",
 +				Tags:    nil,
 +				Columns: []string{"time", "mean"},
 +				Values: [][]interface{}{[]interface{}{
 +					time.Date(1971, 1, 1, 0, 0, 10, 0, time.UTC),
 +					1011.0,
 +				}},
 +			},
 +		},
 +	}
```

nathanielc: The time values are relative and zero based, since the first time in the file is 0001 then the zero time is off by one, hence the difference.

desa: ah I see.

influxql.go
```diff
@@ -187,7 +196,13 @@ func (n *InfluxQLNode) runBatchInfluxQL() error {
...
+			if !ok {
+				n.incrementErrorCount()
+				n.logger.Printf("E! field %s missing from point cannot get type\n", c.field)
```
nathanielc: I think you need to stop the timer here as well.

Also, this skips the entire batch if the first point is missing the field. I think that is ok behavior but we should update the log message so that it is clear. Something like `E! field %s missing, skipping batch`

influxql.go
```diff
@@ -187,7 +196,14 @@ func (n *InfluxQLNode) runBatchInfluxQL() error {
 				kind = reflect.Float64
 			}
 		} else {
-			k := reflect.TypeOf(b.Points[0].Fields[c.field]).Kind()
+			f, ok := b.Points[0].Fields[c.field]
+			if !ok {
+				n.incrementErrorCount()
+				n.logger.Printf("E! field %s missing from point, skipping batch\n", c.field)
```
nathanielc: Small nit, but there is no need for a new line here as the logger will do that already.

Also it looks like you need to rebase as the CHANGELOG diff looks out of date.

desa: Fixed

PR #1389, opened 2017-05-19
Copy batch points slice before modification

### LINKED ISSUE #1379: eval & others nodes add fields to wrong point when in batch

nathanielc:
Fixes #1379 
The idea of this change is to shallow copy the batch points slice so that nodes which modify the batch points slice do not collide.

The existing code already takes care to copy the maps before modification, but that didn't matter since the underlying slice was shared. This change copies the slice to prevent those issues.

This is horrible :), I really dislike how this came together and we need a better way to manage copying/modifying data. My hope is that the migration to message passing #1319 will open up a way to manage this better. There is simply too much onus on each implementation of a node to keep everything straight and not introduce new bugs.

- [x] Rebased/mergable
- [x] Tests pass
- [x] CHANGELOG.md updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased [unreleased]
 
+# Bugfixes
+
+- [#1379](https://github.com/influxdata/kapacitor/issues/1379): Copy batch points slice before modification, fixes potential panics and data corruption.
+
 ## v1.3.0-rc3 [2017-05-18]
 
 ### Release Notes
--- a/default.go
+++ b/default.go
@@ -54,6 +54,7 @@ func (e *DefaultNode) runDefault(snapshot []byte) error {
 	case pipeline.BatchEdge:
 		for b, ok := e.ins[0].NextBatch(); ok; b, ok = e.ins[0].NextBatch() {
 			e.timer.Start()
+			b.Points = b.ShallowCopyPoints()
 			_, b.Tags = e.setDefaults(nil, b.Tags)
 			b.UpdateGroup()
 			for i := range b.Points {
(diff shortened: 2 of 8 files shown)


phemmer:
Looks like alert.go needs to be adjusted too. It's modifying `b.Points` without a copy: https://github.com/influxdata/kapacitor/blob/master/alert.go#L671-L691

nathanielc:
@phemmer Thanks see #1391

PR #1390, opened 2017-05-19
WIP: Add built in functions to convert timestamps to integers

desa:
This PR introduces the ability to convert timestamps to integers. The purpose of this is to allow users to perform mathematical operations using the timestamp. 

- [ ] Rebased/mergable
- [ ] Tests pass
- [ ] CHANGELOG.md updated

### DIFF
--- a/tick/stateful/functions.go
+++ b/tick/stateful/functions.go
@@ -168,6 +168,7 @@ func init() {
 	statelessFuncs["isPresent"] = isPresent{}
 
 	// Time functions
+	statelessFuncs["unixNano"] = unixNano{}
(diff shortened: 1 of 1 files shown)


nathanielc:
Thanks, have you seen this feature request https://github.com/influxdata/kapacitor/issues/169?

desa:
@nathanielc I had not, but that would be useful. In the use-case that this was spawned from, the user was writing a timestamp as a field and comparing the drift of the `time` and the field storing a timestamp.

nathanielc:
So this is more about being able to create a field from a time value, makes sense.

PR #1391, opened 2017-05-19
shallow copy alert batch points

nathanielc:


phemmer:
Looks good. This is the only other case I had found that could modify a shared `b.Points`.

Issue #1399, opened 2017-05-23
Malformed log entry from golang http library


desa:
```
log messages must have 'L!' prefix where L is one of 'D', 'I', 'W', 'E'[log] 2017/05/23 13:37:30 http: Accept error: accept tcp [::]:9092: accept: too many open files; retrying in 1s
log messages must have 'L!' prefix where L is one of 'D', 'I', 'W', 'E'[log] 2017/05/23 13:37:32 http: Accept error: accept tcp [::]:9092: accept: too many open files; retrying in 1s
log messages must have 'L!' prefix where L is one of 'D', 'I', 'W', 'E'[log] 2017/05/23 13:37:37 http: Accept error: accept tcp [::]:9092: accept: too many open files; retrying in 1s
log messages must have 'L!' prefix where L is one of 'D', 'I', 'W', 'E'[log] 2017/05/23 13:37:40 http: Accept error: accept tcp [::]:9092: accept: too many open files; retrying in 1s
log messages must have 'L!' prefix where L is one of 'D', 'I', 'W', 'E'[log] 2017/05/23 13:37:42 http: Accept error: accept tcp [::]:9092: accept: too many open files; retrying in 1s
log messages must have 'L!' prefix where L is one of 'D', 'I', 'W', 'E'[log] 2017/05/23 13:37:47 http: Accept error: accept tcp [::]:9092: accept: too many open files; retrying in 1s
log messages must have 'L!' prefix where L is one of 'D', 'I', 'W', 'E'[log] 2017/05/23 13:37:50 http: Accept error: accept tcp [::]:9092: accept: too many open files; retrying in 1s
log messages must have 'L!' prefix where L is one of 'D', 'I', 'W', 'E'[log] 2017/05/23 13:37:52 http: Accept error: accept tcp [::]:9092: accept: too many open files; retrying in 1s
log messages must have 'L!' prefix where L is one of 'D', 'I', 'W', 'E'[log] 2017/05/23 13:37:57 http: Accept error: accept tcp [::]:9092: accept: too many open files; retrying in 1s
```

nathanielc:
See https://golang.org/pkg/net/http/#Server, we can provide a logger to the server so these message can correctly follow the wlog format.

desa:
Going  to close this as it was delivered a while ago.

PR #1402, opened 2017-05-24
Add http server error logger

desa:
Fixes issue #1399 

- [x] Rebased/mergable
- [x] Tests pass
- [x] CHANGELOG.md updated

### DIFF
--- a/services/httpd/service.go
+++ b/services/httpd/service.go
@@ -37,7 +37,8 @@ type Service struct {
 
 	Handler *Handler
 
-	logger *log.Logger
+	logger           *log.Logger
+	httpServerLogger *log.Logger
 }
 
 func NewService(c Config, hostname string, l *log.Logger, li logging.Interface) *Service {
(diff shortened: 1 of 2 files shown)


services/httpd/service.go
```diff
@@ -68,7 +69,8 @@ func NewService(c Config, hostname string, l *log.Logger, li logging.Interface)
 			li,
 			c.SharedSecret,
 		),
-		logger: l,
+		logger:           l,
+		httpServerLogger: li.NewStaticLevelLogger("[go-http]", log.LstdFlags, logging.ERROR),
```
nathanielc: I think this should have the same name as the rest of the httpd service logging, `"[httpd]"`. Thoughts?

desa: I was thinking that originally, but I wanted there to be a distinction between there being an error in the `httpd` service and there being an error in the http server.

The more I think of it,`httpd` is better.

PR #1413, opened 2017-05-30
Add support for subscription modes.

nathanielc:
This change allows for either the cluster or server ID to be used for establishing uniqueness of subscriptions. The result is that feach server can have its own subscription instead of all servers in a cluster sharing the same subscriptions.

This change includes a bit of a refactor of the `vars` package to expose an interface instead of using the vars directly. 

- [x] Rebased/mergable
- [x] Tests pass
- [x] CHANGELOG.md updated


nathanielc:
@desa I added the configuration option.

PR #1414, opened 2017-05-31
fix derivative preserving fields from wrong point in stream

phemmer:
#1333 changed the derivative node to add the derivative to the current point instead of the previous one. However due to a typo in the code, for stream nodes, it effectively resulted the emitted point containing the values of the previous point, the timestamp of the current point, and the derivative of the current point.

When using batch, the result was as it's supposed to be (values, timestamp, and derivative all from the current point). The batch code had the same typo, it worked because of ordering of the code.

Unfortunately there were no tests that would have caught the behavior, so I added some.

- [X] Rebased/mergable
- [X] Tests pass
- [x] CHANGELOG.md updated


nathanielc:
@phemmer Thanks! I'll take a look at this tomorrow.

nathanielc:
@phemmer We are planning to release a version 1.3.1 with this fix. Can you change the PR to target the `v1.3` branch? Thanks

phemmer:
branch changed.

desa:
LGTM 👍

desa:
@phemmer can you rebase onto `v1.3` and drop the other commits?

phemmer:
Whoops. Didn't notice that after I switched the branch. Done.
Should I also add a note to the CHANGELOG?

nathanielc:
@phemmer 

> Should I also add a note to the CHANGELOG?

Yes please, I added a 1.3.1 section.

phemmer:
done

PR #1481, opened 2017-07-17
Add ability to load tasks/handlers from dir

desa:
# Todo

- [x] Add batch tasks to example of mock directories.

# Done
- [x] Load files from directory
- [x] Extend API to allow for new tickscript definition
- [x] Update Tickscript definition
- [x] Update Template var defition
- [x] Update Handler Definition

# Todo
- [x] Add tests
- [x] Reload directory on OS signal
- [x] Clean up

- [x] Rebased/mergable
- [x] Tests pass
- [x] CHANGELOG.md updated

### DIFF
--- a/client/v1/client.go
+++ b/client/v1/client.go
@@ -9,6 +9,7 @@ import (
 	"io"
 	"io/ioutil"
 	"net/http"
+	"net/http/httptest"
 	"net/url"
 	"path"
 	"strconv"
@@ -73,6 +74,10 @@ type Config struct {
 
 	// Optional credentials for authenticating with the server.
 	Credentials *Credentials
+
+	// Optional Transport https://golang.org/pkg/net/http/#RoundTripper
+	// If nil the default transport will be used
+	Transport http.RoundTripper
 }
 
 // AuthenticationMethod defines the type of authentication used.
@@ -118,6 +123,23 @@ func (c Credentials) Validate() error {
 	return nil
 }
 
+type localTransport struct {
+	h http.Handler
+}
+
+func (l *localTransport) RoundTrip(r *http.Request) (*http.Response, error) {
+	w := httptest.NewRecorder()
+	l.h.ServeHTTP(w, r)
+
+	return w.Result(), nil
+}
+
+func NewLocalTransport(h http.Handler) http.RoundTripper {
+	return &localTransport{
+		h: h,
+	}
+}
+
 // Basic HTTP client
 type Client struct {
 	url         *url.URL
@@ -148,21 +170,28 @@ func New(conf Config) (*Client, error) {
 		}
 	}
 
-	tr := &http.Transport{
-		Proxy: http.ProxyFromEnvironment,
-		TLSClientConfig: &tls.Config{
-			InsecureSkipVerify: conf.InsecureSkipVerify,
-		},
-	}
-	if conf.TLSConfig != nil {
-		tr.TLSClientConfig = conf.TLSConfig
+	rt := conf.Transport
+	var tr *http.Transport
+
+	if rt == nil {
+		tr = &http.Transport{
+			Proxy: http.ProxyFromEnvironment,
+			TLSClientConfig: &tls.Config{
+				InsecureSkipVerify: conf.InsecureSkipVerify,
+			},
+		}
+		if conf.TLSConfig != nil {
+			tr.TLSClientConfig = conf.TLSConfig
+		}
+
+		rt = tr
 	}
 	return &Client{
 		url:       u,
 		userAgent: conf.UserAgent,
 		httpClient: &http.Client{
 			Timeout:   conf.Timeout,
-			Transport: tr,
+			Transport: rt,
 		},
 		credentials: conf.Credentials,
 	}, nil
@@ -203,8 +232,9 @@ type ExecutionStats struct {
 type TaskType int
 
 const (
-	StreamTask TaskType = 1
-	BatchTask  TaskType = 2
+	InvalidTask TaskType = 0
+	StreamTask  TaskType = 1
+	BatchTask   TaskType = 2
 )
 
 func (tt TaskType) MarshalText() ([]byte, error) {
@@ -213,6 +243,8 @@ func (tt TaskType) MarshalText() ([]byte, error) {
 		return []byte("stream"), nil
 	case BatchTask:
 		return []byte("batch"), nil
+	case InvalidTask:
+		return []byte("invalid"), nil
 	default:
 		return nil, fmt.Errorf("unknown TaskType %d", tt)
 	}
@@ -224,6 +256,8 @@ func (tt *TaskType) UnmarshalText(text []byte) error {
 		*tt = StreamTask
 	case "batch":
 		*tt = BatchTask
+	case "invalid":
+		*tt = InvalidTask
 	default:
 		return fmt.Errorf("unknown TaskType %s", s)
 	}
@@ -513,9 +547,9 @@ func (vs *Vars) UnmarshalJSON(b []byte) error {
 }
 
 type Var struct {
-	Type        VarType     `json:"type"`
-	Value       interface{} `json:"value"`
-	Description string      `json:"description"`
+	Type        VarType     `json:"type" yaml:"type"`
+	Value       interface{} `json:"value" yaml:"value"`
+	Description string      `json:"description" yaml:"description"`
 }
 
 // A Task plus its read-only attributes.
@@ -725,13 +759,13 @@ func (c *Client) StorageLink(name string) Link {
 }
 
 type CreateTaskOptions struct {
-	ID         string     `json:"id,omitempty"`
-	TemplateID string     `json:"template-id,omitempty"`
+	ID         string     `json:"id,omitempty" yaml:"id"`
+	TemplateID string     `json:"template-id,omitempty" yaml:"template-id"`
 	Type       TaskType   `json:"type,omitempty"`
-	DBRPs      []DBRP     `json:"dbrps,omitempty"`
+	DBRPs      []DBRP     `json:"dbrps,omitempty" yaml:"dbrps"`
 	TICKscript string     `json:"script,omitempty"`
 	Status     TaskStatus `json:"status,omitempty"`
-	Vars       Vars       `json:"vars,omitempty"`
+	Vars       Vars       `json:"vars,omitempty" yaml:"vars"`
 }
 
 // Create a new task.
@@ -759,13 +793,13 @@ func (c *Client) CreateTask(opt CreateTaskOptions) (Task, error) {
 }
 
 type UpdateTaskOptions struct {
-	ID         string     `json:"id,omitempty"`
-	TemplateID string     `json:"template-id,omitempty"`
+	ID         string     `json:"id,omitempty" yaml:"id"`
+	TemplateID string     `json:"template-id,omitempty" yaml:"template-id"`
 	Type       TaskType   `json:"type,omitempty"`
-	DBRPs      []DBRP     `json:"dbrps,omitempty"`
+	DBRPs      []DBRP     `json:"dbrps,omitempty" yaml:"dbrps"`
 	TICKscript string     `json:"script,omitempty"`
 	Status     TaskStatus `json:"status,omitempty"`
-	Vars       Vars       `json:"vars,omitempty"`
+	Vars       Vars       `json:"vars,omitempty" yaml:"vars"`
 }
 
 // Update an existing task.
@@ -1963,6 +1997,7 @@ func (c *Client) TopicHandler(link Link) (TopicHandler, error) {
 }
 
 type TopicHandlerOptions struct {
+	Topic   string                 `json:"topic" yaml:"topic"`
 	ID      string                 `json:"id" yaml:"id"`
 	Kind    string                 `json:"kind" yaml:"kind"`
 	Options map[string]interface{} `json:"options" yaml:"options"`
@@ -2281,4 +2316,38 @@ func (d *Duration) UnmarshalText(data []byte) error {
 	}
 	*d = Duration(dur)
 	return nil
+}
+
+type DBRPs []DBRP
+
+func (d *DBRPs) String() string {
+	return fmt.Sprint(*d)
+}
+
+type TaskVars struct {
+	ID         string `json:"id,omitempty" yaml:"id"`
+	TemplateID string `json:"template-id,omitempty" yaml:"template-id"`
+	DBRPs      []DBRP `json:"dbrps,omitempty" yaml:"dbrps"`
+	Vars       Vars   `json:"vars,omitempty" yaml:"vars"`
+}
+
+func (t TaskVars) CreateTaskOptions() (CreateTaskOptions, error) {
+	o := CreateTaskOptions{
+		ID:         t.ID,
+		TemplateID: t.TemplateID,
+		Vars:       t.Vars,
+		DBRPs:      t.DBRPs,
+	}
+
+	return o, nil
+}
+
+func (t TaskVars) UpdateTaskOptions() (UpdateTaskOptions, error) {
+	o := UpdateTaskOptions{
+		ID:         t.ID,
+		TemplateID: t.TemplateID,
+		Vars:       t.Vars,
+		DBRPs:      t.DBRPs,
+	}
+	return o, nil
 }
--- a/cmd/kapacitor/main.go
+++ b/cmd/kapacitor/main.go
@@ -542,6 +542,7 @@ var (
 	dtype       = defineFlags.String("type", "", "The task type (stream|batch)")
 	dtemplate   = defineFlags.String("template", "", "Optional template ID")
 	dvars       = defineFlags.String("vars", "", "Optional path to a JSON vars file")
+	dfile       = defineFlags.String("file", "", "Optional path to a YAML or JSON template task file")
 	dnoReload   = defineFlags.Bool("no-reload", false, "Do not reload the task even if it is enabled")
 	ddbrp       = make(dbrps, 0)
 )
@@ -687,33 +688,89 @@ func doDefine(args []string) error {
 		}
 	}
 
+	fileVars := client.TaskVars{}
+	if *dfile != "" {
+		f, err := os.Open(*dfile)
+		if err != nil {
+			return errors.Wrapf(err, "failed to open file %s", *dfile)
+		}
+		data, err := ioutil.ReadAll(f)
+		if err != nil {
+			return errors.Wrapf(err, "failed to read task vars file %q", *dfile)
+		}
+		defer f.Close()
+		switch ext := path.Ext(*dfile); ext {
+		case ".yaml", ".yml":
+			if err := yaml.Unmarshal(data, &fileVars); err != nil {
+				return errors.Wrapf(err, "failed to unmarshal yaml task vars file %q", *dfile)
+			}
+		case ".json":
+			if err := json.Unmarshal(data, &fileVars); err != nil {
+				return errors.Wrapf(err, "failed to unmarshal json task vars file %q", *dfile)
+			}
+		default:
+			return errors.New("bad file extension. Must be YAML or JSON")
+
+		}
+	}
+
 	l := cli.TaskLink(id)
 	task, _ := cli.Task(l, nil)
 	var err error
 	if task.ID == "" {
-		_, err = cli.CreateTask(client.CreateTaskOptions{
-			ID:         id,
-			TemplateID: *dtemplate,
-			Type:       ttype,
-			DBRPs:      ddbrp,
-			TICKscript: script,
-			Vars:       vars,
-			Status:     client.Disabled,
-		})
-	} else {
-		_, err = cli.UpdateTask(
-			l,
-			client.UpdateTaskOptions{
+		if *dfile != "" {
+			o, err := fileVars.CreateTaskOptions()
+			if err != nil {
+				return err
+			}
+			_, err = cli.CreateTask(o)
+			if err != nil {
+				return err
+			}
+		} else {
+			o := client.CreateTaskOptions{
+				ID:         id,
 				TemplateID: *dtemplate,
 				Type:       ttype,
 				DBRPs:      ddbrp,
 				TICKscript: script,
 				Vars:       vars,
-			},
-		)
-	}
-	if err != nil {
-		return err
+				Status:     client.Disabled,
+			}
+			_, err = cli.CreateTask(o)
+			if err != nil {
+				return err
+			}
+		}
+	} else {
+		if *dfile != "" {
+			o, err := fileVars.UpdateTaskOptions()
+			if err != nil {
+				return err
+			}
+			_, err = cli.UpdateTask(
+				l,
+				o,
+			)
+			if err != nil {
+				return err
+			}
+		} else {
+			o := client.UpdateTaskOptions{
+				TemplateID: *dtemplate,
+				Type:       ttype,
+				DBRPs:      ddbrp,
+				TICKscript: script,
+				Vars:       vars,
+			}
+			_, err = cli.UpdateTask(
+				l,
+				o,
+			)
+			if err != nil {
+				return err
+			}
+		}
 	}
 
 	if !*dnoReload && task.Status == client.Enabled {
@@ -822,7 +879,7 @@ func doDefineTemplate(args []string) error {
 }
 
 func defineTopicHandlerUsage() {
-	var u = `Usage: kapacitor define-topic-handler <topic id> <handler id> <path to handler spec file>
+	var u = `Usage: kapacitor define-topic-handler <path to handler spec file>
 
 	Create or update a handler.
 
@@ -832,7 +889,7 @@ For example:
 
 	Define a handler using the slack.yaml file:
 
-		$ kapacitor define-handler system my_handler slack.yaml
+		$ kapacitor define-topic-handler slack.yaml
 
 Options:
 `
@@ -840,14 +897,12 @@ Options:
 }
 
 func doDefineTopicHandler(args []string) error {
-	if len(args) != 3 {
-		fmt.Fprintln(os.Stderr, "Must provide a topic ID, a handler ID and a path to a handler file.")
+	if len(args) != 1 {
+		fmt.Fprintln(os.Stderr, "Must provide a path to a handler file.")
 		defineTopicHandlerUsage()
 		os.Exit(2)
 	}
-	topic := args[0]
-	handlerID := args[1]
-	p := args[2]
+	p := args[0]
 	f, err := os.Open(p)
 	if err != nil {
 		return errors.Wrapf(err, "failed to open handler spec file %q", p)
@@ -870,12 +925,11 @@ func doDefineTopicHandler(args []string) error {
 			return errors.Wrapf(err, "failed to unmarshal json handler file %q", p)
 		}
 	}
-	ho.ID = handlerID
 
-	l := cli.TopicHandlerLink(topic, ho.ID)
+	l := cli.TopicHandlerLink(ho.Topic, ho.ID)
 	handler, _ := cli.TopicHandler(l)
 	if handler.ID == "" {
-		_, err = cli.CreateTopicHandler(cli.TopicHandlersLink(topic), ho)
+		_, err = cli.CreateTopicHandler(cli.TopicHandlersLink(ho.Topic), ho)
 	} else {
 		_, err = cli.ReplaceTopicHandler(l, ho)
 	}
--- a/cmd/kapacitord/main.go
+++ b/cmd/kapacitord/main.go
@@ -89,16 +89,30 @@ func (m *Main) Run(args ...string) error {
 		}
 
 		signalCh := make(chan os.Signal, 1)
-		signal.Notify(signalCh, os.Interrupt, syscall.SIGTERM)
+		signal.Notify(signalCh, os.Interrupt, syscall.SIGTERM, syscall.SIGHUP)
 		m.Diag.Info("listening for signals")
 
-		// Block until one of the signals above is received
-		select {
-		case <-signalCh:
-			m.Diag.Info("signal received, initializing clean shutdown...")
-			go func() {
-				cmd.Close()
-			}()
+	Loop:
+		for s := range signalCh {
+			switch s.String() {
+			case syscall.SIGTERM.String():
+				m.Diag.Info("SIGTERM received, initializing clean shutdown...")
+				go func() {
+					cmd.Close()
+				}()
+				break Loop
+
+			case syscall.SIGHUP.String():
+				m.Diag.Info("SIGHUP received, reloading tasks/templates/handlers directory...")
+				cmd.Server.Reload()
+
+			default:
+				m.Diag.Info("signal received, initializing clean shutdown...")
+				go func() {
+					cmd.Close()
+				}()
+				break Loop
+			}
 		}
 
 		// Block again until another signal is received, a shutdown timeout elapses,
--- a/cmd/kapacitord/run/command.go
+++ b/cmd/kapacitord/run/command.go
@@ -130,6 +130,7 @@ func (cmd *Command) Run(args ...string) error {
 	if err := s.Open(); err != nil {
 		return fmt.Errorf("open server: %s", err)
 	}
+
 	cmd.Server = s
 
 	// Begin monitoring the server's error channel.
--- a/examples/load/README.md
+++ b/examples/load/README.md
@@ -1,0 +1,177 @@
+# File based task/template/handler definition
+
+This proposal introduces directory based task, template, and handler definition.
+
+## Configuration
+The load service configuration is specified under the `load` tag of the
+kapacitor configuration file.
+
+```
+[load]
+ enabled = true
+ dir="/path/to/directory"
+```
+
+`dir` specifies the directory where the definition files exist.
+
+The service will attempt to load data from three subdirectories.
+
+The `tasks` directory should contain task tickscripts and and the associated templated task
+definition files (either yaml or json).
+
+The `templates` directory should contain templated tickscripts.
+
+The `handlers` directory will contain will contain topic handler definitions in yaml or json.
+
+## Tasks
+
+Task files must be placed in the `tasks` subdirectory of the load service
+directory. Defining tasks explicitly will be done according to the following scheme:
+
+* `id` - the file name without the tick extension
+* `type` - determined by introspection of the task (stream, batch)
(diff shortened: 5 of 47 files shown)


tick/ast/node.go
```diff
@@ -361,6 +362,49 @@ func (n *BinaryNode) Equal(o interface{}) bool {
...
+func newDBRP(p position, db, rp *ReferenceNode, c *CommentNode) *DBRPNode {
+	return &DBRPNode{
+		position: p,
+		DB:       db,
+		RP:       rp,
+		Comment:  c,
+	}
+}
+
+// TODO: Use Better typing eventually
```
desa: remove

tick/ast/node.go
```diff
@@ -981,6 +1025,100 @@ func newProgram(p position) *ProgramNode {
...
+			dbrps = append(dbrps, dbrpc)
+		default:
+			continue
+		}
+	}
+
+	return dbrps
+}
+
+// TODO: better type here
```
desa: remove

tick/ast/node.go
```diff
@@ -981,6 +1025,100 @@ func newProgram(p position) *ProgramNode {
...
+// TODO: Not sure how I feel about client.TaskType importing from client
+func (n *ProgramNode) TaskType() client.TaskType {
+	tts := []string{}
+	for _, nn := range n.Nodes {
+		switch nn.(type) {
+		case *DeclarationNode:
+			if cn, ok := nn.(*DeclarationNode).Right.(*ChainNode); ok {
+				// TODO: dont repeat logic
+				var n = cn.Left
+			Loop1:
```
desa: rename DeclLoop

tick/ast/node.go
```diff
@@ -981,6 +1025,100 @@ func newProgram(p position) *ProgramNode {
...
+						if ident := n.(*IdentifierNode).Ident; ident == "batch" || ident == "stream" {
+							tts = append(tts, ident)
+						}
+						break Loop1
+					}
+				}
+			}
+		case *ChainNode:
+			var n = nn.(*ChainNode).Left
+		Loop2:
```
desa: rename ChainLoop

tick/ast/parser.go
```diff
@@ -237,12 +237,25 @@ func (p *parser) statement() Node {
...
 	case TokenVar:
 		return p.declaration()
+	case TokenDBRP:
+		return p.dbrp()
 	default:
 		return p.expression()
 	}
 }
 
 //parse a declaration statement
```
desa: parse dbrp statement

tick/ast/parser_test.go
```diff
@@ -127,6 +127,42 @@ func TestParseStatements(t *testing.T) {
 		Root   Node
 		err    error
 	}{
+		// TODO: Allow use of " in dbrp declarations
```
desa: remove

services/load/service.go
```diff
@@ -0,0 +1,404 @@
...
+	return &Service{
+		config: c,
+		logger: l,
+		cli:    cli,
+	}, nil
+}
+
+// TaskFiles gets a slice of all files with the .tick file extension
+// in the configured task directory.
+func (s *Service) TaskFiles() (tickscripts []string, err error) {
```
desa: unexport

services/load/service.go
```diff
@@ -0,0 +1,404 @@
...
+		}
+	}
+
+	return
+}
+
+// TemplateFiles gets a slice of all files with the .tick file extension
+// and any associated files with .json, .yml, and .yaml file extentions
+// in the configured template directory.
+func (s *Service) TemplateFiles() (tickscripts []string, tmplVars []string, err error) {
```
desa: unexport

services/load/service.go
```diff
@@ -0,0 +1,404 @@
...
+			continue
+		}
+	}
+
+	return
+}
+
+// HandlerFiles gets a slice of all files with the .json, .yml, and
+// .yaml file extentions in the configured handler directory.
+func (s *Service) HandlerFiles() ([]string, error) {
```
desa: unexport

services/load/service.go
```diff
@@ -0,0 +1,404 @@
...
+			handlers = append(handlers, filepath.Join(handlersDir, filename))
+		default:
+			continue
+		}
+	}
+
+	return handlers, nil
+}
+
+func (s *Service) Load() error {
```
desa: Add logging

services/load/service.go
```diff
@@ -0,0 +1,404 @@
...
+
+	err = s.loadHandlers()
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (s *Service) loadTickscripts() error {
```
desa: change to load tasks

services/load/service.go
```diff
@@ -0,0 +1,404 @@
...
+	for _, f := range files {
+		if err := s.loadTickscript(f); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func (s *Service) loadTickscript(f string) error {
```
desa: ditto

services/load/service.go
```diff
@@ -0,0 +1,404 @@
...
+	for _, v := range vars {
+		if err := s.loadVars(v); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func (s *Service) loadTemplate(f string) error {
+	file, err := os.Open(f)
```
desa: `defer file.Close()`

services/task_store/service.go
```diff
@@ -865,14 +903,50 @@ func (ts *Service) handleUpdateTask(w http.ResponseWriter, r *http.Request) {
 			updated.Type = BatchTask
 		}
 
+		oldPn, err := ast.NewProgramNodeFromTickscript(updated.TICKscript)
```
desa: Cant actually do it cleanly.

desa:
@nathanielc This should be in a reviewable state. I still need to add a few more tests to the server package so that I can be completely confident in the changes, but it should reviewable.

task.go
```diff
@@ -18,6 +18,7 @@ type TaskType int
 const (
 	StreamTask TaskType = iota
 	BatchTask
+	InvalidTask
```
desa: yes

tick/ast/node.go
```diff
@@ -981,6 +1024,101 @@ func newProgram(p position) *ProgramNode {
...
+	// This should never happen
+	if !ok {
+		return nil, errors.New("invalid TICKscript")
+	}
+
+	return pn, nil
+}
+
+// TODO: Not sure how I feel about client.TaskType importing from client
+func (n *ProgramNode) DBRPs() []client.DBRP {
```
nathanielc: In general importing the client types in the server code makes sense as it ensures that the server and client are using the same types. But here we are not server code but rather an independent package that handles parsing scripts.

I recommend that you use the AST nodes as the public API from the `tick` package and move these methods into the server code where needed.  So it would look something like `dbrpsFromProgram(*ast.ProgramNode) []client.DBRP`. Does that help?

tick/ast/node.go
```diff
@@ -981,6 +1024,101 @@ func newProgram(p position) *ProgramNode {
 	}
 }
 
+func NewProgramNodeFromTickscript(tickscript string) (*ProgramNode, error) {
```
nathanielc: I don't really like this function name, but I get that it is a common need to be able to type assert the ast.Node to its concrete type. Should we just have the `Parse` method return the concrete type?

nathanielc: FWIW The reasons I don't like the name is without reading the function's source I thought it was doing some kind of special operation to return a ProgramNode from a TICKscript instead of just using the Parse method.

Another option would be to move this method to the package where it is consumed as a helper method.

services/load/service.go
```diff
@@ -0,0 +1,416 @@
...
+	"strings"
+	"sync"
+
+	"github.com/ghodss/yaml"
+
+	"github.com/influxdata/kapacitor/client/v1"
+	"github.com/pkg/errors"
+)
+
+var defaultURL = "http://localhost:9092"
```
desa: Yeah I really like that.

client/v1/client.go
```diff
@@ -2282,3 +2289,107 @@ func (d *Duration) UnmarshalText(data []byte) error {
 	*d = Duration(dur)
 	return nil
 }
+
+// TODO: this is also used in kapacitor/main.go
```
desa: Old comment. Previously this code was in `kapacitor/main.go` and unexported. I moved it here and exported it. The `TODO` item was meant to remind me to update `kapacitor/main.go` to use this code instead.

cmd/kapacitor/main.go
```diff
@@ -822,7 +804,7 @@ func doDefineTemplate(args []string) error {
 }
 
 func defineTopicHandlerUsage() {
-	var u = `Usage: kapacitor define-topic-handler <topic id> <handler id> <path to handler spec file>
+	var u = `Usage: kapacitor define-topic-handler <path to handler spec file>
```
desa: Yup.

cmd/kapacitord/main.go
```diff
@@ -89,15 +89,31 @@ func (m *Main) Run(args ...string) error {
...
+			case <-signalCh:
+				m.Logger.Println("I! Signal received, initializing clean shutdown...")
+				go func() {
+					cmd.Close()
+				}()
+				break Loop
+			case <-reloadSignal:
+				m.Logger.Println("I! SIGHUP received, Reloading tasks/templates/directory directory...")
+				if err := cmd.Server.LoadService.Load(); err != nil {
+					// TODO: should this be a sufficient condition to shutdown the server?
```
desa: > What kinds of errors can LoadService.Load return?

Errors are either from reading files or making requests to the API.

> I would prefer the operation to be atomic, meaning if it fails it logs the failure and then reverts to its previous state.
> That is not really possible since you are going through the API instead of calling into the DB services directly.

I think this might be possible, albeit painful. And I'd have to make some assumptions.
I could figure out what tasks, templates, and handlers would result from a new `Load` operation, save their state temporarily, and then, in the case of a failure, rewrite the old values back through the API.

> I am not sure what we should do here. Thoughts?

I'm very much in the same boat.  The three possibilities I see

1. Best effort, don't crash the server, log the errors
2. Crash the server and treat the files like configuration
3. Don't crash the server, restore to a previous state

I'm inclined to lean towards 2 or 3, but I think that it might be possible to do things to minimize the likelihood of an error.

For example, we could add a `dry-run` query param to the API that attempts to do everything that the regular request to the API would, but without actually committing any changes? I think this should mostly be feasible (although possibly not desirable).

This way we could do a dry run to "validate" the files that are currently there. Or something like this. Essentially a best effort to validate that the entire `Load` operation will be successful, without touching the API.

Alternatively, I think the case could be for 2, in that these files should be treated like configuration and bad configuration should stop the runtime.

cmd/kapacitor/main.go
```diff
@@ -687,29 +629,69 @@ func doDefine(args []string) error {
...
 			ID:         id,
 			TemplateID: *dtemplate,
 			Type:       ttype,
 			DBRPs:      ddbrp,
 			TICKscript: script,
 			Vars:       vars,
 			Status:     client.Disabled,
-		})
+		}
+		if *dfile != "" {
```
desa: Not sure I totally follow. Are you saying to do 

```go

		if *dfile != "" {
			o, err := fileVars.CreateTaskOptions()
			if err != nil {
				return err
			}
		_, err = cli.CreateTask(o)
		} else {
		o := client.CreateTaskOptions{
			ID:         id,
			TemplateID: *dtemplate,
			Type:       ttype,
			DBRPs:      ddbrp,
			TICKscript: script,
			Vars:       vars,
			Status:     client.Disabled,
		}
		_, err = cli.CreateTask(o)
            }
```

this instead?

cmd/kapacitord/main.go
```diff
@@ -89,15 +89,31 @@ func (m *Main) Run(args ...string) error {
 
 		signalCh := make(chan os.Signal, 1)
 		signal.Notify(signalCh, os.Interrupt, syscall.SIGTERM)
+		reloadSignal := make(chan os.Signal, 1)
+		signal.Notify(reloadSignal, os.Interrupt, syscall.SIGHUP)
```
desa: I think we'll still need the loop since the `SIGHUP` doesn't necessarily crash the runtime anymore.

server/server.go
```diff
@@ -296,6 +302,8 @@ func (s *Server) appendStorageService() {
 
 	s.StorageService = srv
 	s.AppendService("storage", srv)
+
+	return
```
desa: Accident

services/load/service.go
```diff
@@ -0,0 +1,416 @@
...
+// HandlerFiles gets a slice of all files with the .json, .yml, and
+// .yaml file extentions in the configured handler directory.
+func (s *Service) handlerFiles() ([]string, error) {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	handlers := []string{}
+
+	handlersDir := s.config.handlersDir()
+
+	files, err := ioutil.ReadDir(handlersDir)
```
desa: in `load/config.go` the `Validate` method requires that the directory specified must have subdirectories `tasks`, `templates`, and `handlers`. Should I relax this requirement?

cmd/kapacitord/main.go
```diff
@@ -89,33 +89,42 @@ func (m *Main) Run(args ...string) error {
...
-				if err := cmd.Server.LoadService.Load(); err != nil {
-					m.Logger.Println(fmt.Sprintf("E! Failed to reload tasks/templates/handlers: %s", err))
-
-					if _, ok := err.(load.HardError); ok {
-						go func() {
-							cmd.Close()
-						}()
-						break Loop
+
+			case s := <-signalCh:
```
desa: @nathanielc there has to be a better way than this, but I couldn't find an example. This could be my end-of-day brain not thinking straight. Thoughts?

desa:
https://github.com/influxdata/chronograf/pull/1721 will be able to remove the `stream/batch` button from tickscript UI when this is merged

desa:
@121watts ^^

desa:
Seems to be having an issue with batch tasks. Not exactly sure why. Will investigate.

F4ncyMooN:
Hi, I'm looking for a convergence method for alerts. 
For example, I want to monitor on 100 hosts which are located in 2 different DCs(50 in DC1, 50 in DC2), if 90% of DC1's hosts failed, only DC's alerts will be delivered.
I think this should be implemented by myself in some handlers.
Is it possible if I create some handlers without change the source code of kapacitor(just like udf) after this PR?

desa:
@F4ncyMooN This PR is a little bit different. This PR introduces a way to load tasks and templates from a directory.

cmd/kapacitor/main.go
```diff
@@ -542,72 +541,15 @@ var (
...
 
-type dbrps []client.DBRP
-
-func (d *dbrps) String() string {
-	return fmt.Sprint(*d)
-}
-
-// Parse string of the form "db"."rp" where the quotes are optional but can include escaped quotes
-// within the strings.
-func (d *dbrps) Set(value string) error {
```
nathanielc: Why was this moved to the client package? The `Set` method is part of the `flag.Value` interface, which seems like it belongs here in the main package.

cmd/kapacitord/main.go
```diff
@@ -88,16 +89,44 @@ func (m *Main) Run(args ...string) error {
...
 		// Block until one of the signals above is received
-		select {
-		case <-signalCh:
-			m.Logger.Println("I! Signal received, initializing clean shutdown...")
-			go func() {
-				cmd.Close()
-			}()
+	Loop:
+		for {
+			select {
```
nathanielc: This select statement has only a single case, which means this can be simplified to `for s := range signalCh {...}`.

cmd/kapacitord/main.go
```diff
@@ -88,16 +89,44 @@ func (m *Main) Run(args ...string) error {
...
+						m.Logger.Println(fmt.Sprintf("E! Failed to reload tasks/templates/handlers: %s", err))
+						if _, ok := err.(load.HardError); ok {
+							go func() {
+								cmd.Close()
+							}()
+							break Loop
+						}
+					}
+
+					// This should never happen
```
nathanielc: What about os.Interrupt?

cmd/kapacitord/main.go
nathanielc: Hmm, ok so further down we check for a second signal. Maybe we should just keep a state about the signals we have received so its clear how transitions occur.

cmd/kapacitord/main.go
```diff
@@ -88,16 +89,44 @@ func (m *Main) Run(args ...string) error {
...
+					go func() {
+						cmd.Close()
+					}()
+					break Loop
+
+				case syscall.SIGHUP.String():
+					m.Logger.Println("I! SIGHUP received, Reloading tasks/templates/handlers directory...")
+					if err := cmd.Server.LoadService.Load(); err != nil {
+						m.Logger.Println(fmt.Sprintf("E! Failed to reload tasks/templates/handlers: %s", err))
+						if _, ok := err.(load.HardError); ok {
```
nathanielc: What other kinds of errors are returned? Should we push this logic into the server type? Maybe something in this method https://github.com/influxdata/kapacitor/blob/master/server/server.go#L882. Plus a method on the Server type to reload.

cmd/kapacitord/run/command.go
```diff
@@ -127,6 +127,11 @@ func (cmd *Command) Run(args ...string) error {
 	if err := s.Open(); err != nil {
 		return fmt.Errorf("open server: %s", err)
 	}
+
+	if err := s.LoadService.Load(); err != nil {
```
nathanielc: Why isn't this part of the Open process for the Server?

cmd/kapacitor/main.go
desa: `client.TaskVars` ended up needing the same DBRP logic so rather than having two copies of the type, one in the `main` package and one in the `client` package I figured that I'd move it into client. I'm happy to have two separate types if you think thats best.

cmd/kapacitord/main.go
desa: @nathanielc  what should happen on `os.Interrupt`?

desa: removing comment.

cmd/kapacitord/main.go
desa: Not totally sure I follow. I think this logic might belong here. `Load` can return few different error types, but there's a configuration setting that specifies what should happen to the runtime in the case of an error. If the configuration says crash on error we wrap any errors coming from `Load` the error with a `load.HardError`.

cmd/kapacitord/run/command.go
desa: fixed.

desa:
@nathanielc addressed most of the issues, only thing I was a bit unsure of was https://github.com/influxdata/kapacitor/pull/1481#discussion_r138128327

cmd/kapacitord/main.go
```diff
@@ -87,16 +88,38 @@ func (m *Main) Run(args ...string) error {
...
+			case syscall.SIGTERM.String():
+				m.Diag.Info("SIGTERM received, initializing clean shutdown...")
+				go func() {
+					cmd.Close()
+				}()
+				break Loop
+
+			case syscall.SIGHUP.String():
+				m.Diag.Info("SIGHUP received, reloading tasks/templates/handlers directory...")
+				if err := cmd.Server.LoadService.Load(); err != nil {
```
nathanielc: I was thinking, instead of doing cmd.Server.LoadService.Load() and handing the error types here in the main package, create a method on the server called Reload or something, that if it returns an error at all causes the Server to crash.
Then inside that method determine if the underlying errors returned from LoadService.Load are work crashing. That way the main function crashes if it gets any error, and the Server contains the logic on what errors cause crashes.

Does that help?

desa: yup.

cmd/kapacitor/main.go
nathanielc: Why did the `client.TaskVars` need to implement the `flag.Value` interface?

nathanielc: Maybe the type in the client package should implement a TextUnmarshaler interface and then the type in the main package uses the type from the client package and adds the Set method to its own type? Does that work?

desa: Yeah I like that.

desa: @nathanielc Looks like implementing `TextUnmarshaler` causes issues with the regular JSON marshaling. Does having a method `Unmarshal` on the `client.DBRP` type work?

cmd/kapacitord/main.go
desa: Fixed

desa:
@nathanielc I think I addressed all the previous issues.

nathanielc:
I am testing this out locally, here are some of the issues I ran into:

* The `kapacitor config` command produces a load section with an empty directory, it would be good to apply a default dir like we do for the other dirs. See https://github.com/influxdata/kapacitor/blob/master/server/config.go#L181-L184
* After configuring a directory Kapacitor now crashes on startup because the `<dir>/tasks` dir doesn't exist. Kapacitor should treat a lack of dir as there being no tasks configured, not as a condition to crash the server
* The `hard` config option seems to not do anything different whether its on or off. Specifically I had a tick in the dir with some invalid syntax. NO matter what I could not get kapacitor to start without crashing.
* When defining a handler I followed the example [here](https://github.com/influxdata/kapacitor/blob/dbd0d971a614ee563182e95ebf2b29b7d55d0365/examples/load/README.md#handlers) and I got an error since the handler ID was missing from the file.
* I changed the ID of a handler and the old handler is still around, it was not deleted. I expected the old handler to be removed
* I renamed a task and now both the old and new tasks are defined. I expected the old task to be deleted.
* At first I could not figure out how to create a task from a template. I originally put it in the `tasks` directory since its a task and just refers back to the template. Once I move the yaml file back to the template I got it working. I think tasks should exist in the `tasks` dir even if they refer to a template.
* I expected an error if I forgot to define a `template-id` in a template task that already exists, but it doesn't error. It works, which means the files can drift from the real state since its possible to edit a file that would cause an error if created new, but otherwise not.

Those are the issues I ran into, mostly usability issues and the fact the the README in the examples dir is out of date so it lead me astray. 

Its super nice to be able to define the tasks from dirs like this! Thanks!

desa:
I am testing this out locally, here are some of the issues I ran into:

> The kapacitor config command produces a load section with an empty directory, it would be good to apply a default dir like we do for the other dirs. See ?>https://github.com/influxdata/kapacitor/blob/master/server/config.go#L181-L184

Fixed

> After configuring a directory Kapacitor now crashes on startup because the <dir>/tasks dir doesn't exist. Kapacitor should treat a lack of dir as there being no tasks configured, not as a condition to crash the server

fixed. had a negation

> At first I could not figure out how to create a task from a template. I originally put it in the tasks directory since its a task and just refers back to the template. Once I move the yaml file back to the template I got it working. I think tasks should exist in the tasks dir even if they refer to a template.

fixed

desa:
@nathanielc 

> The hard config option seems to not do anything different whether its on or off. Specifically I had a tick in the dir with some invalid syntax. NO matter what I could not get kapacitor to start without crashing.

This was intentional. Hard errors only apply when re-loading from a directory when the process is running. The idea here was that if you’re starting launching from a directory that the tasks/templates/handlers should be treated as configuration and therefore the server should not start until they are valid. However if you’re running already and you make a change to your tasks/templates/handlers an error with one of those should not crash your server.

> I changed the ID of a handler and the old handler is still around, it was not deleted. I expected the old handler to be removed
I renamed a task and now both the old and new tasks are defined. I expected the old task to be deleted.

How do we do this? My current thought would be to have some kind of `LoadTasks.lock` file that will contain the tasks the were loading from a directory and then diff against that list. That seem reasonable?

> When defining a handler I followed the example here and I got an error since the handler ID was missing from the file.
> Those are the issues I ran into, mostly usability issues and the fact the the README in the examples dir is out of date so it lead me astray.

Which examples were/are problematic? When I use that directory as my load directory everything seems to work fine.

> I expected an error if I forgot to define a template-id in a template task that already exists, but it doesn't error. It works, which means the files can drift from the real state since its possible to edit a file that would cause an error if created new, but otherwise not.

Should this be done by validating the file before even passing it to the API?

nathanielc:
>This was intentional. Hard errors only apply when re-loading when the process is running. The idea here was that if you’re starting launching from a directory that the tasks/templates/handlers should be treated as configuration and therefore the server should not start until they are valid. However if you’re running already and you make a change to your tasks/templates/handlers an error with one of those should not crash your server.


Hmm, I am beginning to think we should just remove the `hard` option and always have the reload log errors and never crash the server. This is how apache's httpd works. This way the processes doesn't assume any ownership of the workflow and lets the operator decided how to manage loading tasks from a dir. We should be however logging errors and incrementing a counter if the reload fails. This way it can be easily monitored.

> How do we do this? My current thought would be to have some kind of LoadTasks.lock file that will contain the tasks the were loading from a directory and then diff against that list. That seem reasonable?

My though was track a list of tasks/templates/handlers that are created from the load service and then when the load service runs, if the file system no longer defines that task/template/handler then delete it. This list should be stored in the kapacitor.db file. It could be an explicit list or simply a new property on the data being stored for each item. Does that make sense?


>Which examples were/are problematic? When I use that directory as my load directory everything seems to work fine.

I'll comment in-line on the issues I had.

> Should this be done by validating the file before even passing it to the API?

No, I'd rather all validation logic live in the API, that way if someone isn't using the client they still get proper validation.

examples/load/README.md
```diff
@@ -0,0 +1,138 @@
...
+```
+[load]
+ enabled = true
+ hard = false
+ dir="/path/to/directory"
+```
+
+`dir` specifies the directory where the definition files exist.
+
+The specified directory should have three subdirectories `tasks`, `templates`, and `handlers`.
```
nathanielc: This line implies that the directories must exist, which is not the case.

examples/load/README.md
```diff
@@ -0,0 +1,138 @@
...
+
+The `tasks` directory will contain will contain task tickscripts.
+
+The `templates` directory will contain both template tickscripts and the associated templated task
+definition files (either yaml or json).
```
nathanielc: These lines need to be updated to reflect the new behavior.

examples/load/README.md
```diff
@@ -0,0 +1,138 @@
...
+
+## tickscript (files with a `.tick` extension)
+
+Tickscripts can be marked up with commented key-value pairs for predefined attributes.
+Those attributes may be any of the following
+
+* `id` - the file name without the tick extension
+* `type` - determined by introspection of the task (stream, batch)
+* `kind` - task, template. defined using a `set statement`
+* `subscriptions` - defined using the `subscribe` keyword followed by a specified subscription
```
nathanielc: Isn't the syntax `dbrp` now?

examples/load/README.md
```diff
@@ -0,0 +1,138 @@
...
+  "measurement": {"type" : "string", "value" : "cpu" },
+  "where_filter": {"type": "lambda", "value": "\"cpu\" == 'cpu-total'"},
+  "groups": {"type": "list", "value": [{"type":"string", "value":"host"},{"type":"string", "value":"dc"}]},
+  "field": {"type" : "string", "value" : "usage_idle" },
+  "warn": {"type" : "lambda", "value" : "\"mean\" < 30.0" },
+  "crit": {"type" : "lambda", "value" : "\"mean\" < 10.0" },
+  "window": {"type" : "duration", "value" : "1m" },
+  "slack_channel": {"type" : "string", "value" : "#alerts_testing" }
+}
+```
```
nathanielc: This is technically correct yaml but quite non-standard a confusing, this reads more like yaml

```yaml
dbrp:
   - telegraf.autogen
   - telegraf.not_autogen
 template: base_template
 vars: 
   measurement:
    type: string
    value: cpu
   where_filter:
    type: lambda
    value: "\"cpu\" == 'cpu-total'"
   groups:
    type: list
    value:
        - type: string
          value: host
        - type: string
          value: dc
   field:
    type: string
    value : usage_idle
   warn:
    type: lambda
    value: "\"mean\" < 30.0"
   crit:
    type: lambda
    value: "\"mean\" < 10.0"
   window:
    type: duration
    value : 1m
   slack_channel:
    type: string
    value: "#alerts_testing"
```

examples/load/README.md
```diff
@@ -0,0 +1,138 @@
...
+
+```
+topic: cpu
+kind: slack
+match: changed() == TRUE
+options:
+  channel: '#alerts'
+```
+
+the name of the file will be used as the handler id.
```
nathanielc: This is false, the id is read from the file and the filename is not used.

desa:
@nathanielc removed the hard error condition and added the logic for deleting stale tasks/templates/handlers that were loaded from a directory. The todo items I have left are

* Add tests for DAO
* Make edits to README
* Cleanup remove logic a little

desa:
@nathanielc added tests. Should be ready for review

examples/load/README.md
```diff
@@ -0,0 +1,155 @@
...
+
+```
+[load]
+ enabled = true
+ dir="/path/to/directory"
+```
+
+`dir` specifies the directory where the definition files exist.
+
+The specified directory will check to see if any of three subdirectories `tasks`, `templates`, and `handlers` exists.
```
nathanielc: The configured `dir` may contain any of these sub-directories: `tasks`, `templates`, and `handlers`.

examples/load/README.md
```diff
@@ -0,0 +1,155 @@
...
+
+The `tasks` directory will contain will contain task tickscripts and and the associated templated task
+definition files (either yaml or json).
+
+The `templates` directory will contain both template tickscripts
```
nathanielc: The `templates` directory will contain template tickscripts

examples/load/README.md
```diff
@@ -0,0 +1,155 @@
...
+
+The `handlers` directory will contain will contain all topic handler definitions in yaml or json.
```
nathanielc: The `handlers` directory will contain will contain topic handler definitions in yaml or json.

examples/load/README.md
```diff
@@ -0,0 +1,155 @@
...
+
+* `id` - the file name without the tick extension
+* `type` - determined by introspection of the task (stream, batch)
+* `dbrp` - defined using the `dbrp` keyword followed by a specified database and retention policy
+
+For example
+```
+dbrp "telegraf"."autogen"
+
+var measurement string
```
nathanielc: This example is actually a template so it can't be placed in the tasks dir.

examples/load/README.md
```diff
@@ -0,0 +1,155 @@
...
+* `id` - filename without the `yaml` or `yml` extension
+* `dbrp` - required
+* `template-id` - required
+* `vars` - list of template vars
+
+```yaml
+dbrp:
+  - telegraf.autogen
+  - telegraf.not_autogen
+template: base_template
```
nathanielc: This should be `templated-id`

examples/load/README.md
```diff
@@ -0,0 +1,155 @@
...
+    |alert()
+        .warn(lambda: "usage_idle" < 20)
+        .crit(lambda: "usage_idle" < 10)
+        // Send alerts to the `cpu` topic
+        .topic('cpu')
+```
+
+### Templated Tasks
+
+Template variables may be added as either json or yaml.
```
nathanielc: I think the README should have a simple TICKscript above to demonstrate creating tasks.

Then in this Template section it should first define a TICKscript template. Then show how that template can be referenced from a file in the `tasks` directory.

We want to make it clear that anything in the `tasks` dir will create a task, and anything in the `templates` dir will create a template.

server/server.go
```diff
@@ -896,7 +896,7 @@ func (s *Server) Open() error {
 	}
 
 	if err := s.LoadService.Load(); err != nil {
-		return fmt.Errorf("loading tasks/templates/handlers: %s", err)
```
nathanielc: I think its fine for the server to refuse to start if it has bad configuration, but we do not want it to crash from a simple typo.

server/server.go
```diff
@@ -1060,7 +1060,7 @@ func (s *Server) writeID(file string, id uuid.UUID) error {
 func (s *Server) Reload() error {
```
nathanielc: If this function can't error anymore we should remove the `error` return value and simplify the caller to not need to handle the error.

examples/load/README.md
desa: fixed

examples/load/README.md
desa: fixed.

examples/load/README.md
desa: fixed.

server/server.go
desa: fixed.

server/server.go
desa: not sure exactly what to do with this. Should I keep the error here and do some checking of the type of error, or just leave it as is?

nathanielc: What types of errors would you check for? Seems like if it errors at all it should return the error.

examples/load/README.md
desa: Added a section on defining a template before and the templated tasks now use that template. Think that  should be sufficient? Or did you have something else in mind?

services/load/service.go
```diff
@@ -0,0 +1,666 @@
...
+			continue
+		}
+		diffs = append(diffs, x)
+	}
+
+	return diffs
+}
+
+func (s *Service) loadedTasks() ([]string, error) {
+	items, err := s.items.List("tasks")
```
nathanielc: We should create constants for the `tasks` , `templates` and `handlers` strings, since they are part of the key for the DOA. A typo would corrupt the data :(.

services/load/service.go
```diff
@@ -0,0 +1,666 @@
...
+		Store(namespace string) storage.Interface
+		Register(name string, store storage.StoreActioner)
+	}
+
+	diag Diagnostic
+}
+
+func NewService(c Config, h http.Handler, d Diagnostic) (*Service, error) {
+	cfg := client.Config{
+		URL: defaultURL,
```
nathanielc: Can we add a unique user agent here so its clear the consumer of the API is this internal client from Load?

desa: fixed

services/load/service.go
desa: fixed.

desa:
@nathanielc  I think I got the completion right, but I'm not totally sure. I sometimes have issues with it on OSX.

Does this seem right?

https://github.com/influxdata/kapacitor/pull/1481/commits/396ab78e847b94949c973e8655161fd9d5b96854

desa:
@nathanielc just rebased. I think things should be good to go now.

PR #1500, opened 2017-07-27
Update the UDF socket server to cleanly stop.

nathanielc:
Now its possible to use defer to Stop the server safely.

Also add two fields to the initialize response for the task ID and node ID of the UDF. This way the Kapacitor context can be passed down to the UDF node for correct reporting.

- [x] Rebased/mergable
- [x] Tests pass
- [x] CHANGELOG.md updated


services/udf/service.go
```diff
@@ -109,7 +111,7 @@ func (s *Service) Refresh(name string) error {
 }
 
 func (s *Service) loadUDFInfo(name string) (udf.Info, error) {
-	u, err := s.Create(name, s.logger, nil)
+	u, err := s.Create(name, "", "", s.logger, nil)
```
desa: why are these nil?

desa: scratch that. I see.

nathanielc: Generally because the UDF is not being created in the context of a task or node.

More specificially because loadUDFInfo only makes the `Info` request and never sends the `Init` request so the taskID and nodeID will never be sent or subsequently used.

I should probably add a comment to that fact since its not clear.

task_master.go
```diff
@@ -50,7 +50,7 @@ type LogService interface {
 type UDFService interface {
 	List() []string
 	Info(name string) (udf.Info, bool)
-	Create(name string, l *log.Logger, abortCallback func()) (udf.Interface, error)
+	Create(name, taskID, nodeID string, l *log.Logger, abortCallback func()) (udf.Interface, error)
```
desa: What's the motivation for exposing the task and node IDs? Is it just adding some additional nice to have information? Or something else?

udf/agent/server.go
```diff
@@ -32,59 +31,88 @@ func NewServer(l net.Listener, a Accepter) *Server {
...
+// Stop closes the listener and stops all server activity.
 func (s *Server) Stop() {
 	s.mu.Lock()
-	defer s.mu.Unlock()
 	if s.stopped {
+		s.mu.Unlock()
 		return
 	}
 	s.stopped = true
+	s.listener.Close()
```
desa: Is ignoring this error okay?

desa: I think it is.

task_master.go
nathanielc: The motivation came from developing Morgoth more where I realized that as a UDF the process gets lots of connections from Kapacitor but it doesn't know the context of the connection. This way when Morgoth get a connection and is told to initialize it knows under which context it is operating, thus providing the user of Morgoth with more context in its logs and exposed metrics.

In short, it pushes down context so that it is not lost.

udf/agent/server.go
```diff
@@ -32,59 +31,88 @@ func NewServer(l net.Listener, a Accepter) *Server {
...
 		}
 	}()
 }
 
 func (s *Server) run() error {
+	conns := make(chan net.Conn)
 	errC := make(chan error, 1)
+	s.wg.Add(1)
 	go func() {
+		s.wg.Done()
```
desa: Why the immediate `Done()`? I havent' seen this pattern before

udf/agent/server.go
nathanielc: Yes, although I don't know why :) It was being ignored previously. And is typically closed via defer, meaning its error is typically ignored.

udf/agent/server.go
nathanielc: That's a bug :)

PR #1508, opened 2017-08-01
add feature to detect string edge 

rezamirzaei:
- [ ] Rebased/mergable
- [ ] Tests pass
- [ ] CHANGELOG.md updated

### DIFF
--- a/derivative.go
+++ b/derivative.go
@@ -156,11 +156,12 @@ func numToFloat(num interface{}) (float64, bool) {
 		return 0, false
 	}
 }
+
 func diffFunc(f0,f1 interface{})(float64, bool){
-	n0 , ok := numToFloat(f0);
-	n1 , ok1 := numToFloat(f1);
+	n0 , ok := numToFloat(f0)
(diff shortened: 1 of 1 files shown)


nathanielc:
@rezamirzaei I don't think this approach is going to work well. Adding string support to the derivative node makes little sense. We could add a an edgeDetect node that supports strings and is well defined for how that works. Adding it to the derivative node will be confusing to users and will not be very discoverable as a feature. I can help provide pointers on how to add an `edgeDetect` node if you want.

Closing.

rezamirzaei:
@nathanielc Thank you for pay attention. i agree with you. i'm glad if you help me in the way you say.
Thank you

nathanielc:
@rezamirzaei Have a look at the StateCount node https://docs.influxdata.com/kapacitor/v1.3/nodes/state_count_node/

You might be able to use the StateCount node directly if your edge is in a binary state. If not then create a copy of the StateCount node and modify it to track the last state and output values when an edge is found.

rezamirzaei:
@nathanielc Thank you for your response.

PR #1512, opened 2017-08-04
Use details field from alert node in pagerduty

### LINKED ISSUE #743: Pagerduty alert handler should allow customizing details
Right now the pagerduty handler is hard-coded to send the alert data in the details field. This data is not very useful as the way it is formatted is very hard to read (it's hard to match a value in `.values` with a name in `.columns`, and the pagerduty details field is meant to be human readable). Instead the details field should be able to be customized. Some of the other handlers use the `.details` attribute of the alert, though the documentation seems to emphasize that `.details` is meant to be html, where as pagerduty's "details" field is arbitrary data. So I guess the options are to either use this field ignoring the html bit, or provide a new node key that can be used.

While we're at it, pagerduty has another field, `contexts`, that would be useful to be able to set as well.

desa:
Fixes #743 

Previously the details field specified on an alert node was not being used in the pagerduty service. Instead, the results object that generated was being used. This PR now uses the details field specified.

@nathanielc is this all, or is there something I'm missing?

- [ ] Rebased/mergable
- [ ] Tests pass
- [ ] CHANGELOG.md updated

### DIFF
--- a/integrations/streamer_test.go
+++ b/integrations/streamer_test.go
@@ -7828,6 +7828,8 @@ func TestStream_AlertPagerDuty(t *testing.T) {
 	ts := pagerdutytest.NewServer()
 	defer ts.Close()
 
+	defaultDetails := "{&#34;Name&#34;:&#34;cpu&#34;,&#34;TaskName&#34;:&#34;TestStream_Alert&#34;,&#34;Group&#34;:&#34;host=serverA&#34;,&#34;Tags&#34;:{&#34;host&#34;:&#34;serverA&#34;},&#34;ServerInfo&#34;:{&#34;Hostname&#34;:&#34;&#34;,&#34;ClusterID&#34;:&#34;&#34;,&#34;ServerID&#34;:&#34;&#34;},&#34;ID&#34;:&#34;kapacitor/cpu/serverA&#34;,&#34;Fields&#34;:{&#34;count&#34;:10},&#34;Level&#34;:&#34;CRITICAL&#34;,&#34;Time&#34;:&#34;1971-01-01T00:00:10Z&#34;,&#34;Message&#34;:&#34;CRITICAL alert for kapacitor/cpu/serverA&#34;}\n"
+
 	var script = `
 stream
 	|from()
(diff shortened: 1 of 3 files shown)


desa:
@phemmer thoughts?

phemmer:
Seems straightforward enough. Lets `.details()` behave like all the other alert services 👍

integrations/streamer_test.go
```diff
@@ -7844,6 +7844,7 @@ stream
 		.info(lambda: "count" > 6.0)
 		.warn(lambda: "count" > 7.0)
 		.crit(lambda: "count" > 8.0)
+		.details('details')
```
nathanielc: Why change the test here? I'd like to test that the default behavior is not changed. Meaning if details is not specified that the pageDuty handler remains the same.

integrations/streamer_test.go
```diff
@@ -7828,6 +7828,8 @@ func TestStream_AlertPagerDuty(t *testing.T) {
 	ts := pagerdutytest.NewServer()
 	defer ts.Close()
 
+	defaultDetails := "{&#34;Name&#34;:&#34;cpu&#34;,&#34;TaskName&#34;:&#34;TestStream_Alert&#34;,&#34;Group&#34;:&#34;host=serverA&#34;,&#34;Tags&#34;:{&#34;host&#34;:&#34;serverA&#34;},&#34;ServerInfo&#34;:{&#34;Hostname&#34;:&#34;&#34;,&#34;ClusterID&#34;:&#34;&#34;,&#34;ServerID&#34;:&#34;&#34;},&#34;ID&#34;:&#34;kapacitor/cpu/serverA&#34;,&#34;Fields&#34;:{&#34;count&#34;:10},&#34;Level&#34;:&#34;CRITICAL&#34;,&#34;Time&#34;:&#34;1971-01-01T00:00:10Z&#34;,&#34;Message&#34;:&#34;CRITICAL alert for kapacitor/cpu/serverA&#34;}\n"
```
desa: This is the format that gets generated.

desa:
@nathanielc this should be ready to review again.

desa:
kk

desa:
Whoops. Just realized i didn't update the changelog. Going to open another PR for that.

PR #1518, opened 2017-08-09
Add Database proprety to QueryNode

goakley:
The Database property is added to the InfluxDB API call's URL query
string, removing the need to specify the database in the query's
SELECT statement.

- [ ] Rebased/mergable
- [ ] Tests pass
- [ ] CHANGELOG.md updated


nathanielc:
@goakley The database from the query is also used in the permission check of valid dbrps for a task. When you specify a task dbrp with the `-dbrp` flag the databases and retention policies of the query must also match the list the task has. Can you update this PR to also account for checking the dbrp list?

goakley:
@nathanielc I believe I understand.

A slight problem with how retention policies work in this situation is with the client CLI, namely, that it crashes when trying to specify the default retention policy:
```
$ kapacitor define test -type batch -tick tick.tick -dbrp "database".""
```
It seems like this is a bug, as the InfluxDB documentation states that retention policy is optional when writing a query of the form: `( db_name "." [ policy_name ] "." measurement_name )`

nathanielc:
@goakley Kapacitor requires the retention policy as it doesn't know what the default is.

goakley:
@nathanielc why does Kapacitor need to know the default retention policy?  It seems like the workflow here works as intended using an empty/default retention policy.

PR #1535, opened 2017-08-24
WIP: Second attempt at logging refactor

desa:


derivative.go
```diff
@@ -125,7 +126,7 @@ func (n *DerivativeNode) derivative(prev, curr models.Fields, prevTime, currTime
 	f1, ok := numToFloat(curr[n.d.Field])
 	if !ok {
 		n.incrementErrorCount()
-		n.logger.Printf("E! cannot apply derivative to type %T", curr[n.d.Field])
+		n.diag.Error("cannot perform derivative", fmt.Errorf("field %s is the wrong %T", n.d.Field, curr[n.d.Field]))
```
nathanielc: I think this one should read:

```
n.diag.Error("cannot perform derivative", errors.New("field is the wrong type") keyvalue.KV("field",n.d.Field), keyvalue.KV("type", reflect.TypeOf(curr[n.d.Field]).String()))
```

derivative.go
```diff
@@ -140,7 +141,7 @@ func (n *DerivativeNode) derivative(prev, curr models.Fields, prevTime, currTime
 	elapsed := float64(currTime.Sub(prevTime))
 	if elapsed == 0 {
 		n.incrementErrorCount()
-		n.logger.Printf("E! cannot perform derivative elapsed time was 0")
+		n.diag.Error("cannot perform derivative", errors.New("elaspsed time was 0"))
```
nathanielc: Can the n.diag.Error() method call n.incrementErrorCount() internally? That would remove the onus on each node to correctly count their errors and rather only need to report them to the Diagnostic interface. This probably means you need a small shim implementation of the interface within the kapacitor package to intercept Error and With* calls.

Does that sound reasonable and useful?

log.go
```diff
@@ -63,24 +62,14 @@ func (n *LogNode) EndBatch(end edge.EndBatchMessage) (edge.Message, error) {
...
 
 func (n *LogNode) BufferedBatch(batch edge.BufferedBatchMessage) (edge.Message, error) {
-	n.buf.Reset()
-	if err := n.enc.Encode(batch); err != nil {
-		n.incrementErrorCount()
-		n.logger.Println("E!", err)
-		return batch, nil
-	}
-	n.logger.Println(n.key, n.buf.String())
+	// TODO: fix prefix and other loger here
```
nathanielc: Is this TODO and others in the file complete?

node.go
```diff
@@ -132,7 +153,8 @@ func (n *node) start(snapshot []byte) {
 					err = fmt.Errorf("%s: Trace:%s", r, string(trace[:n]))
 				}
 				n.abortParentEdges()
-				n.logger.Println("E!", err)
+				n.diag.Error("encountered panic", err)
```
nathanielc: Its not always true that a panic was encountered here. This error message should be something like `node failed`

server/config.go
```diff
@@ -61,12 +61,13 @@ import (
...
-	Task           task_store.Config `toml:"task"`
-	InfluxDB       []influxdb.Config `toml:"influxdb" override:"influxdb,element-key=name"`
-	Logging        logging.Config    `toml:"logging"`
+	HTTP     httpd.Config      `toml:"http"`
+	Replay   replay.Config     `toml:"replay"`
+	Storage  storage.Config    `toml:"storage"`
+	Task     task_store.Config `toml:"task"`
+	InfluxDB []influxdb.Config `toml:"influxdb" override:"influxdb,element-key=name"`
+	// TODO(desa): this should change to diagnostic eventually
+	Logging        diagnostic.Config `toml:"logging"`
```
nathanielc: This config is about logging, but logging is implemented through diagnostics so I think this name is fine.

services/alert/service.go
```diff
@@ -717,14 +729,19 @@ func decodeStringToTextUnmarshaler(f, t reflect.Type, data interface{}) (interfa
 func (s *Service) createHandlerFromSpec(spec HandlerSpec) (handler, error) {
 	var h alert.Handler
 	var err error
+	ctx := []keyvalue.T{
+		keyvalue.KV("handler", spec.ID),
+		keyvalue.KV("topic", spec.Topic),
+	}
+	handlerDiag := s.diag.WithHandlerContext(ctx...)
```
nathanielc: While redundant should we move this line down the only the cases that use handlerDiag so its not created when its not needed?

services/diagnostic/handlers.go
```diff
@@ -0,0 +1,1222 @@
...
+	}
+
+	if len(ctx) == 2 {
+		x := ctx[0]
+		y := ctx[1]
+		l.Error(msg, zap.Error(err), zap.String(x.Key, x.Value), zap.String(y.Key, y.Value))
+		return
+	}
+
+	// This isn't great wrt to allocation, but should not ever actually occur
```
nathanielc: It can occur if the context has more than two elements, which may be uncommon but not impossible correct?

services/diagnostic/handlers.go
```diff
@@ -0,0 +1,1222 @@
...
+	l.Error(msg, fields...)
+}
+
+// Alert Service Handler
+
+type AlertServiceHandler struct {
+	l *zap.Logger
+}
+
+func zapFieldsFromContext(ctx []keyvalue.T) []zapcore.Field {
```
nathanielc: Why not use this above on lines 62-67?

services/diagnostic/handlers.go
```diff
@@ -0,0 +1,1222 @@
...
+	fields := []zapcore.Field{}
```
nathanielc: We know the length of ctx, why not pre allocate the exact size?

services/diagnostic/handlers.go
```diff
@@ -0,0 +1,1222 @@
...
+func (h *AlertServiceHandler) CreatingNewHandlers(length int) {
+	h.l.Debug("creating new handlers in place of old handlers", zap.Int("handler_row_count", length))
+}
+
+func (h *AlertServiceHandler) FoundNewHandler(key string) {
+	h.l.Debug("found new handler skipping", zap.String("handler", key))
+}
+
+func (h *AlertServiceHandler) Error(msg string, err error, ctx ...keyvalue.T) {
+	if len(ctx) == 0 {
```
nathanielc: Is this duplicate from the above Error function?

services/diagnostic/handlers.go
```diff
@@ -0,0 +1,1222 @@
...
+		zap.String("group", string(point.GroupID())),
+		zap.Strings("dimensions", point.Dimensions().TagNames),
+		zap.Strings("tags", TagPairs(point.Tags())),
+		zap.Strings("fields", FieldPairs(point.Fields())),
+		zap.Time("time", point.Time()),
+	)
+}
+
+// TODO: deal with prefix and level
+func (h *KapacitorHandler) LogBatchData(level, prefix string, batch edge.BufferedBatchMessage) {
```
nathanielc: I think prefix can just be a String field. The meaning prefix will be wrong at that point, but the intent is to make the logs unique which a field will do fine.

services/diagnostic/handlers.go
```diff
@@ -0,0 +1,1222 @@
...
+
+	h.l.Debug(msg, fields...)
+}
+
+type ReplayHandler struct {
+	l *zap.Logger
+}
+
+func (h *ReplayHandler) Error(msg string, err error, ctx ...keyvalue.T) {
+	if len(ctx) == 0 {
```
nathanielc: Again repeated code...

services/diagnostic/handlers.go
```diff
@@ -0,0 +1,1222 @@
...
+}
+
+// Edge Handler
+
+type EdgeHandler struct {
+	l *zap.Logger
+}
+
+func (h *EdgeHandler) Collect(mtype edge.MessageType) {
+	h.l.Debug("collected message", zap.Stringer("message_type", mtype))
```
nathanielc: Does this mean we are logging every message?

services/diagnostic/handlers.go
```diff
@@ -0,0 +1,1222 @@
...
+}
+func (h *EdgeHandler) Emit(mtype edge.MessageType) {
+	h.l.Debug("emitted message", zap.Stringer("message_type", mtype))
+}
+
+type LogLevel int
```
nathanielc: Does these need to be exported types?

services/diagnostic/handlers.go
```diff
@@ -0,0 +1,1222 @@
...
+	level LogLevel
+}
+
+func (h *StaticLevelHandler) Write(buf []byte) (int, error) {
+	switch h.level {
+	case LLDebug:
+		h.l.Debug(string(buf))
+	case LLError:
+		h.l.Error(string(buf))
+	case LLFatal:
```
nathanielc: Why do we need Fatal?

derivative.go
desa: Yeah, I thought about doing something like that, but I wasn't sure exactly the best way to do that.

desa:
> How much more work is it to just consume the keyvalue.T type and directly write data in logfmt at this point? We could make the package internal to the diagnostic service so its clear its not general purpose.

Don't think it would be too much work, but I could be off. Currently the `keyvalue.T`s are both strings. Would have to do a bit of work to make them a little more general.

services/diagnostic/handlers.go
desa: yeah, but I think all actual occurrences are less than two.  I'll clarify the comment to reflect that.

services/diagnostic/handlers.go
desa: The first entry of the slice  on line 63 is `fields[0] = zap.Error(err)` is an error. This method expects that they all be strings.

services/diagnostic/handlers.go
desa: fixed

services/diagnostic/handlers.go
desa: yup. don't know how I missed this.

services/diagnostic/handlers.go
desa: yup. don't know how I missed this.

services/diagnostic/handlers.go
desa: I think this comes from https://github.com/influxdata/kapacitor/blob/md-log-v2/edge/log.go#L24-L35 and I don't think the `LogEdge` is ever used. At least I don't think that it is.

services/diagnostic/handlers.go
desa: nope. No reason they need to be.

services/diagnostic/handlers.go
desa: Nope.

node.go
desa: fixed

server/config.go
desa: fixed

services/alert/service.go
desa: fixed

services/diagnostic/handlers.go
desa: fixed now.

services/diagnostic/handlers.go
desa: @nathanielc any idea what the field should be named?

nathanielc:
@desa If the types of keyvalue.T are strings doesn't that work well for logging, since everything will be encoded as text to be able to log it?

derivative.go
desa: fixed

services/diagnostic/handlers.go
desa: fixed

derivative.go
desa:
```go
reflect.TypeOf(curr[n.d.Field]).String()
``` 
panics when `curr[n.d.Field]` is nil
https://play.golang.org/p/gjlr3kuusL

Went with
```go
keyvalue.KV("type", fmt.Sprintf("%T", curr[n.d.Field]))
```
instead

desa:
@nathanielc I think I've addressed all of the issues from the review so far.

desa:
@nathanielc this should be in a ready for another review.

services/diagnostic/handlers.go
```diff
@@ -0,0 +1,1190 @@
...
+	}
+
+	return ts
+}
+
+func FieldPairs(tags models.Fields) []klog.Field {
+	ts := []klog.Field{}
+	for k, v := range tags {
+		var el klog.Field
+		switch v.(type) {
```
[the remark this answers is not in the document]
desa: I didn't know that 👍 .
Fixed.

services/diagnostic/handlers.go
```diff
@@ -0,0 +1,1190 @@
...
+		case int:
+			el = klog.Int(k, v.(int))
```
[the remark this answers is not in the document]
desa: Fixed.

services/diagnostic/handlers.go
```diff
@@ -0,0 +1,1190 @@
...
+		case int64:
+			el = klog.Int64(k, v.(int64))
+		case string:
+			el = klog.String(k, v.(string))
+		case float64:
+			el = klog.Float64(k, v.(float64))
+		case bool:
+			el = klog.Bool(k, v.(bool))
+		default:
+			el = klog.String(k, fmt.Sprintf("%v", v))
```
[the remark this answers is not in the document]
desa: I think I had already needed the other types elsewhere and I was trying to be somewhat performance conscious. In the past whenever I've used `fmt.Sprintf` a ton I've had a very sad CPU.

desa:
@nathanielc I think it's ready to merge on your signal.

services/httpd/logger.go
```diff
@@ -92,25 +90,90 @@ func buildLogLine(l *responseLogger, r *http.Request, start time.Time) string {
...
+	//	r.Proto + `"`,
+	//	detect(strconv.Itoa(l.Status()), "-"),
+	//	strconv.Itoa(l.Size()),
+	//	`"` + detect(referer, "-") + `"`,
+	//	`"` + detect(userAgent, "-") + `"`,
+	//	r.Header.Get("Request-Id"),
+	//	// response time, report in microseconds because this is consistent
+	//	// with apache's %D parameter in mod_log_config
+	//	strconv.FormatInt(time.Since(start).Nanoseconds()/1000, 10),
+	//}
```
nathanielc: Can you remove this dead code?

services/httpd/logger.go
```diff
@@ -92,25 +90,90 @@ func buildLogLine(l *responseLogger, r *http.Request, start time.Time) string {
...
```
nathanielc: This too.

services/influxdb/service.go
```diff
@@ -1225,15 +1232,15 @@ func (c *influxdbCluster) startUDPListener(se subEntry, port string) (*net.UDPAd
 	conf.Buffer = c.udpBuffer
 	conf.ReadBuffer = c.udpReadBuffer
 
-	l := c.LogService.NewLogger(fmt.Sprintf("[udp:%s.%s] ", se.db, se.rp), log.LstdFlags)
-	service := udp.NewService(conf, l)
+	d := c.diag.WithUDPContext(fmt.Sprintf("udp:%s.%s", se.db, se.rp))
```
nathanielc: Why not have the WithUDPContext accept two string arguments for db and rp? Then the context is correctly tagged instead of having a composite string.

services/influxdb/service.go
```diff
@@ -1225,15 +1232,15 @@ func (c *influxdbCluster) startUDPListener(se subEntry, port string) (*net.UDPAd
...
+	service := udp.NewService(conf, d)
 	service.PointsWriter = c.PointsWriter
 	err := service.Open()
 	if err != nil {
 		return nil, err
 	}
 	c.services[se] = service
-	c.logger.Println("I! started UDP listener for", se.db, se.rp)
+	c.diag.StartedUDPListener(fmt.Sprintf("%s.%s", se.db, se.rp))
```
nathanielc: Again here, why not pass the db and rp strings to the method directly?

task.go
```diff
@@ -551,7 +554,8 @@ func (et *ExecutingTask) runSnapshotter() {
 		case <-ticker.C:
 			snapshot, err := et.Snapshot()
 			if err != nil {
-				et.logger.Println("E! failed to snapshot task", et.Task.ID, err)
+				// TODO(desa): I think this might already have the task id
```
nathanielc: What is the result of this TODO?

task.go
```diff
@@ -562,7 +566,8 @@ func (et *ExecutingTask) runSnapshotter() {
 			if size > 0 {
 				err = et.tm.TaskStore.SaveSnapshot(et.Task.ID, snapshot)
 				if err != nil {
-					et.logger.Println("E! failed to save task snapshot", et.Task.ID, err)
+					// TODO(desa): I think this might already have the task id
```
nathanielc: This TODO?

udf.go
```diff
@@ -266,7 +265,10 @@ func (p *UDFProcess) logStdErr() {
 	defer p.logStdErrGroup.Done()
 	scanner := bufio.NewScanner(p.stderr)
 	for scanner.Scan() {
-		p.logger.Println("I!P", scanner.Text())
+		// TODO(desa): fix the issue here. issue comes from loadUDFInfo in services/udf/service.go
```
nathanielc: Is this TODO complete?

services/udf/service.go
```diff
@@ -106,15 +109,18 @@ func (s *Service) Refresh(name string) error {
...
 
 func (s *Service) loadUDFInfo(name string) (udf.Info, error) {
 	// loadUDFInfo creates a UDF connection outside the context of a task or node
 	// because it only makes the Info request and never makes an Init request.
 	// As such it does not need to provide actual task and node IDs.
-	u, err := s.Create(name, "", "", s.logger, nil)
+
+	// TODO(desa): that okay? Nil check against this happens in udf.go logStdErr()
+	// alternative would be to have a mock implementation that gets passed in here
+	u, err := s.Create(name, "", "", nil, nil)
```
nathanielc: I don't think this is okay, its possible valuable information is lost if we do not provide a diagnostic here. Looking at it the simplest approach would be to add a WithUDFDiagnotic method to the Diagnostic interface within this package. The udf.Diagnostic interface is quite simple so its shouldn't be hard to accommodate. Thoughts?

services/httpd/logger.go
desa: fixed.

services/httpd/logger.go
desa: fixed.

services/influxdb/service.go
desa: good idea.

services/influxdb/service.go
desa: fixed.

task.go
desa: it's set.

task.go
desa: it's set.

udf.go
desa: fixed.

Issue #1549, opened 2017-09-07
[Feature Request] Add stateless function now() to get the current local time


mbresson:
Hi,
it would be nice to have a function now() returning the current local time, as it could be used in conjunction with [unixNano()](https://github.com/influxdata/kapacitor/pull/1390) to create alert conditions by comparing a timestamp stored in any field with the current time (as discussed in [this discussion thread](https://community.influxdata.com/t/solved-compare-a-date-field-with-now-in-kapacitor/1746)).

For instance, it could be used like so:

```
batch
  |query('''SELECT * FROM "mydb"."autogen"."subscription"''')
    .period(10h)
    .every(10s)
  |alert()
    .crit(lambda: "expiration" < unixNano(now()))
    .log('/tmp/dates-alerts.log')
```

PR #1550, opened 2017-09-07
Add now built in function

mbresson:
This pull request adds a stateless function now() that could be used in conjunction with the recently merged function unixNano() to compare fields containing time encoded in nanoseconds with the current time.
Relevant feature request: [Add stateless function now() to get the current local time](https://github.com/influxdata/kapacitor/issues/1549)

- [X] Rebased/mergable
- [X] Tests pass
- [X] CHANGELOG.md updated

(pull request beginner here, please let me know if I've made a mistake in the process, thanks)

### DIFF
--- a/integrations/streamer_test.go
+++ b/integrations/streamer_test.go
@@ -9037,6 +9037,83 @@ stream
 	}
 }
 
+func TestStream_LambdaNow(t *testing.T) {
+	var script = `
+stream
+	|from()
+		.measurement('account')
+	|where(lambda: "expiration" < unixNano(now()))
+	|groupBy('owner')
+	|httpOut('TestStream_LambdaNow')
+`
+
+	expectedOutput := models.Result{
+		Series: models.Rows{
+			{
+				Name:    "account",
+				Tags:    map[string]string{"owner": "ownerA"},
+				Columns: []string{"time", "expiration"},
+				Values: [][]interface{}{
+					{
+						time.Date(1971, 1, 1, 0, 0, 0, 0, time.UTC),
+						float64(3.15533e+17), // 1980-01-01 00:00:00 (Unix ns timestamp)
+
+						// we expect "expiration" to be float64 and not int64 (even with input data consisting of ints)
+						// because httpOut uses JSON as serialization format for the results,
+						// resulting in the integers becoming floats
+					},
+				},
+			},
+
+			// the point with expiration = 4102440000000000000 should not be in the results
+			// as it represents a date past now (2100-01-01 00:00:00)
+
+			{
+				Name:    "account",
+				Tags:    map[string]string{"owner": "ownerC"},
+				Columns: []string{"time", "expiration"},
+				Values: [][]interface{}{
+					{
+						time.Date(1971, 1, 1, 0, 0, 0, 0, time.UTC),
+						float64(6.56419e+17), // 1990-10-20 10:42:42
+					},
+				},
+			},
+		},
+	}
+
+	testStreamerWithOutput(t, "TestStream_LambdaNow", script, time.Second, expectedOutput, false, nil)
+}
+
+func TestStream_EvalNow(t *testing.T) {
+	var script = `
+stream
+	|from()
+		.measurement('account')
+	|eval(lambda: year(now()))
+		.as('currentYear')
+	|httpOut('TestStream_EvalNow')
+`
+
+	expectedOutput := models.Result{
+		Series: models.Rows{
+			{
+				Name:    "account",
+				Tags:    map[string]string{"owner": "ownerA"},
+				Columns: []string{"time", "currentYear"},
+				Values: [][]interface{}{
+					{
+						time.Date(1971, 1, 1, 0, 0, 0, 0, time.UTC),
+						float64(time.Now().Year()),
+					},
+				},
+			},
+		},
+	}
+
+	testStreamerWithOutput(t, "TestStream_EvalNow", script, time.Second, expectedOutput, false, nil)
+}
+
 func TestStream_Autoscale(t *testing.T) {
 	testCases := map[string]struct {
 		script           string
--- a/tick/stateful/functions.go
+++ b/tick/stateful/functions.go
@@ -225,6 +225,7 @@ func init() {
 	statelessFuncs["day"] = day{}
 	statelessFuncs["month"] = month{}
 	statelessFuncs["year"] = year{}
+	statelessFuncs["now"] = now{}
 
 	// Humanize functions
 	statelessFuncs["humanBytes"] = humanBytes{}
@@ -1352,6 +1353,34 @@ func (year) Call(args ...interface{}) (v interface{}, err error) {
 
 func (year) Signature() map[Domain]ast.ValueType {
 	return timeFuncSignature
+}
+
+var nowFuncSignature = map[Domain]ast.ValueType{}
+
+// Initialize Now function signature
+func init() {
+	d := Domain{}
+	nowFuncSignature[d] = ast.TTime
+}
+
+type now struct {
+}
+
+func (now) Reset() {
+}
+
+// Return the current local time.
+func (now) Call(args ...interface{}) (v interface{}, err error) {
+	if len(args) != 0 {
+		return 0, errors.New("now expects exactly zero argument")
+	}
+	v = time.Now()
+
+	return
+}
+
+func (now) Signature() map[Domain]ast.ValueType {
+	return nowFuncSignature
 }
 
 type humanBytes struct {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
     The breaking change is that the Combine and Flatten nodes previously, but erroneously, operated across batch boundaries; this has been fixed.
 - [#1497](https://github.com/influxdata/kapacitor/pull/1497): Add support for Docker Swarm autoscaling services.
 - [#1485](https://github.com/influxdata/kapacitor/issues/1485): Add bools field types to UDFs.
+- [#1549](https://github.com/influxdata/kapacitor/issues/1549): Add stateless now() function to get the current local time.
 - [#1545](https://github.com/influxdata/kapacitor/pull/1545): Add support for timeout, tags and service template in the Alerta AlertNode
 - [#1568](https://github.com/influxdata/kapacitor/issues/1568): Add support for custom HTTP Post bodies via a template system.
 - [#1569](https://github.com/influxdata/kapacitor/issues/1569): Add support for add the HTTP status code as a field when using httpPost
--- a/integrations/data/TestStream_EvalNow.srpl
+++ b/integrations/data/TestStream_EvalNow.srpl
@@ -1,0 +1,3 @@
+dbname
+rpname
+account,owner=ownerA expiration=315533000000000000i 0000000001
--- a/integrations/data/TestStream_LambdaNow.srpl
+++ b/integrations/data/TestStream_LambdaNow.srpl
@@ -1,0 +1,9 @@
+dbname
+rpname
+account,owner=ownerA expiration=315533000000000000i 0000000001
+dbname
+rpname
+account,owner=ownerB expiration=4102440000000000000i 0000000001
+dbname
+rpname
+account,owner=ownerC expiration=656419000000000000i 0000000001


tick/stateful/functions.go
```diff
@@ -1354,6 +1355,27 @@ func (year) Signature() map[Domain]ast.ValueType {
...
+		return 0, errors.New("now expects exactly zero argument")
+	}
+
+	v = time.Now()
+
+	return
+}
+
+func (now) Signature() map[Domain]ast.ValueType {
+	return timeFuncSignature
```
desa: I think we want a slightly different function signature here.

The `timeFuncSignature` is

```
fn: ast.TTime -> ast.TInt
```

I think we want

```
fn: nil -> ast.TTime
```

mbresson: Ouch, you're right, I didn't pay attention to that bit. I'll rework on the PR and get back to you.

mbresson:
Hi! In case you don't have time to have a look at my recent changes on my PR, I:

- fixed the signature for the function now()
- added an integration test
- made the function now() rely on a package-local variable (timeNowFunction). I needed to do so in order to have deterministic tests by stubbing now(). If you prefer another approach or feel like it doesn't fit well enough in Kapacitor's source code and code style, just let me know, I'm open to other ideas.

Also, I know that my test is failing right now and I'm working on it, however I have an issue with not being able to run the integration tests properly on my machine. If I do

```./test.sh```

I end up with lots of tests failing (besides of the one I wrote) with lots of messages of this form:

```
Get http://[::]:36319/kapacitor/v1/tasks/TestStream_....: dial tcp [::]:36319: connect: cannot assign requested address
```

I guess it comes from my personal machine and system configuration, and I know here is not the proper place to ask this, but if you have any clue of what could cause this problem, I'd be pleased to know.

**Comment by desa** [comment]
@mbresson if you're using go 1.9+, I'd recommend just running `go test ./...`. If you're using a version less than that you'll need to specify the packages you'd like to test explicitly `go test . ./alert (...)`.

**Comment by mbresson** [comment]
Hi!
Thanks for the tip, it worked. I have fixed the test. Can you have a look and tell me if you are satisfied with my PR? Otherwise I can rework it, add more tests or fix whatever is not right.

File: tick/stateful/functions.go [hunk 2]
```diff
@@ -1354,6 +1355,41 @@ func (year) Signature() map[Domain]ast.ValueType {
...
+
+// Initialize Now function signature
+func init() {
+	d := Domain{}
+	nowFuncSignature[d] = ast.TTime
+}
+
+var timeNowFunc func() time.Time = time.Now
+
+func SetTimeNowFunc(newTimeNowFunc func() time.Time) {
```
desa: This function is only for testing right? In general, I'm not a huge fan of having package level functions like this. I think we should just get rid of this and use `time.Now` explicitly in `now.Call`.

integrations/streamer_test.go
```diff
@@ -9027,6 +9028,59 @@ stream
 	}
 }
 
+func TestStream_LambdaNow(t *testing.T) {
```
desa: This test is great! I'd love to see what happens if we did something like

```
stream
    |from()
      .measurement('account')
    |eval(lambda: "expiration" < now())
      .as('now')
    |httpOut('TestStream_EvalNow')
```

I'd also be curious to know what happens when we try to write this to Influx

```
stream
    |from()
      .measurement('account')
    |eval(lambda: "expiration" < now())
      .as('now')
    |influxDBOut()
	.database('db')
	.retentionPolicy('rp')
	.measurement('m')
	.precision('s')
	.tag('key', 'value')
	.flushInterval(1ms)
```
See the `TestStream_InfluxDBOut` for an example of this. I suspect there might be issues when we turn the time value into line protocol.

integrations/streamer_test.go
```diff
@@ -9027,6 +9028,59 @@ stream
...
+					{
+						time.Date(1971, 1, 1, 0, 0, 0, 0, time.UTC),
+						float64(6.56419e+17), // 1990-10-20 10:42:42
+					},
+				},
+			},
+		},
+	}
+
+	var timeNowFunc func() time.Time = func() time.Time {
```
desa: I think this test should work perfectly fine if we rely on the current`time.Now()`. I don't think we need to mock time.

tick/stateful/functions.go
mbresson: Yes, it's only for testing. I'm not fan of this way of doing things either, but without it, the tests will be reliant on a parameter (time) that's not deterministic. But I understand your point of view and I'll rework my PR.

integrations/streamer_test.go
mbresson: Thanks. I'll try out adding another test with this and I'll get back to you when I'm done.

integrations/streamer_test.go
mbresson: The input I use for the test has a field named expiration, with the values 1980, 2100, 1990. If I rely directly on time.Now() it means the test will break after 2100. I know it's far, far away and you and I won't exist anymore (and maybe kapacitor will not exist anymore, who knows) but the test will not be deterministic because of this. Is it ok?

integrations/streamer_test.go
mbresson: Hi! I don't understand what you want me to do with these tests actually. Because 
* ```|eval(lambda: "expiration" < now())``` is going to output a boolean, right? Not a time value
* Even if we get a time value, I don't understand the use for this, because in my opinion now() is intended to be used with other functions such as unixnano() or minute() or something else but never alone.

Anyway, I have committed the tests I just wrote and which fail, but in order to fix them I'll need to know what kind of output you would expect for them because it's not entirely clear for me :).

Thanks in advance

desa: Ah! meant to do this

```
...
    |eval(lambda: now())
      .as('now')
...
```


> Even if we get a time value, I don't understand the use for this, because in my opinion now() is intended to be used with other functions such as unixnano() or minute() or something else but never alone.

I agree that they'll be used with `unixNano`, but my concern is what happens if I try to use it as a field value. In particular, what happens when we try to set it as a field.

mbresson: I updated the tests. They are failing, but I don't know why. I have read some part of the code *tick/eval.go* but I don't understand a lot about it.
Here is the output of one of the failed tests:

```
--- FAIL: TestStream_EvalNowInfluxDBOut (0.00s)
	streamer_test.go:11032: digraph TestStream_EvalNowInfluxDBOut {
		stream0 -> from1;
		from1 -> eval2;
		eval2 -> influxdb_out3;
		}
	streamer_test.go:9171: got  exp db
	streamer_test.go:9174: got  exp rp
	streamer_test.go:9177: got  exp s
	streamer_test.go:9180: got 0 exp 1
```

Can you give me any pointers on this? I don't have any useful log to understand what's going on. I would have expected the test to at least create the database, but it seems not. It must mean I wrote them wrong.

integrations/streamer_test.go
desa: Yeah I think it's okay. I don't think this project is still around in its current form in 2100. :)

integrations/streamer_test.go
```diff
@@ -9027,6 +9027,178 @@ stream
...
+
+	if database != "db" {
+		t.Errorf("got %v exp %v", database, "db")
+	}
+	if rp != "rp" {
+		t.Errorf("got %v exp %v", rp, "rp")
+	}
+	if precision != "s" {
+		t.Errorf("got %v exp %v", precision, "s")
+	}
```
desa: I think we can disregard these checks here since they're not what we're really testing.

integrations/streamer_test.go
```diff
@@ -9027,6 +9027,178 @@ stream
...
+
+	tmInit := func(tm *kapacitor.TaskMaster) {
+		tm.InfluxDBService = influxdb
+	}
+	testStreamerNoOutput(t, "TestStream_EvalNowInfluxDBOut", script, 15*time.Second, tmInit)
+
+	if database != "db" {
+		t.Errorf("got %v exp %v", database, "db")
+	}
+	if rp != "rp" {
```
desa: I think we can disregard these checks here since they're not what we're really testing.

integrations/streamer_test.go
```diff
@@ -9027,6 +9027,178 @@ stream
...
+		points, err = imodels.ParsePointsWithPrecision(b, time.Unix(0, 0), precision)
+		done <- err
+	}))
+
+	tmInit := func(tm *kapacitor.TaskMaster) {
+		tm.InfluxDBService = influxdb
+	}
+	testStreamerNoOutput(t, "TestStream_EvalNowInfluxDBOut", script, 15*time.Second, tmInit)
+
+	if database != "db" {
```
desa: I think we can disregard these checks here since they're not what we're really testing.

integrations/streamer_test.go
```diff
@@ -9027,6 +9027,178 @@ stream
...
+		if r.URL.Path == "/ping" {
+			w.WriteHeader(http.StatusNoContent)
+			return
+		}
+		//Respond
+		var data client.Response
+		w.WriteHeader(http.StatusOK)
+		_ = json.NewEncoder(w).Encode(data)
+		//Get request data
+		database = r.URL.Query().Get("db")
```
desa: we can ignore these

integrations/streamer_test.go
```diff
@@ -9027,6 +9027,178 @@ stream
...
+
+	influxdb := NewMockInfluxDBService(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		if r.URL.Path == "/ping" {
+			w.WriteHeader(http.StatusNoContent)
+			return
+		}
+		//Respond
+		var data client.Response
+		w.WriteHeader(http.StatusOK)
+		_ = json.NewEncoder(w).Encode(data)
```
desa: we should probably check the error here. This might be out issue.

integrations/streamer_test.go
```diff
@@ -9027,6 +9027,178 @@ stream
...
+		rp = r.URL.Query().Get("rp")
+		precision = r.URL.Query().Get("precision")
+
+		b, err := ioutil.ReadAll(r.Body)
+		if err != nil {
+			done <- err
+			return
+		}
+		points, err = imodels.ParsePointsWithPrecision(b, time.Unix(0, 0), precision)
+		done <- err
```
desa: might want to wait on the channel send output of this function before we do all the other checks.

integrations/streamer_test.go
```diff
@@ -9027,6 +9027,178 @@ stream
...
+		database = r.URL.Query().Get("db")
+		rp = r.URL.Query().Get("rp")
+		precision = r.URL.Query().Get("precision")
+
+		b, err := ioutil.ReadAll(r.Body)
+		if err != nil {
+			done <- err
+			return
+		}
+		points, err = imodels.ParsePointsWithPrecision(b, time.Unix(0, 0), precision)
```
desa: I think we might be getting an error here and that's causing the issue

integrations/streamer_test.go
mbresson: Hi. I made the changes and added a lot of logging (nothing committed) to try to understand what was going on, since I'm not familiar with how kapacitor is built. It looks like there is no HTTP request made at all to InfluxDB. And it seems that eval is not even called (tried to log inside NewEvalLambdaNode for testing purposes and got nothing). I'll continue my investigations another day.

desa:
@mbresson just curious if you've made any more progress on this issue. If the InfluxDBOut stuff is still causing issues, maybe we should just make it so that `now` returns an integer. This will un-complicate the InfluxDB issues that may come up.

mbresson:
Hi, I have made some progress but I'm still not there yet. Nonetheless I prefer not to make now() return an integer and keep working on it, because having now() returning a time.Time is more interesting, especially if it can be used with other functions taking a time in parameter. I'll keep you updated.

desa:
@mbresson Sounds good. Just a heads up, we're getting ready to start the 1.4 release process and I'd love to see this PR make it in there!

> And I prefer not to make now() return an integer and keep working on it, because having now() returning a time.Time is more interesting, especially if it can be used with other functions taking a time in parameter. 

I definitely agree that having a `time.Time` value would be more useful, but I think that change might be more involved. In fact I think it might involve a change with InfluxDB, I could be wrong though, maybe just before it's written to InfluxDB convert it to an int? Just kind of thinking out loud.

mbresson:
I should have time to push my updates by tomorrow evening or the day before tomorrow, just didn't have time to do it these last days.
No need to change anything in InfluxDB, but there is a choice to make: before writing it to InfluxDB, should we convert it to an int (Unix timestamp), or to a string representation of the time ? (such as ISO8601)
Tell me what you prefer.

mbresson:
Hi, finally, here are my updates:

- In order to be able to evaluate time.Time values, in tick/ast/node.go, I added a TimeNode type, and I also added the file tick/stateful/eval_time_node.go for the same reasons.
- In edge/messages's ToRow methods, I convert time.Time values to unix timestamps in nanoseconds (int64). I do the same in influxdb/client.go's Bytes method. I do it so that we can convert the time.Time values we get when now() is evaluated to integral values. Otherwise, they are automatically converted to string, ISO8601 representation of dates, which is not what we want.
- I have changed my test TestStream_EvalNow because it was impossible to have it pass if I made it output now(), since the expected value could never be exactly the same, nanoseconds-wise, as the value produced. Instead I altered the test to get the current year() and compare it, but I think it makes the test much less useful since initially it was intended to check what would happen if now() was fed to httpOut. I'm not very satisfied for one more reason, it is that if that test happens to run at the very end of a year, it may fail because the expected value for year may be the next year. That's the problem with testing non-deterministic time values :P.
- I have completed the test TestStream_EvalNowInfluxDBOut, and it checks that the time.Time value resulting from lambda: now() is indeed sent as a unix nanoseconds timestamp, encoded as an int64.

I'm not sure my explanations are clear because to be blunt, my understanding of Kapacitor's source code is incomplete, so I'd appreciate if you could double-check the new code that I brought in. Maybe some more tests could be added, if you feel like it's needed, I can work on it another day.

Thanks

mbresson:
It seems that the CI checks have failed. I didn't have any error when running the tests on my machine, and I cannot access the CI logs (the loader keep spinning forever). I'll check again tomorrow.

desa:
@mbresson  looks like its just a flakey test. I'm rerunning the build.

edge/messages.go
```diff
@@ -330,7 +330,19 @@ func (pm *pointMessage) ToRow() *models.Row {
 	row.Values[0][0] = pm.time
 	for i, c := range row.Columns[1:] {
 		if v, ok := pm.fields[c]; ok {
-			row.Values[0][i+1] = v
+			finalValue := v
+
+			switch v.(type) {
```
desa: This could be turned into

```go
val, ok := v.(time.Time)
if ok {
v = val.UnixNano()
}

row.Values[0][i+1] = v

```

edge/messages.go
```diff
@@ -701,7 +713,19 @@ func (bb *bufferedBatchMessage) ToRow() (row *models.Row) {
 		row.Values[i][0] = p.Time()
 		for j, c := range row.Columns[1:] {
 			if v, ok := p.Fields()[c]; ok {
-				row.Values[i][j+1] = v
+				finalValue := v
+
+				switch v.(type) {
```
desa: Same point as the one above.

influxdb/client.go
```diff
@@ -517,7 +517,22 @@ type Point struct {
 // Returns byte array of a line protocol representation of the point
 func (p Point) Bytes(precision string) []byte {
 	key := imodels.MakeKey([]byte(p.Name), imodels.NewTags(p.Tags))
-	fields := imodels.Fields(p.Fields).MarshalBinary()
+
+	influxDBFields := imodels.Fields(p.Fields)
+
+	for fieldName, fieldValue := range influxDBFields {
+		switch fieldValue.(type) {
+		case time.Time:
```
desa: this could be

```
val, ok := fieldValue.(time.Time)
if ok {
influxDBFields[fieldName] =  = val.UnixNano()
}
```

tick/ast/node.go
```diff
@@ -214,6 +214,36 @@ func (n *DurationNode) Equal(o interface{}) bool {
 	return false
 }
 
+type TimeNode struct {
```
desa: What do we need this for?

mbresson:
Hi, thanks for your feedback! I made the changes you suggested, I think your code suggestions are clearer and more concise.
About the TimeNode, you are right, it is not needed, as other things that I have removed in my latest commit. I added these because I misunderstood some parts of the code in kapacitor. My fault. :) 
I'll be without internet until beginning of next week, so I hurried a little bit for my latest commit. I'm looking forward to your feedback.

desa:
@mbresson looks good! After a rebase and squash into a single commit I think this should be good to merge.

mbresson:
Sorry, I've failed my squash & rebase because now it appears that I'm the (co-)author of commits that are not mine. (f1f9bb0, 8933719, aa33d98, e48f9f0, 81576da, 156349f, 871d796, 77c91f9, ffcb25a, 0dd1d3e, 3a4d43f, 860b95c).
I'm not good at git, do you know how to squash & rebase and retain the original author information?
Either way I'll try and clean up my mess when I'm back next week.

desa:
@mbresson The process should be something like

`git rebase -i master`

that will take you a screen that looks like this

```
  1 pick d25510d0 Add example of all nodes in load tasks style dir
  2
  3 # Rebase 6e45b39d..d25510d0 onto 6e45b39d (1 command)
  4 #
  5 # Commands:
  6 # p, pick = use commit
  7 # r, reword = use commit, but edit the commit message
  8 # e, edit = use commit, but stop for amending
  9 # s, squash = use commit, but meld into previous commit
 10 # f, fixup = like "squash", but discard this commit's log message
 11 # x, exec = run command (the rest of the line) using shell
 12 # d, drop = remove commit
 13 #
 14 # These lines can be re-ordered; they are executed from top to bottom.
 15 #
 16 # If you remove a line here THAT COMMIT WILL BE LOST.
 17 #
 18 # However, if you remove everything, the rebase will be aborted.
 19 #
 20 # Note that empty commits are commented out
```

Change all of the `pick` commits to `squash` except for the very first one and change that one to `reword`. Then quit and reword the commit appropriately.

mbresson:
I actually did **git rebase -i HEAD~10** and squashed all my commits into f1f9bb0ecdd9ff6f653b88e18b9cc9de886dd64f and left other people's commits intact.
But I don't understand why, on Github, it looks like I'm the committer of all commits, not only mine:

commits

Anyway, I'll fix that when I'm back.

mbresson:
Hi! I'm finished at last. I see the checks have failed but it seems to come from another test than mine:
```
=== RUN   TestServer_RecordReplayStreamWithPost
--- FAIL: TestServer_RecordReplayStreamWithPost (1.19s)
	server_test.go:3528: failed to finish recording
```

Maybe it's an issue with the CI system again?

mbresson:
Hi, is everything ok with my pull request? Please let me know if some rework or enhancement is needed.

desa:
@mbresson sorry about the delay in my response. After a bit of internal discussion, we think that allowing `time.Time`s to be passed around as fields is a bit more of a substantial change. There are a number of places where we make assumptions that the fields will be an `int`, `float`, `string`, or `bool`.

We think the `now` function is still a great addition, the only change from whats here will be

```
stream
    |from()
        .measurement('cpu')
    |eval(lambda: now())
        .as('now')
```

should result in

```
ts=2017-10-27T12:02:35.313-04:00 lvl=error msg="error evaluating expression" service=kapacitor task_master=main task=now node=log4 err="expression returned unexpected type time"
```

But 

```
stream
    |from()
        .measurement('cpu')
    |where(lambda: "time" > now())
```
and
```
stream
    |from()
        .measurement('cpu')
    |eval(lambda: unixNano(now()))
        .as('now')
```
are perfectly fine.

My apologies the last second changes. I wasn't thinking clearly when I suggested that we allow times be passed around like values. I'll go through and note the changes necessary to get us to the state suggested above.

edge/messages.go
```diff
@@ -330,6 +330,15 @@ func (pm *pointMessage) ToRow() *models.Row {
 	row.Values[0][0] = pm.time
 	for i, c := range row.Columns[1:] {
 		if v, ok := pm.fields[c]; ok {
+
+			// force converting time.Time fields (other than the built-in time column)
+			// to an integer value representing a Unix nanoseconds timestamp
+			// otherwise, they'll be turned into ISO8601 strings during marshalling
+			timeValue, isTime := v.(time.Time)
```
desa: we shouldn't need this any more.

edge/messages.go
```diff
@@ -701,6 +710,15 @@ func (bb *bufferedBatchMessage) ToRow() (row *models.Row) {
 		row.Values[i][0] = p.Time()
 		for j, c := range row.Columns[1:] {
 			if v, ok := p.Fields()[c]; ok {
+
+				// force converting time.Time fields (other than the built-in time column)
+				// to an integer value representing a Unix nanoseconds timestamp
+				// otherwise, they'll be turned into ISO8601 strings during marshalling
+				timeValue, isTime := v.(time.Time)
```
desa: we shouldn't need this any more.

influxdb/client.go
```diff
@@ -517,7 +517,22 @@ type Point struct {
 // Returns byte array of a line protocol representation of the point
 func (p Point) Bytes(precision string) []byte {
 	key := imodels.MakeKey([]byte(p.Name), imodels.NewTags(p.Tags))
-	fields := imodels.Fields(p.Fields).MarshalBinary()
+
```
desa: we shouldn't need this any more.

integrations/streamer_test.go
```diff
@@ -9037,6 +9037,162 @@ stream
...
+					},
+				},
+			},
+		},
+	}
+
+	testStreamerWithOutput(t, "TestStream_EvalNow", script, time.Second, expectedOutput, false, nil)
+}
+
+func TestStream_EvalNowInfluxDBOut(t *testing.T) {
```
desa: we shouldn't need this any more.

tick/stateful/expr.go
```diff
@@ -19,6 +19,7 @@ type Expression interface {
 	EvalString(scope *Scope) (string, error)
 	EvalBool(scope *Scope) (bool, error)
 	EvalDuration(scope *Scope) (time.Duration, error)
+	EvalTime(scope *Scope) (time.Time, error)
```
desa: we shouldn't need this any more.

tick/stateful/expr.go
```diff
@@ -84,6 +85,10 @@ func (se *expression) EvalDuration(scope *Scope) (time.Duration, error) {
 	return se.nodeEvaluator.EvalDuration(scope, se.executionState)
 }
 
+func (se *expression) EvalTime(scope *Scope) (time.Time, error) {
```
desa: we shouldn't need this any more.

tick/stateful/expr.go
```diff
@@ -131,6 +136,12 @@ func (se *expression) Eval(scope *Scope) (interface{}, error) {
 			return nil, err
 		}
 		return result, err
+	case ast.TTime:
```
desa: we shouldn't need this any more.

mbresson:
No problem, I understand. I have just made the relevant changes.

PR #1555, opened 2017-09-12
Ability to use basic auth in kapacitor CLI

gburanov:
I am using kapacitor cmd utility in all my script.

The problem is that if we are trying to connect to kapacitor from outside of our vpn, we need auth.

I found no way to set the auth in cmd line, so I am trying this approach. WDYT?

### DIFF
--- a/cmd/kapacitor/main.go
+++ b/cmd/kapacitor/main.go
@@ -104,8 +104,14 @@ func main() {
 		url = *kapacitordURL
 	}
 
+	var creds *client.Credentials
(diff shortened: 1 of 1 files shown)


desa:
@gburanov why isn't specifying basic auth credentials in the `KAPACITOR_URL` an option? e.g.

```sh
KAPACITOR_URL=http://user:pass@localhost:9092 kapacitor ...
```

nathanielc:
@gburanov I agree with @desa using `KAPACITOR_URL` with the credentials in the URL should be sufficient here. Looks like we should  mention that its supported in the CLI help somewhere. Closing this PR. Would accept another PR to improve the docs around this.

Issue #1304, opened 2017-04-07
"Reduce" node and reacting to silent periods


sputnik13:
We're wanting to do something like the following...

```
stream
  | from()
      .measurement('some_measurement')
  | where(lambda: "some condition")
  | stats(1s)
  | derivative('emitted')
      .unit(1s)
      .nonNegative()
  | window()
      .periodCount('3600s')
      .everyCount(1)
  | reduce(lambda: \"emitted\" + \"_previous\")        <-- new node type
      .as("count")
  | alert()
      .crit(lambda: \"count\" > 0)
```

here "reduce" would take a batch, apply the lambda against each point, and supply intermediate values to the lambda as "_previous" (open to suggestion for a better name)

As for why we want this, we're trying to apply rules/tasks that react to messages/events/metrics that come and go (max-throughput, deadman, etc), we're having to deal with silent periods where the data we're trying to alert on doesn't just go down in frequency but goes also doesn't emit at all for extended periods.

We also have requirements to alert in a timely manner (e.g. evaluate max-throughput over a 1 hour window but alert within a few seconds of alert being triggered) that forces us to use a sliding window rather than stats based approach to evaluating alert conditions.

After some thought along the lines of joining the evaluated stream with a separate stats stream to force evaluation of the alert logic, we discarded this approach as it requires lambda expressions to be modified to account for the prefixes applied by the join node to each data field.  Also creating a stream just to join it and discard it to force evaluation of the pipeline just felt...  dirty.

We're willing to do the implementation of the node, we're looking for feedback as to whether this seems useful enough to include in kapacitor proper.

sputnik13:
looks like we can use sum('emitted').as('count') instead...  I guess I shouldn't be making decisions at 2 am :)

although, a general reduce function does seem useful, it's not absolutely necessary for our current needs, so...  closing request

sputnik13:
So we're in need of this functionality again, due to some changing requirements...  I think this could be either a separate reduce node or a modified eval node.  Having a separate reduce node would probably duplicate some code from the eval node but from the user's standpoint, the symantics would be simpler.

nathanielc:
@sputnik13 I want to make sure I understand the request. Basically the reduce node will maintain a single point and reduce each new point into that point via the provided lambda expression?

Meaning I could implement a `count` function by doing something like this?

```
|reduce(lambda: "_previous" + 1)
```

And a sum could be (assuming the data has a field named `value`):

```
[...]

PR #1566, opened 2017-09-17
Add BarrierNode to emit BarrierMessage periodically

sputnik13:
Adding a BarrierNode to emit a BarrierMessage on intervals that are driven by the clock rather than by message timestamps.

Addresses portion of #1304 wherein "silent periods" need to be handled by minimum throughput detection pipelines.

- [x] Rebased/mergable
- [x] Tests pass
- [x] CHANGELOG.md updated

### DIFF
--- a/barrier.go
+++ b/barrier.go
@@ -8,19 +8,14 @@ import (
 	"sync/atomic"
 
 	"github.com/influxdata/kapacitor/edge"
+	"github.com/influxdata/kapacitor/models"
 	"github.com/influxdata/kapacitor/pipeline"
 )
 
 type BarrierNode struct {
 	node
-	b *pipeline.BarrierNode
-
-	idle   time.Duration
-	period time.Duration
-	lastT  atomic.Value
-	timer  *time.Timer
-	ticker *time.Ticker
-	wg     sync.WaitGroup
+	b              *pipeline.BarrierNode
+	barrierStopper map[models.GroupID]func()
 }
 
 // Create a new  BarrierNode, which emits a barrier if data traffic has been idle for the configured amount of time.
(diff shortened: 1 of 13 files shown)


sputnik13:
There's a slight departure in the integration test from other integration tests.  Rather than using a data set from a file, data set is being generated ad hoc with a negative time offset from current time and passed in via channel to the test fixture.  Could not see a better way to simulate having a silent period after a dataset is fed to the fixture.

sputnik13:
So it seems like the only way to solve the race condition may be to have edge.Collect take a lock every time, but that would mean every edge would require a lock for every point, which is probably going to be expensive.  Would it be better to have a periodic barrier node of some sort?

pipeline/barrier.go
```diff
@@ -0,0 +1,43 @@
...
+//        |top('value', 10)
+//        //Post the top 10 results over the last 10s updated every 5s.
+//        |httpPost('http://example.com/api/top10')
+//
+type BarrierNode struct {
+	chainnode
+
+	// How often a barrier is emitted into the pipeline.
+	// Must be greater than zero.
+	Every time.Duration
```
nathanielc: We should add a comment there that the time is real time, not data time. I know you have it above but it would be good to reiterate here as well.

sputnik13:
@nathanielc I will incorporate your feedback and update the PR, thanks for the comments

sputnik13:
@nathanielc so after some time implementing this, I've come to the realization using idle wouldn't provide the desired behavior for us, we actually need a periodic barrier, the reason being that an idle barrier would either emit too few barriers or too many barriers to window, depending on whether the barrier node itself is a grouped consumer or a non-grouped consumer.

If Barrier is a Grouped consumer
- we'd need to have timers for each group
- a barrier would be emitted for each group that goes idle
- window would receive each of these barriers, and send each barrier to every group due to GroupedConsumer treating each barrier as being intended for every group (https://github.com/influxdata/kapacitor/blob/master/edge/grouped.go#L111)

If Barrier is not a Grouped consumer
- we'd need to have a single timer for all data
- a barrier would be emitted when no data is received, so if a single group goes idle that would not trigger a barrier
- window node would receive a barrier only when all groups go idle, but it would forward the barrier to the window for every group

I think the treatment of barrier, by grouped consumer, as being intended for every group is making this complicated.

I'm wondering whether GroupedConsumer should treat each Barrier as being intended for a specific group rather than applying it to every group.  Otherwise forwarding Barrier from any node becomes dangerous (and perhaps it's not intended to be forwarded) as chaining multiple GroupedConsumers would result in exponential growth of Barriers.

Perhaps the intent is that Barrier is a special message type that only gets consumed, and never forwarded, and should always be at the head of the pipeline, in which case the current treatment of Barrier by GroupedConsumer would seem to make sense.

The above explanation for the design not withstanding, if GroupedConsumer isn't changed to handle Barrier differently, our use case requires a .periodic() rather than .idle() (although I don't mind implementing both).

Thoughts?

sputnik13:
Another unexpected detail about the side effect of barrier being fanned out to every member of the groupedconsumer, it looks like barrier is never received by any group handler in a groupedconsumer if it has never received a message before.  This makes sense once you understand the implementation details, but at a high level it was outside of expectation, as I wrote a test task like the following

'''
var period = 14s
var every = 2s
var idle = 2s
stream
	|from()
		.database('dbname')
		.retentionPolicy('rpname')
		.measurement('cpu')
	|barrier().idle(idle)
	|window()
		.period(period)
		.every(every)
	|httpPost('` + ts.URL + `')
'''

expecting that when there have never been any messages, a zero length window would be emitted by window().  But since no group has ever been created (due to lack of any messages going through the window node) the barrier is actually never seen by the BarrierMessage handler in the window.

Does this behavior make sense or is it an unintended side effect?

nathanielc:
@sputnik13 The Barrier concept is new and is not used by any existing system as of yet. The first intended use is actually what this PR is doing. In other words we can change the behavior of Barrier as we currently have it, to meet the needs of this PR and hopefully future features as well. Its current behavior was only a best guess as to how it would be useful.

Feel free to experiment with changing Barrier to be specific to each group. My initial thought would be that barriers are still forwarded on because they carry more information than just data arrival.

My thinking is that all GroupedConsumers would process the barrier message and then forward on the message. Not in fan out but on a per group basis.

Non-grouped consumers will have to decide when a down stream barrier message can be sent for each downstream group. For example the groupBy node is necessarily changing the groups and so simply forwarding the Barrier messages will make no sense. My thinking is that the groupBy node will still need to emit Barrier messages if it receives them, but only after all dependent Barrier messages have been received. For example if groups A and B are being merged into group C by the groupBy node, once Barrier messages for A and B are received then a barrier message for C will be emitted.

My reasoning for wanting the Barrier messages to go further than one step is for things like join, that can use Barrier messages to emit joined points much sooner. 

Thoughts? 

Like I said the Barrier concept is new, so we have the freedom to change it to be the correct design for the use case.

barrier.go
```diff
@@ -0,0 +1,273 @@
...
+	if !m.Time().Before(n.lastT.Load().(time.Time)) {
+		return m, nil
+	}
+	return nil, nil
+}
+
+func (n *periodicBarrier) emitBarrier() error {
+	nowT := time.Now()
+	n.lastT.Store(nowT)
+	return edge.Forward(n.outs, edge.NewBarrierMessage(n.group, nowT))
```
nathanielc: Instead of using an atomic.Value on `n.lastT` I think using a sync.Mutex will be helpful, since you will want to hold that lock when your are writing to the `n.outs` since you do not want to concurrently write the Barrier message with any other message.

barrier.go
```diff
@@ -0,0 +1,273 @@
...
+
+func (n *periodicBarrier) Point(m edge.PointMessage) (edge.Message, error) {
+	if !m.Time().Before(n.lastT.Load().(time.Time)) {
+		return m, nil
+	}
+	return nil, nil
+}
+
+func (n *periodicBarrier) emitBarrier() error {
+	nowT := time.Now()
```
nathanielc: This time needs to be the last time seen in the data stream, not the current time. This is because its possible to write historical data, i.e. in a replay.

sputnik13: Replay already seems to work.  My only reservation with using last message time here is that we'd need to store the last message time for every received message, which, if we're using a Mutex to protect access to the variable, would mean we're acquiring a write lock for every time a message is received.  If we can continue using atomic rather than Mutex this might be OK.  Also I have to admit I don't know what the cost of Mutex is in go vs atomic operations.

barrier.go
```diff
@@ -89,7 +89,7 @@ type idleBarrier struct {
 	timer *time.Timer
 	wg    sync.WaitGroup
 	outs  []edge.StatsEdge
-	stopC chan bool
+	stopC chan interface{}
```
nathanielc: This should be `chan struct{}` as `struct{}` will be an empty value and the compiler will optimize it out.  On the other hand an interface needs room for a pointer to type information and a pointer to the instance.

barrier.go
```diff
@@ -158,9 +158,7 @@ func (n *idleBarrier) Point(m edge.PointMessage) (edge.Message, error) {
 }
 
 func (n *idleBarrier) resetTimer() {
-	if n.idle > 0 {
-		n.timer.Reset(n.idle)
-	}
+	n.timer.Reset(n.idle)
```
nathanielc: How does this change prevent the race on resetting the timer?

barrier.go
nathanielc: Using atomic to store the last time is fine, you can even use an int64 as nanoseconds since the epoch. You should only need to store the greatest time per group.

In general atomic is faster than getting a mutex lock, but atomic only works if each individual piece of state can be modified independently, otherwise you need a mutex lock to ensure all state is modified together. But in either case making an atomic write or getting a mutex lock shouldn't be much cost compared to the cost of processing messages in general.

barrier.go
nathanielc: The channels prevent a race condition during write but you still need to guarantee that you do not mix messages in a invalid order. For example during a batch write only batchpoint messages should exist between begin and end messages.

barrier.go
sputnik13: the n.idle > 0 check is an unnecessary carry-over from when the idle and periodic barrier were one implementation, since idleBarrier wouldn't be instantiated unless n.idle is > 0 this is an unnecessary check..

also potential race conditions on resetting the timer doesn't seem necessary to check for, timer.Reset will advance the expiration to now + duration and since we're always advancing/resetting by n.idle any minute difference between two simultaneous writers shouldn't have much of an impact

barrier.go
sputnik13: So beyond the performance impact (which I agree is negligible) it seems like setting barrier time to the last time seen in the data stream creates a problem when trying to handle a stream with no data.

My thought was that window should be forced to emit when a barrier is received with a timestamp that is past the nextEmit time, which would make sense from the window's perspective since no data past the time in the barrier will be received, therefore it can assume that the window needs to be emitted.  However, setting barrier time to the last time seen in the data stream makes it such that the window would either
1) never emit until a message past nextEmit is received, which makes it so that silent periods can't be handled, or
2) emit any time a barrier is received, irregardless of whether the barrier time is before nextEmit, which logically doesn't make sense to me

So the first would bring us back to square one since we need the ability to force emission of window when there are silent periods.  So I think to support the forced emit behavior either barrier time gets set to the wall clock or we force window to emit on receipt of barrier irregardless of timestamp on the barrier...  thoughts?

sputnik13:
@nathanielc posted some responses to your comments/responses above

barrier.go
nathanielc: OK, removing the check makes sense. 

The issue with the race is that it can cause undefined state in the timer, thus rendering the timer useless. It may even cause a panic depending on how the timer is implemented. To safely reuse a timer we have to protect the Reset process as documented here https://golang.org/pkg/time/#Timer.Reset.

pipeline/barrier.go
```diff
@@ -0,0 +1,57 @@
...
+// clock.  Any messages received after an emitted barrier that is older than the last
+// emitted barrier will be dropped.
+//
+// Example:
+//    stream
+//        |barrier().idle(5s)
+//        |window()
+//            .period(10s)
+//            .every(5s)
+//        |top('value', 10)
```
nathanielc: Small typo, this should be
```
|top(10, 'value')
```

barrier.go
```diff
@@ -0,0 +1,277 @@
...
+	}
+	return nil, nil
+}
+
+func (n *idleBarrier) resetTimer() {
+	n.resetTimerC <- struct{}{}
+}
+
+func (n *idleBarrier) emitBarrier() error {
+	nowT := time.Now()
```
nathanielc: This should be `time.Now().UTC()` as we want to keep all time value in the UTC timezone.

Issue #894, opened 2016-09-12
Enhancement: Exposing TICK Vars to Alert template expansion


octalthorpe:
Currently there is no way to access TICK variable from within a alert template, so pre expansion needs to be done.  As an example the following will not work as 'host_count_warn' and 'host_count_critical' are not present in the template expansion variable scope.
```
var host_count_warn = 2
var host_count_critical = 1

.... Stuff ....
|alert()
.message('Healty Host Count {{ .TaskName }} is {{ .Level }}: Got a value of {{ index .Fields "_host_count" }} expecting > {{ if .Level eq "WARN" }} {{ .host_count_warn }} {{ else }} {{ .host_count_critical }} {{ end }}')
```

The current solution is to pre-expanded using the "string" function the variable before the template is executed, i.e.
```
'{{ if eq .Level "WARN" }} ' + string(host_count_warn) + '{{ else }}' + string(host_count_critical) + '{{ end }}'
```

It would be nice and much cleaner if we could simply access the variables from a tickscript directly in the alert message as part of the template expansion.  

I would propose a new map "TaskVars" or "TickTaskVars" etc that will be populated with the tick script variables, and referenced as such:
```
Healthy Host Count {{.TaskName}} is {{ .Level }}: Got a value of {{ index .Fields "_host_count" }} expecting > {{ if eq .Level "WARN" }} {{ index .TaskVars "host_count_warn"}} {{ else }} {{ index TaskVars "host_count_critical"}} {{ end }}
```

evbo:
In addition, another use case is accessing fields or tags by a variable key, for instance:

```
var field string = "\"some_field\""
|alert()
    .message('Some alert on field:  {{ index .Fields field }}')
```

I have not seen a workaround that can accomplish this yet, other than writing individual templates for each field of interest being alerted on.

avdhoot:
@nathanielc Any updadate :)

mcassaniti:
If you're going to extend this for the message property, then extending for id and log would be ideal too.

ashishkaransingh:
May be it is assumption that with flux this is not getting probably a priority but this is super important or us as we need to send alerts in a format which servicenow itom can accept and therefore must be able to customize HTTP post bodies.

Wwe should be able to send something like this

Expected
Must be able to access  .Tags outside .Message

{
"id": "{{.ID}}",
"message": "{{.Message}}",
"node": "{{index .Tags "host"}}", ---------> Currently I cannot access .Tags outside .Message
[...]

PR #1572, opened 2017-09-19
Add support for custom HTTP post bodies via a template system

### LINKED ISSUE #1568: HTTP Post endpoints with custom HTTP Body
The HTTP Post Node should allow for the body to be customizable. This way Kapacitor will be able to inter-operate with many different systems since the body can be customized. 

Proposal:

* Add a `template` option to an httppost endpoint configuration that contains a Go template.
 Use a similar structure to alert messages to populate data for the template evaluation step.
* Add a `template_file` option which lets you define a template via an absolute file path instead of placing the template directly in the configuration.

@desa Thoughts?

nathanielc:
Fixes #1568 

- [x] Rebased/mergable
- [x] Tests pass
- [x] CHANGELOG.md updated

### DIFF
--- a/http_post.go
+++ b/http_post.go
@@ -32,7 +32,7 @@ func newHTTPPostNode(et *ExecutingTask, n *pipeline.HTTPPostNode, d NodeDiagnost
 
 	// Should only ever be 0 or 1 from validation of n
 	if len(n.URLs) == 1 {
-		e := httppost.NewEndpoint(n.URLs[0], nil, httppost.BasicAuth{})
+		e := httppost.NewEndpoint(n.URLs[0], nil, httppost.BasicAuth{}, nil, nil)
 		hn.endpoint = e
 	}
 
@@ -109,23 +109,34 @@ func (g *httpPostGroup) DeleteGroup(d edge.DeleteGroupMessage) (edge.Message, er
 }
 
 func (n *HTTPPostNode) postRow(row *models.Row) {
-	result := new(models.Result)
-	result.Series = []*models.Row{row}
 
 	body := n.bp.Get()
 	defer n.bp.Put(body)
-	err := json.NewEncoder(body).Encode(result)
-	if err != nil {
-		n.diag.Error("failed to marshal row data json", err)
-		return
+
+	var contentType string
+	if n.endpoint.RowTemplate() != nil {
+		mr := newMappedRow(row)
+		n.endpoint.RowTemplate().Execute(body, mr)
+	} else {
+		result := new(models.Result)
+		result.Series = []*models.Row{row}
+		err := json.NewEncoder(body).Encode(result)
+		if err != nil {
+			n.diag.Error("failed to marshal row data json", err)
+			return
+		}
+		contentType = "application/json"
 	}
+
 	req, err := n.endpoint.NewHTTPRequest(body)
 	if err != nil {
 		n.diag.Error("failed to marshal row data json", err)
 		return
 	}
 
-	req.Header.Set("Content-Type", "application/json")
+	if contentType != "" {
+		req.Header.Set("Content-Type", contentType)
+	}
 	for k, v := range n.c.Headers {
 		req.Header.Set(k, v)
 	}
@@ -135,4 +146,25 @@ func (n *HTTPPostNode) postRow(row *models.Row) {
 		return
 	}
 	resp.Body.Close()
+}
+
+type mappedRow struct {
+	Name   string
+	Tags   map[string]string
+	Values []map[string]interface{}
+}
+
+func newMappedRow(row *models.Row) *mappedRow {
+	values := make([]map[string]interface{}, len(row.Values))
+	for i, v := range row.Values {
+		values[i] = make(map[string]interface{}, len(row.Columns))
+		for c, col := range row.Columns {
+			values[i][col] = v[c]
+		}
+	}
+	return &mappedRow{
+		Name:   row.Name,
+		Tags:   row.Tags,
+		Values: values,
+	}
 }
--- a/services/httppost/service.go
+++ b/services/httppost/service.go
@@ -3,15 +3,16 @@ package httppost
 import (
 	"bytes"
 	"encoding/json"
-	"errors"
 	"fmt"
 	"io"
 	"net/http"
 	"sync"
+	"text/template"
 
 	"github.com/influxdata/kapacitor/alert"
 	"github.com/influxdata/kapacitor/bufpool"
 	"github.com/influxdata/kapacitor/keyvalue"
+	"github.com/pkg/errors"
 )
 
 type Diagnostic interface {
@@ -21,18 +22,22 @@ type Diagnostic interface {
 
 // Only one of name and url should be non-empty
 type Endpoint struct {
-	mu      sync.RWMutex
-	url     string
-	headers map[string]string
-	auth    BasicAuth
-	closed  bool
+	mu            sync.RWMutex
+	url           string
+	headers       map[string]string
+	auth          BasicAuth
+	alertTemplate *template.Template
+	rowTemplate   *template.Template
+	closed        bool
 }
 
-func NewEndpoint(url string, headers map[string]string, auth BasicAuth) *Endpoint {
+func NewEndpoint(url string, headers map[string]string, auth BasicAuth, at, rt *template.Template) *Endpoint {
 	return &Endpoint{
-		url:     url,
-		headers: headers,
-		auth:    auth,
+		url:           url,
+		headers:       headers,
+		auth:          auth,
+		alertTemplate: at,
+		rowTemplate:   rt,
 	}
 }
 func (e *Endpoint) Close() {
@@ -42,12 +47,35 @@ func (e *Endpoint) Close() {
 	return
 }
 
-func (e *Endpoint) Update(c Config) {
+func (e *Endpoint) Update(c Config) error {
 	e.mu.Lock()
 	defer e.mu.Unlock()
 	e.url = c.URL
 	e.headers = c.Headers
 	e.auth = c.BasicAuth
+	at, err := c.getAlertTemplate()
+	if err != nil {
+		return err
+	}
+	e.alertTemplate = at
+	rt, err := c.getRowTemplate()
+	if err != nil {
+		return err
+	}
+	e.rowTemplate = rt
+	return nil
+}
+
+func (e *Endpoint) AlertTemplate() *template.Template {
+	e.mu.RLock()
+	defer e.mu.RUnlock()
+	return e.alertTemplate
+}
+
+func (e *Endpoint) RowTemplate() *template.Template {
+	e.mu.RLock()
+	defer e.mu.RUnlock()
+	return e.rowTemplate
 }
 
 func (e *Endpoint) NewHTTPRequest(body io.Reader) (req *http.Request, err error) {
@@ -79,12 +107,15 @@ type Service struct {
 	diag      Diagnostic
 }
 
-func NewService(c Configs, d Diagnostic) *Service {
-	s := &Service{
-		diag:      d,
-		endpoints: c.index(),
+func NewService(c Configs, d Diagnostic) (*Service, error) {
+	endpoints, err := c.index()
+	if err != nil {
+		return nil, err
 	}
-	return s
+	return &Service{
+		diag:      d,
+		endpoints: endpoints,
+	}, nil
 }
 
 func (s *Service) Endpoint(name string) (*Endpoint, bool) {
@@ -108,10 +139,20 @@ func (s *Service) Update(newConfigs []interface{}) error {
 			}
 			e, ok := s.endpoints[c.Endpoint]
 			if !ok {
-				s.endpoints[c.Endpoint] = NewEndpoint(c.URL, c.Headers, c.BasicAuth)
+				at, err := c.getAlertTemplate()
+				if err != nil {
+					return errors.Wrapf(err, "failed to get alert template for endpoint %q", c.Endpoint)
+				}
+				rt, err := c.getRowTemplate()
+				if err != nil {
+					return errors.Wrapf(err, "failed to get row template for endpoint %q", c.Endpoint)
+				}
+				s.endpoints[c.Endpoint] = NewEndpoint(c.URL, c.Headers, c.BasicAuth, at, rt)
 				continue
 			}
-			e.Update(c)
+			if err := e.Update(c); err != nil {
+				return errors.Wrapf(err, "failed to update endpoint %q", c.Endpoint)
+			}
 
 			endpointSet[c.Endpoint] = true
 		} else {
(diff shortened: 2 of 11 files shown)


nathanielc:
@desa This is ready for review.

http_post.go
```diff
@@ -132,7 +134,9 @@ func (n *HTTPPostNode) postRow(row *models.Row) {
 		return
 	}
 
-	req.Header.Set("Content-Type", "application/json")
+	if contentType != "" {
```
desa: Should `contentType` be settable from the config?

services/httppost/service.go
```diff
@@ -266,8 +300,11 @@ func (h *handler) Handle(event alert.Event) {
 		return
 	}
 
+	if contentType != "" {
```
desa: Same as previous comment. Maybe contentType should be settable on an endpoint.

http_post.go
nathanielc: It is already via the headers system.

desa: Ah good point.

dishmael:
Is there documentation on how to use a template with custom JSON?

nathanielc:
@dishmael Currently the docs around this feature exists in the configuration [here](https://github.com/influxdata/kapacitor/blob/master/etc/kapacitor/kapacitor.conf#L313) as the feature is controlled via the config.

epsylonix:
Could you please clarify how this is supposed to work?

when doing a post from a .tick file with `alert().post('http://localhost:8084/kapacitor')` when there is a url defined in the kapacitor's `[[httppost]]` section, for example `url = "http://example.com"`, will a post be made to 'http://localhost:8084/kapacitor' or to 'http://example.com`?

I can't make this work unfortunately, when defining in the config file:
`[[httppost]]
alert-template = "{{.Message}}"`

I still get a default json format with all the default fields - details, id, etc. If I define both an inline template and a template file in the config file, I do get an error message so the config is loaded correctly, why isn't the custom template used? 

The same is true for a httpPost node, uncommenting `row-template = "{{.Name}} host={{index .Tags \"host\"}}{{range .Values}} {{index . \"time\"}} {{index . \"value\"}}{{end}}"` in the config file doesn't change the format data is posted in. 

Alert is written to a log file, so alert is generated correctly. There is no errors in the kapacitor log.

I'm new to the TICK stack, so I must be missing something, what else can I check? Thanks!

epsylonix:
got it, so basically for a `httpPost()` the config data from the kapacitor's conf file is only used when an `endpoint` property is provided, different endpoints can have different configs, which config is used is determined by the endpoints name.

dishmael:
@epsylonix I wasn't able to get this working properly. If you do, I would be interested in your approach/implementation.

epsylonix:
@dishmael when you do a `post(url)` from `alertNode` or `httpRequest(url)` then the json with default fields is sent (fields are different for post and httpPost). To make use of this format feature you have to 

1. add a `[[httpPost]]` section in the .conf file. `[[` in toml means it's an array of objects, so there can be multiple `[[httpPost]]` sections. In that section you should define a url where data shold be posted, `alert-template` (used for `post()`), or/and `row-template` (for `httpPost()`), and the endpoint property, which is used as a key to reference this endpoint config fro a .tick file.

2. in the .tick file sue `alert().post().endpoint('foo')` or `httpPost().endpoint('foo')`, if you have `endpoint = "foo"` in the .conf file the config for that endpoint will be used, including a template.

By the way I didn't end up using this feature because it is not very flexible  - your template is a string with interpolation, so if you want json, you have to format it manually, and you can't pass random variables there - you have control over `.Message` and `.ID` or you can add tags in the alert data but it is not very convenient.

There is an `exec` property in `alertNode` and you can use any script with it and pass variables there as arguments, the alert data is written to STDIN. In that script it is easy to transform data and send it over http or any over channel in any format you like using you language of choice.

dishmael:
Sadly, I tried all of that and could not get it to send an event.  I did not get anything in the log file that indicated an error associated with the settings so gave up.

plalwa:
can we get a sample exec to pass any additional variables

crashedserver:
@nathanielc A working example would be of great help

caengcjd:
good job

ednjv:
Seems like there's not much documentation around some hidden stuff here, also the default example doesn't work for me, it doesn't output anything for both `row-template` or `row-template-file`

I noticed that there's a custom golang `json` function inside the template, I can't see that being documented anywhere

ashishkaransingh:
I used alert-template-file it works, also i can send hard coded tags (Hard_coded_Tag).
See sample post.html
{
"id": "{{.ID}}",
"message": "{{.Message}}:{{range .Data.Series}}{{.Tags}},{{range .Values}}{{.}}{{end}}{{end}}",
"Hard_coded_Tag": "TEST_007"
}

But i cannot use just {{.Tags}} or {{ index .Fields "value" }}?
Example:
{
"id": "{{.ID}}",
"message": "{{.Message}}:{{.Tags}}",
"Site_ID": "TEST_01"
}
Or
{
"id": "{{.ID}}",
"message": "{{.Tags}}",
"Site_ID": "TEST_01"
}

I get an error "httppost failed to execute alert template"

Any idea why

ednjv:
@ashishkaransingh Alert templates don't have access to .Tags it's probably that, read here https://docs.influxdata.com/kapacitor/v1.5/event_handlers/post/#alert-templates

ashishkaransingh:
@ednjv Testing with row-template
Has anyone got a way out of this?

ednjv:
@ashishkaransingh I have take a look at this https://github.com/influxdata/kapacitor/issues/2149#issuecomment-479462608

ashishkaransingh:
@ednjv  It was promising :) 
I tried this in kapacitor.conf 
[[httppost]]
  endpoint = "testendpoint"
  url = "https://..............................................."
  headers = {Content-Type = "application/json; charset=UTF-8"}
  #alert-template = ""
  #alert-template-file = "/temp/template/post.html"
  row-template = "{name_1: {{.Name}}}"

does not output {{.Name}}?

ednjv:
@ashishkaransingh you need to use double quotes and escape them like this:

row-template = "{\"name_1\": \"{{.Name}}\"}"

ashishkaransingh:
It did not work.
Is it like it only worked for you using kapacitor get/post api?

I used it in kapacitor.conf

[[httppost]]
  endpoint = "mytestendpoint"
  url = "https://testserver.net.................."
  headers = {Content-Type = "application/json; charset=UTF-8"}
  alert-template = ""
  ##alert-template-file = "/temp/template/post.html"
  row-template = "{\"name_1\": \"{{.Name}}\"}"

and Using .tick file where i have defined the endpoint:

var trigger = data
    |alert()
        .post()
        .endpoint('mytestendpoint')
        // .details('Details Removed')
        .crit(lambda: "value" > crit)
        .id(idVar)
        .idTag(idTag)

ashishkaransingh:
@ednjv Unfortunately it did not work either.

ednjv:
@ashishkaransingh wait, you are mixing stuff, you are mixing alert() with row-templates, if you want to use alert you should use alert-template instead of row-template

But if you want to use row-template you can't use alerts

ashishkaransingh:
Ok, i see your point.
So far i am only able to make alert-template-file work.

But anyway it seems it is much easier to use variables inside tickscripts.
I think this depicts the request perfectly https://github.com/influxdata/kapacitor/issues/894

y-ard:
Hi Guys,

Maybe I'll ask strange question, but do you know howto create custom json fields, by extracting value from the '.Data'. It is available as a template, but i don't know howto extract tags' values from it.
Any ideas?

reactcker:
I tried something like the following but cannot access .Tags.
Was able to access {{.Data}} but could not extract key values.
I am sure will not be able to access fields ex: {{index .Fields "value"}}'

Can someone please confirm if we can make it work using alert templates or will this be possible in future with flux etc?

```
{
"id": "{{.ID}}",
"message": "{{.Message}}",
"details": "{{.Details}}",
"time": "{{.Time}}",
"duration": "{{ .Duration}}",
"level": "{{.Level}}",
"recoverable": {{.Recoverable}},
"data": {{ .Data }},
"Instance": "{{.Tags}}"
}

```
I tried this as well: "Instance": "{{index .Tags "instance"}}" same error as the one you here.

received error:

> ts=2020-03-27T21:29:10.006+05:30 lvl=error msg="failed to execute alert template" service=httppost task=http_post_alert_template_test err="template: body:10:15: executing \"body\" at <.Tags>: can't evaluate field Tags in type alert.Data"

ashishkaransingh:
I tried something like the following but cannot access .Tags.
Was able to access {{.Data}} but could not extract key values.
I am sure will not be able to access fields ex: {{index .Fields "value"}}'

Can someone please confirm if we can make it work using alert templates or will this be possible in future with flux etc?

```
{
"id": "{{.ID}}",
"message": "{{.Message}}",
"details": "{{.Details}}",
"time": "{{.Time}}",
"duration": "{{ .Duration}}",
"level": "{{.Level}}",
"recoverable": {{.Recoverable}},
"data": {{ .Data }},
"Instance": "{{.Tags}}"
}

```
I tried this as well: "Instance": "{{index .Tags "instance"}}" same error as the one you here.

received error:

> ts=2020-03-27T21:29:10.006+05:30 lvl=error msg="failed to execute alert template" service=httppost task=http_post_alert_template_test err="template: body:10:15: executing \"body\" at <.Tags>: can't evaluate field Tags in type alert.Data"

desa:
@ashishkaransingh this is the data that is available for the message https://docs.influxdata.com/kapacitor/v1.5/working/alerts/#alert-event-data

ashishkaransingh:
> @ashishkaransingh this is the data that is available for the message https://docs.influxdata.com/kapacitor/v1.5/working/alerts/#alert-event-data

Thank you @desa!

I was looking to access/extract just .Tags or Vars 
It was important for us as we will integrate it with servicenow itom.. any help or if you can point me to golang templating and to the code to achieve this then i can contribute to it...

ashishkaransingh:
> @dishmael when you do a `post(url)` from `alertNode` or `httpRequest(url)` then the json with default fields is sent (fields are different for post and httpPost). To make use of this format feature you have to
> 
> 1. add a `[[httpPost]]` section in the .conf file. `[[` in toml means it's an array of objects, so there can be multiple `[[httpPost]]` sections. In that section you should define a url where data shold be posted, `alert-template` (used for `post()`), or/and `row-template` (for `httpPost()`), and the endpoint property, which is used as a key to reference this endpoint config fro a .tick file.
> 2. in the .tick file sue `alert().post().endpoint('foo')` or `httpPost().endpoint('foo')`, if you have `endpoint = "foo"` in the .conf file the config for that endpoint will be used, including a template.
> 
> By the way I didn't end up using this feature because it is not very flexible - your template is a string with interpolation, so if you want json, you have to format it manually, and you can't pass random variables there - you have control over `.Message` and `.ID` or you can add tags in the alert data but it is not very convenient.
> 
> There is an `exec` property in `alertNode` and you can use any script with it and pass variables there as arguments, the alert data is written to STDIN. In that script it is easy to transform data and send it over http or any over channel in any format you like using you language of choice.

Can someone please provide an example to transform data using exec?

Thanks

ashishkaransingh:
Good Tidings!
For anyone who is interested in accessing .tags for alerting let me know i found a way...

ashishkaransingh:
1)
Use |alert node & httpPost() node in your tickscript Batch/Stream it works with both:
Note you can also add custom tags.. by using default() node.

 |alert()
        .details('details')
        .crit(lambda: "cpuusage" > crit)
        .message(message)
        .id(idVar)
        .idTag(idTag)
        .levelTag('level')
        .messageField(messageField)
        .durationField(durationField)
    |default()
        .tag('folder_name', 'new_sd343f')
        .tag('dashboard_name', 'cpu_ds')
        .tag('metric_name', 'cpuusage')
    |httpPost()
        .endpoint('vnexttopulse')
        .captureResponse()
2)
Use this in your kapacitor.conf

[[httppost]]
  endpoint = "someendpoint"
  url = "http://localhost:8081"
  headers = {Content-Type = "application/json"}
  row-template = '{"records":[{"source": "from_my_app", "event_class": "Kapacitor", "node": "{{index .Tags "host"}}", "type": "{{.Name}}", "metric_name": "{{index .Tags "metric_name"}}", "resource": "{{index .Tags "instance"}}", "additional_info": "Host: {{index .Tags "host"}} Instance: {{index .Tags "instance"}}", "description": "Browse Deeplink: (https://testsite.com/{{index .Tags "folder_name"}}/{{index .Tags "dashboard_name"}}?refresh=1m&orgId=1&var-host={{index .Tags "host"}}) {{range .Values}} Alert_Time: {{index . "time"}} Last_Measure: {{index . "cpuusage"}}" {{end}}}]}'
  [httppost.basic-auth]
    username = "xxxxxxxxxx"
    password = "yyyyyyyyyyyyy"

ibnfalcon:
@ashishkaransingh I am using:

```
|httpPost()
.endpoint('test')

```

And chronograf is telling me that endpoint does not exist even though I have already declared it in the `kapacitor.conf` file. Can you please provide some feedback on this.


Thank you.

reactcker:
@ibnfalcon 
can you can provide your how you got your [[httppost]] set up in kapacitor.conf
And it would be better if you post this to [InfluxData Community](https://community.influxdata.com/) I can answer there as well...

y-ard:
agreed it's very confusing when someone is using chronograf 'httppost' configuration.
because chronograf isn't aware of this configuration in capacitor's config. Probably it should be a feature request for chronograf.

ibnfalcon:
> @ibnfalcon
> What was exact error?
> Did you save and restart kapacitor?
> 
> can you can provide how you got your [[httppost]] set up in kapacitor.conf
> And it would be better if you post this to [InfluxData Community](https://community.influxdata.com/) I can answer there as well...

Eventually I was able to figure it out that if you make any changes to configuration, it will only when work if you restart. So yes, restarting fixed the issue.

Thank you.

reactcker:
Great, You are welcome!

ibnfalcon:
@reactcker 

Do you know if this syntax is supported in `alert-template-file`:

{
"id": "{{.ID}}",
"message": "{{.Message}}",
"time": "{{.Time}}",
`"level": "{{ if eq .Level "OK" }}Up{{ else }}Down{{ end }}",`
"data": {{ .Data }}
}


I am currently getting `error calling eq: incompatible types for comparison` because of `if else` condition for `.Level`.

reactcker:
@ibnfalcon 
Yes,  syntax is supported.
However error points to "level": "{{ if eq .Level "OK" }}Up{{ else }}Down{{ end }}"

try something like this:
{{ if eq (.Level | print) "Warning" }}"3"{{else if eq (.Level | print) "OK"}}"0"{{ end }},

I strongly recommend you to put your topics of such nature on:
https://community.influxdata.com/

PR #1573, opened 2017-09-19
Bootstrap main diagnostic

desa:
Previously if an error was encountered with `cmd.Run`, the diagnostic would be nil and so when `m.Diag.Error` was called it would panic.

This PR now returns an error to the main function.

### DIFF
--- a/cmd/kapacitord/main.go
+++ b/cmd/kapacitord/main.go
@@ -13,6 +13,7 @@ import (
 
 	"github.com/influxdata/kapacitor/cmd/kapacitord/help"
 	"github.com/influxdata/kapacitor/cmd/kapacitord/run"
+	"github.com/influxdata/kapacitor/services/diagnostic"
 )
 
 type Diagnostic run.Diagnostic
@@ -56,6 +57,7 @@ type Main struct {
 // NewMain return a new instance of Main.
 func NewMain() *Main {
 	return &Main{
+		Diag:   diagnostic.BootstrapMainHandler(),
 		Stdin:  os.Stdin,
 		Stdout: os.Stdout,
 		Stderr: os.Stderr,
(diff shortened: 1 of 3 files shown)


cmd/kapacitord/main.go
```diff
@@ -77,15 +78,17 @@ func (m *Main) Run(args ...string) error {
 		cmd.Branch = branch
 
 		err := cmd.Run(args...)
-		// Use logger from cmd since it may have special config now.
-		if cmd.Diag != nil {
-			m.Diag = cmd.Diag
-		}
 		if err != nil {
-			m.Diag.Error("encountered error", err)
```
nathanielc: This is one case were we want to handle the error twice. Returning the error here will just print it to stderr. We also want to log it so the log shows why the startup failed.

desa: The issue is that it's possible that the diagnostic service didn't open. 

```
		dir := path.Dir(s.c.File)
		if _, err := os.Stat(dir); os.IsNotExist(err) {
			err := os.MkdirAll(dir, 0755)
			if err != nil {
				return err
			}
		}


		f, err := os.OpenFile(s.c.File, os.O_CREATE|os.O_WRONLY|os.O_APPEND, 0640)
		if err != nil {
			return err
		}
```
Any thoughts on what to do? Looking back at what was there before I think the same panic was possible, unless I'm missing something.

nathanielc: Previously I believe a dumb logger was always present on the main struct. Could we do something similar?

nathanielc: I checked, previously a new logger was always present on the main struct https://github.com/influxdata/kapacitor/blob/d36f37172ed95ccd37c753f599bdf0942cb94310/cmd/kapacitord/main.go#L59.

desa: Got it. I'll do something similar

desa:
@nathanielc ready for review

PR #1574, opened 2017-09-19
Add support for http status code as a field

### LINKED ISSUE #1569: The HTTP Post node should allow logging on HTTP error codes
Kapacitor should expose information about the response it receives after making the POST request.

* If the HTTP error code is not a 2xx, then the error code should be logged.
* Optionally the entire response body should be logged, if the error code is not 2xx
* Optionally the error code should be placed as a new field on the data.

@desa  Thoughts?

nathanielc:
Fixes #1569

- [x] Rebased/mergable
- [x] Tests pass
- [x] CHANGELOG.md updated

### DIFF
--- a/http_post.go
+++ b/http_post.go
@@ -159,7 +159,6 @@ func (n *HTTPPostNode) doPost(row *models.Row) int {
 }
 
 func (n *HTTPPostNode) postRow(row *models.Row) (*http.Response, error) {
-
 	body := n.bp.Get()
 	defer n.bp.Put(body)
 
--- a/http_post.go
+++ b/http_post.go
@@ -3,14 +3,18 @@ package kapacitor
 import (
 	"encoding/json"
 	"fmt"
+	"io/ioutil"
 	"net/http"
+	"strconv"
 	"sync"
 
 	"github.com/influxdata/kapacitor/bufpool"
 	"github.com/influxdata/kapacitor/edge"
+	"github.com/influxdata/kapacitor/keyvalue"
 	"github.com/influxdata/kapacitor/models"
 	"github.com/influxdata/kapacitor/pipeline"
 	"github.com/influxdata/kapacitor/services/httppost"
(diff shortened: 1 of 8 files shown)


http_post.go
```diff
@@ -108,30 +134,54 @@ func (g *httpPostGroup) DeleteGroup(d edge.DeleteGroupMessage) (edge.Message, er
...
+			var body []byte
+			body, err = ioutil.ReadAll(resp.Body)
+			if err == nil {
+				// Use the body content as the error
+				err = errors.New(string(body))
+			}
+		} else {
+			err = errors.New("unknown error, use .captureResponse() to capture the HTTP response")
+		}
+		n.diag.Error("POST returned non 2xx status code", err, keyvalue.KV("code", strconv.Itoa(resp.StatusCode)))
```
desa: I just now understood what you were saying here https://github.com/influxdata/kapacitor/pull/1535#discussion_r139201996 about making everything strings.

services/httppost/service.go
```diff
@@ -310,5 +316,21 @@ func (h *handler) Handle(event alert.Event) {
...
+			var body []byte
+			body, err = ioutil.ReadAll(resp.Body)
+			if err == nil {
+				// Use the body content as the error
+				err = errors.New(string(body))
+			}
+		} else {
+			err = errors.New("unknown error, use .captureResponse() to capture the HTTP response")
+		}
+		h.diag.WithContext(keyvalue.KV("code", strconv.Itoa(resp.StatusCode))).Error("POST returned non 2xx status code", err)
```
desa: `WithContext` returns a new diagnostic. Might be worth it to change the `Error` method on `Diagnostic` to

```go
Error(msg string, err error, ctx ...keyvalue.T)
```

nathanielc: will do

nathanielc: Fixed

PR #1578, opened 2017-09-20
WIP: Add diagnostic session service

desa:


desa:
@nathanielc I still need to add a few tests, but for the time it should be reviewable at this point.

client/v1/client.go
```diff
@@ -659,6 +660,39 @@ func (c *Client) Do(req *http.Request, result interface{}, codes ...int) (*http.
 	return resp, nil
 }
 
+func (c *Client) Logs(w io.Writer, q map[string]string) error {
```
nathanielc: This really isn't too useful since it never returns, unless there is an error.

We should add a context.Context argument so users can cancel the request if they like.

cmd/kapacitor/main.go
```diff
@@ -2246,3 +2258,48 @@ func doBackup(args []string) error {
...
+	Watch logs associated with a task.
+`
+	fmt.Fprintln(os.Stderr, u)
+}
+
+func doWatch(args []string) error {
+	if len(args) != 1 {
+		return errors.New("must provide task ID.")
+	}
+	err := cli.Logs(os.Stdout, map[string]string{"task": args[0]})
```
nathanielc: Should we setup a signal handler here to cancel the context and cleanly as possible exit?

cmd/kapacitor/main.go
```diff
@@ -2246,3 +2258,48 @@ func doBackup(args []string) error {
 	}
 	return nil
 }
+
+func watchUsage() {
+	var u = `Usage: kapacitor watch <output file>
```
nathanielc: I think writing to Stdout is fine, no need for an output file. Also allowing optional list of key value pairs after the task id would be nice. That way you could filter down by node or whatever within a task easily.

cmd/kapacitor/main.go
```diff
@@ -2246,3 +2258,48 @@ func doBackup(args []string) error {
...
+func doLogs(args []string) error {
+	m := map[string]string{}
+	for _, s := range args {
+		pair := strings.Split(s, "=")
+		if len(pair) != 2 {
+			return fmt.Errorf("bad keyvalue pair: '%v'", s)
+		}
+		m[pair[0]] = pair[1]
+	}
+	err := cli.Logs(os.Stdout, m)
```
nathanielc: Same here I think we want to catch signals and cancel the context.

server/server.go
```diff
@@ -421,6 +423,14 @@ func (s *Server) appendTaskStoreService() {
 	s.AppendService("task_store", srv)
 }
 
+func (s *Server) appendSessionService() {
+	// TODO: add diagnostic
```
nathanielc: The irony :)

services/diagnostic/api.go
```diff
@@ -0,0 +1,95 @@
...
+}
+
+func (s *SessionService) Open() error {
+
+	s.routes = []httpd.Route{
+		{
+			Method:      "GET",
+			Pattern:     sessionsPath,
+			HandlerFunc: s.handleSessions,
+			NoGzip:      true,
```
nathanielc: A comment here explaining why these flag are set would be helpful, since it took us quite  a while to figure out why they are needed.

services/diagnostic/api.go
```diff
@@ -0,0 +1,95 @@
...
+func NewSessionService() *SessionService {
+	return &SessionService{
+		SessionsStore: &sessionsStore{
+			sessions: make(map[uuid.UUID]*Session),
+		},
+	}
+}
+
+func (s *SessionService) Close() error {
+	return nil
```
nathanielc: In close you should remove the routes from the HTTPD service.

services/diagnostic/sessions.go
```diff
@@ -0,0 +1,172 @@
...
+	mu       sync.RWMutex
+	sessions map[uuid.UUID]*Session
+}
+
+func (kv *sessionsStore) Create(w http.ResponseWriter, contentType string, tags []tag) *Session {
+	kv.mu.Lock()
+	defer kv.mu.Unlock()
+
+	flusher, ok := w.(http.Flusher)
+	var wf WriteFlusher = &noopWriteFlusher{w: w}
```
nathanielc: You could put this creation of the noopWriteFlusher in the else block. It would save an allocation :)

services/diagnostic/sessions.go
```diff
@@ -0,0 +1,172 @@
...
+	if ok {
+		wf = &httpWriteFlusher{w: w, f: flusher}
```
nathanielc: You shouldn't need the type httpWriteFlusher, as w is already a writer and you have asserted that its a flusher, so you can now just type assert it to be a WriteFlusher interface. 

```
wf = w.(WriteFlusher)
```

Actually you can skip the direct flusher check and just check it its a `WriteFlusher` to begin with.

cmd/kapacitor/main.go
desa: yup. fixed.

cmd/kapacitor/main.go
desa: typo :) fixed.

client/v1/client.go
desa: fixed.

services/diagnostic/api.go
desa: fixed

services/diagnostic/api.go
desa: fixed.

services/diagnostic/sessions.go
desa: fixed

cmd/kapacitor/main.go
```diff
@@ -2246,3 +2261,81 @@ func doBackup(args []string) error {
...
 }
+
+func watchUsage() {
+	var u = `Usage: kapacitor watch <task id> [<tags> ...]
+
+	Watch logs associated with a task.
+
+	Examples:
+
+		$ kapacitor logs mytask
```
nathanielc: These examples are wrong

cmd/kapacitor/main.go
```diff
@@ -2246,3 +2261,81 @@ func doBackup(args []string) error {
...
+	done := false
+	sigs := make(chan os.Signal, 1)
+	signal.Notify(sigs, syscall.SIGINT)
+	go func() {
+		<-sigs
+		cancel()
+		done = true
+	}()
+
+	if err := cli.Logs(ctx, os.Stdout, m); err != nil && !done {
```
nathanielc: This creates a race on the `done` variable.

cmd/kapacitor/main.go
```diff
@@ -2246,3 +2261,81 @@ func doBackup(args []string) error {
...
+
+	Watch arbitrary kapacitor logs.
+
+		$ kapacitor logs service=http lvl=info
+`
+	fmt.Fprintln(os.Stderr, u)
+}
+
+func doLogs(args []string) error {
+	m := map[string]string{}
```
nathanielc: Should we create a helper function that both doLogs and doWatch use so that this logic is not duplicated?

cmd/kapacitor/main.go
```diff
@@ -2246,3 +2261,81 @@ func doBackup(args []string) error {
...
+	done := false
+	sigs := make(chan os.Signal, 1)
+	signal.Notify(sigs, syscall.SIGINT)
+	go func() {
+		<-sigs
+		cancel()
+		done = true
+	}()
+
+	if err := cli.Logs(ctx, os.Stdout, m); err != nil && !done {
```
nathanielc: Again same race on  `done`.

desa:
@nathanielc made the changes you suggested.

cmd/kapacitor/main.go
desa: fixed

cmd/kapacitor/main.go
desa: fixed

desa:
@nathanielc made changes you suggested

desa:
@nathanielc anything else here that needs to be done?

cmd/kapacitor/main.go
```diff
@@ -2246,3 +2262,80 @@ func doBackup(args []string) error {
...
+	}
+
+	return tailLogs(m)
+}
+
+func tailLogs(m map[string]string) error {
+	ctx, cancel := context.WithCancel(context.Background())
+	done := false
+	sigs := make(chan os.Signal, 1)
+	signal.Notify(sigs, syscall.SIGINT)
```
nathanielc: This should read like this https://github.com/influxdata/kapacitor/blob/master/cmd/kapacitord/main.go#L92

desa: Fixed

desa:
@nathanielc Just rebased. Should be good to merge on your signal.

nathanielc:
@desa The server shutsdown if an invalid TICkscript is provided. We need the soft behavior, i.e. keep running if at all possible. A typo is a script should not cause the server to shutdown.

desa:
@nathanielc Fixed. Removed the logic for hard errors previously, but forgot to ignore the error it returned.

desa:
whoops just realized that I think the previous comment applies to the other PR

nathanielc:
Oops, yeah I commented on the wrong one :-P

nathanielc:
@desa Now actually testing this PR...

How can we make it easy to get logs for all levels greater than info?

nathanielc:
Testing this out I realized we haven't updated the bash completion scripts for the new CLI options.

nathanielc:
I know this isn't directly related but why is starting a task `warn`?

```
ts=2017-10-06T11:06:53.850-06:00 lvl=warn msg="starting task" service=kapacitor task_master=main task=log
```

Looking around I see quite a few things are warns. In general I don't think we need warns. In fact I'd be fine if we removed it entirely. Maybe we should open up a separate PR for remove warns?

desa:
> I know this isn't directly related but why is starting a task warn?

I think this might be something for this PR. In the

desa:
> I know this isn't directly related but why is starting a task warn?

I think this is an issue with the sessions logger. I fixed it. It should be `debug`.

> Maybe we should open up a separate PR for remove warns?

I'll add another PR for removing warns in general.

@nathanielc  added completions.

desa:
@nathanielc just rebased. I think things should be good to go now. (hopefully :))

desa:
@nathanielc fixed issue with `kapacitor logs lvl=error+` not correctly logging levels.

desa:
going to add some tests related to the log level setting

desa:
@nathanielc added test for issue you uncovered

PR #1580, opened 2017-09-21
Add Changelog entry for logging refactor

desa:
Just realized that I forgot to do this :(


desa:
@nathanielc Totally forgot to do this with the PR itself

PR #1415, opened 2017-06-01
Add ProxyFromEnvironment to slack services

desa:
Not sure how I missed this in the first place.

###### Required for all non-trivial PRs
- [ ] Rebased/mergable
- [ ] Tests pass
- [ ] CHANGELOG.md updated
- [ ] Sign [CLA](https://influxdata.com/community/cla/) (if not already signed)

PR #1592, opened 2017-09-28
Added ProxyFromEnvironment support to telegram service

ar7z1:
- [ ] Rebased/mergable
- [ ] Tests pass
- [ ] CHANGELOG.md updated
- [ ] Sign [CLA](https://influxdata.com/community/cla/)

In this PR I've added ProxyFromEnvironment support for Telegram service.

I've used #1415 as an example.

### DIFF
--- a/services/telegram/service.go
+++ b/services/telegram/service.go
@@ -25,6 +25,7 @@ type Diagnostic interface {
 
 type Service struct {
 	configValue atomic.Value
+	clientValue atomic.Value
 	diag        Diagnostic
 }
 
@@ -33,6 +34,11 @@ func NewService(c Config, d Diagnostic) *Service {
 		diag: d,
 	}
 	s.configValue.Store(c)
+	s.clientValue.Store(&http.Client{
+		Transport: &http.Transport{
+			Proxy: http.ProxyFromEnvironment,
+		},
+	})
 	return s
 }
 
@@ -56,6 +62,11 @@ func (s *Service) Update(newConfig []interface{}) error {
 		return fmt.Errorf("expected config object to be of type %T, got %T", c, newConfig[0])
 	} else {
 		s.configValue.Store(c)
+		s.clientValue.Store(&http.Client{
+			Transport: &http.Transport{
+				Proxy: http.ProxyFromEnvironment,
+			},
+		})
 	}
 	return nil
 }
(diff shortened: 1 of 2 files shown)


nathanielc:
@desa  Can you take a look at this one?

services/telegram/service.go
```diff
@@ -33,6 +34,11 @@ func NewService(c Config, d Diagnostic) *Service {
 		diag: d,
 	}
 	s.configValue.Store(c)
+	s.clientValue.Store(&http.Client{
```
desa: any reason why you decided to use `atomic.Value` instead of just using a mutex and a `*http.Client`?

services/telegram/service.go
```diff
@@ -108,7 +119,8 @@ func (s *Service) Alert(chatId, parseMode, message string, disableWebPagePreview
 		return err
 	}
 
-	resp, err := http.Post(url, "application/json", post)
+	client := s.clientValue.Load().(*http.Client)
```
desa: I'd probably just use a mutex, but if you keep with the `atomic.Value` and you do something like

```go
client, ok := s.clientValue.Load().(*http.Client)

if !ok {
    return <some error here>
}
```

nathanielc:
@desa I think the change is fine, the use of atomic.Value works and is aligns with Slack and other similar services. 

What I don't get is why this change is necessary, I thought the default transport and by extension the default client already had the ProxyFromEnvironment set? https://golang.org/pkg/net/http/#RoundTripper

ar7z1:
@nathanielc, @desa wrote in #1238:
>Most services and clients throughout Kapacitor use the Default HTTP
Client which uses the Default HTTP Transport, which uses HTTP proxies as
directed by the $HTTP_PROXY and $NO_PROXY (or $http_proxy and $no_proxy)
environment variables.
>
>For the services that use a custom Transport, we use
http.ProxyFromEnvironment.

I think that's the reason. :-)

services/telegram/service.go
desa: 👍

desa:
@ar7z1 I just looked over this again and @nathanielc is right, I don't think we need this change. that we don't need a custom http client. The default http client https://github.com/influxdata/kapacitor/pull/1592/files#diff-dd32ba460e5730655eef9ac7cb839699L111 should proxy from environment by default.

nathanielc:
Closing, since this should be a non-issue. Please comment otherwise.

Issue #1547, opened 2017-09-05
panic: runtime error: invalid memory address or nil pointer dereference when running a replay when tick script contains post()


pshemk:
Hi,

I was following instructions on [https://docs.influxdata.com/kapacitor/v1.3/introduction/getting_started/](https://docs.influxdata.com/kapacitor/v1.3/introduction/getting_started/)

Kapacitor is running in a container from the official image tagged _:latest_

Tick script i tried to use:
```
root@kapacitor:/var/lib/kapacitor# kapacitor show load_alert
ID: load_alert
Error:
Template:
Type: stream
Status: disabled
Executing: false
Created: 05 Sep 17 00:31 UTC
Modified: 05 Sep 17 00:31 UTC
LastEnabled: 01 Jan 01 00:00 UTC
Databases Retention Policies: ["telegraf"."autogen"]
TICKscript:
stream
    |from()
        .measurement('system')
        .where(lambda: "alerting_load" == 'over3')
    |alert()
        .crit(lambda: "load1" > 0.1)
        .post('http://nagios-api:8080/message')
        .log('/tmp/alerts.log')

DOT:
digraph load_alert {
stream0 -> from1;
from1 -> alert2;
}
```
I recorded the stream:

```
root@kapacitor:/var/lib/kapacitor# kapacitor record stream -task load_alert -duration 20s
e25d8724-82e6-4da9-be6e-6cf6ba797da2
root@kapacitor:/var/lib/kapacitor# rid=e25d8724-82e6-4da9-be6e-6cf6ba797da2
root@kapacitor:/var/lib/kapacitor# kapacitor list recordings $rid
ID                                   Type    Status    Size      Date
e25d8724-82e6-4da9-be6e-6cf6ba797da2 stream  finished  284 B     05 Sep 17 00:41 UTC
```
But when I tried to replay it:

`root@kapacitor:/var/lib/kapacitor# kapacitor replay -recording $rid -task load_alert`

container crashed:
```
[httpd] 127.0.0.1 - - [05/Sep/2017:00:42:19 +0000] "POST /kapacitor/v1/replays HTTP/1.1" 201 258 "-" "KapacitorClient" 1283059a-91d3-11e7-8079-000000000000 11429
[task_master:1e50f514-94da-482c-a39e-a96446a056a2] 2017/09/05 00:42:19 I! opened
[task_master:1e50f514-94da-482c-a39e-a96446a056a2] 2017/09/05 00:42:19 I! closed
[...]

PR #1612, opened 2017-10-11
Fix panic with replaying a recording

desa:
Fixes https://github.com/influxdata/kapacitor/issues/1547

TODO:

- [x] Write Test to verify fix


nathanielc:
@desa Looks good, just add a CHANGELOG.

Issue #1575, opened 2017-09-19
[Feature Request] Persist replay status after it finishes


gpestana:
Based on the [documentation](https://docs.influxdata.com/kapacitor/v1.3/api/api/#get-task), it is possible to access a replay status by providing a _replay-id_ as a query parameter:

GET `/kapacitor/v1/tasks/TASK_ID?replay-id=REPLAY_ID`

However, if the replay has finished, the replay status is lost. When the above endpoint is queried and the replay has finished, the response is the following: `{"error": "no running replay with ID: REPLAY_ID"}`

What do you think about persisting the replay status after its finished? A simple solution would be to store that information in Kapacitor's local database.

For a use case where this feature would be helpful, check [this issue in kapacitor-unit project.](https://github.com/gpestana/kapacitor-unit/issues/10)

PR #1614, opened 2017-10-11
Persists replay data and stats after it finishes [#1575]

gpestana:
## How:

_(edited after e9efc6c commit)_

When a task replay finishes running, the task's stats are stored in the replay.
The user interface for requesting replay data does not change. However, if the user sets the URL query parameter "finished" as true, the finished replay is returned even if the task replay is not running anymore. Assume a replay REPLAY1 has finished:

`GET /kapacitor/v1/replays/REPLAY1`

```
{
  "link": {...},
  "id": "REPLAY1",
  "task": "...",
  "recording": "...",
  "recording-time": false,
  "clock": "fast",
  "date": "2017-10-21T23:41:41.933535592+03:00",
  "error": "",
  "status": "finished",
  "progress": 1
}
```

`GET /kapacitor/v1/replays/<replay-id>?finished=true`

```
{
  "link": {...},
  "id": "REPLAY1",
  "task": "...",
  "recording": "...",
  "recording-time": false,
  "clock": "fast",
  "date": "2017-10-21T23:41:41.933535592+03:00",
  "error": "",
  "status": "finished",
  "progress": 1,
  "stats": {
    "task-stats": {...},
    "node-stats": {...}
  }
}
```
 #
## Breaking changes:

None, since the replay user interface remains the as it is now.

## Todo:

- [x] confirm if design/idea is good
- [x] documentation
- [x] tests

- [x] CHANGELOG.md updated


nathanielc:
@gpestana The description of how you plan to implement this looks good. The PR needs rebasing so that the diff is of only your changes. Once that's fixed I'll take a look at the code. Thanks!

services/replay/service.go
```diff
@@ -967,6 +1000,30 @@ func (s *Service) handleCreateReplay(w http.ResponseWriter, req *http.Request) {
...
+	}
+
+	fr := Replay{
+		ID:            r.ID,
+		RecordingID:   r.RecordingID,
+		TaskID:        r.TaskID,
+		RecordingTime: r.RecordingTime,
+		Clock:         r.Clock,
+		Date:          r.Date,
+		Status:        r.Status,
```
gpestana: `Status` should be `Finished` here. Will updated.

services/replay/service.go
```diff
@@ -83,6 +84,7 @@ type Service struct {
 		Get(string) *kapacitor.TaskMaster
 		Set(*kapacitor.TaskMaster)
 		Delete(*kapacitor.TaskMaster)
+		Main() *kapacitor.TaskMaster
```
gpestana: This is not needed. Will remove.

gpestana:
@nathanielc good call, it should look better now. Added a couple of comments already, so let me know what you think. Thanks!

services/task_store/service.go
```diff
@@ -459,15 +460,49 @@ func (ts *Service) handleTask(w http.ResponseWriter, r *http.Request) {
 		httpd.HttpError(w, fmt.Sprintf("invalid dot-view parameter %q", dotView), true, http.StatusBadRequest)
 		return
 	}
+
 	tmID := r.URL.Query().Get("replay-id")
 	if tmID == "" {
 		tmID = kapacitor.MainTaskMaster
 	}
+
+	finished := r.URL.Query().Get("finished")
```
nathanielc: Why are we making a request to the `/tasks` endpoint to get replay stats? Why not just make a request to `/replays?finished=true` and then return the stats there? 

It means the external system needs to remember which replay-ids map to which tasks but that seems reasonable since the external system made the original request to create a replay with the specified task.

Also not all replays are associated with a task, which is why Kapacitor doesn't store the association itself.

client/v1/client.go
```diff
@@ -600,16 +600,17 @@ type Recording struct {
...
+	ID            string      `json:"id"`
+	Task          string      `json:"task"`
+	Recording     string      `json:"recording"`
+	RecordingTime bool        `json:"recording-time"`
+	Clock         Clock       `json:"clock"`
+	Date          time.Time   `json:"date"`
+	Error         string      `json:"error"`
+	Status        Status      `json:"status"`
+	Progress      float64     `json:"progress"`
+	ReplayStats   interface{} `json:"stats,omitempty"`
```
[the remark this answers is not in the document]
gpestana: No special reason. interface{} was left as it worked and overlooked before the PR. I will fix it.

services/replay/dao.go
```diff
@@ -224,6 +225,8 @@ type Replay struct {
 	Error         string
 	Status        Status
 	Progress      float64
+	// Stores snapshot of finished replayed Task status
+	ReplayStats kapacitor.ExecutionStats
```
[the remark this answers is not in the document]
gpestana: That's a really good point. I will add the ExecutionStats type to the replay package

services/task_store/service.go
gpestana: Yes, it makes sense. I will move the handler code as is under the `replay` package.

services/replay/service.go
```diff
@@ -120,13 +125,21 @@ func (s *Service) Open() error {
...
 	s.StorageService.Register(recordingsAPIName, s.recordings)
+
 	replays, err := newReplayKV(s.StorageService.Store(replayNamespace))
 	if err != nil {
 		return err
 	}
 	s.replays = replays
 	s.StorageService.Register(replaysAPIName, s.replays)
 
+	finishedReplays, err := newReplayKV(s.StorageService.Store(finishedReplaysNamespace))
```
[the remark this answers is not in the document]
gpestana: There is no backwards compatibility issue. Adding the execution stats to the replay KV does not break anything and makes the code simpler. I updated the code accordingly. Thanks for the suggestion!

services/replay/service.go
```diff
@@ -799,6 +830,7 @@ func (s *Service) handleDeleteReplay(w http.ResponseWriter, req *http.Request) {
 	}
 	//TODO: Cancel running replays
 	s.replays.Delete(id)
+	s.finishedReplays.Delete(id)
```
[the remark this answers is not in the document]
gpestana: Removed the need for dedicated KV store for finished replays. This line of code will not be needed anymore.

gpestana:
@nathanielc thanks for the review and suggestions. Updated the code accordingly, as well as the initial commit comment. 

The biggest change is that the execution stats of the replay's task are now stored in the Replay KV instead of having a dedicated KV for finished replays. Also the interface for checking the execution stats when the replay is finished has changed to ´GET /kapacitor/v1/replays/<replay-id>?finished=true´.

services/replay/dao.go
```diff
@@ -214,6 +214,11 @@ const (
 	Real
 )
 
+type ExecutionStats struct {
+	TaskStats map[string]interface{}
+	NodeStats map[string]map[string]interface{}
+}
```
nathanielc: Does this struct need json tags?

services/replay/service.go
```diff
@@ -56,7 +56,7 @@ type Diagnostic interface {
 	Debug(msg string, ctx ...keyvalue.T)
 }
 
-// Handles recording, starting, and waiting on replays
+// Handles recording, starting, and waiting on 	replays
```
nathanielc: Can you fix this whitespace change?

services/replay/service.go
```diff
@@ -780,9 +799,22 @@ func (s *Service) handleReplay(w http.ResponseWriter, req *http.Request) {
 	}
 	replay, err := s.replays.Get(id)
 	if err != nil {
-		httpd.HttpError(w, "could not find replay: "+err.Error(), true, http.StatusNotFound)
+		httpd.HttpError(w, "could not find replay: "+id, true, http.StatusNotFound)
```
nathanielc: Why the change to hide the error?

services/replay/service.go
```diff
@@ -780,9 +799,22 @@ func (s *Service) handleReplay(w http.ResponseWriter, req *http.Request) {
...
+		return
+	}
+
+	finished := req.URL.Query().Get("finished")
+	if finished == "true" && replay.Status != Finished {
+		httpd.HttpError(w, "replay status is not Finished", true, http.StatusBadRequest)
+		return
+	}
```
nathanielc: Since previously the `execution-stats` was not a field of the replay why do we need the `finished` query parameter? Seems like if we simply add the  `execution-stats`  field to the replay response struct we remain backwards compatible while providing the new feature we need.

services/replay/service.go
```diff
@@ -780,9 +799,22 @@ func (s *Service) handleReplay(w http.ResponseWriter, req *http.Request) {
...
+
+	if finished == "true" && replay.Status == Finished {
```
nathanielc: Seems like we don't need this branch, see above comment about always adding the stats to the response.

services/task_store/service.go
```diff
@@ -459,15 +459,19 @@ func (ts *Service) handleTask(w http.ResponseWriter, r *http.Request) {
...
 		tmID = kapacitor.MainTaskMaster
 	}
+
 	tm := ts.TaskMasterLookup.Get(tmID)
+
 	if tm == nil {
 		httpd.HttpError(w, fmt.Sprintf("no running replay with ID: %s", tmID), true, http.StatusBadRequest)
 		return
 	}
+
```
nathanielc: Can you revert these whitespace changes here? They don't seem relevant to the changes.

services/replay/service.go
gpestana: Sure. Will fix all the whitespace and line breaks added.

services/replay/service.go
gpestana: Yep, will remove.

services/replay/service.go
gpestana: Sounds good. The idea was to keep the endpoint response intact and without any changes, but it does make sense to just send the execution stats if they are available.

services/task_store/service.go
gpestana: Sure. Will fix all the whitespace and line breaks added.

services/replay/service.go
```diff
@@ -1000,27 +1000,22 @@ func (s *Service) handleCreateReplay(w http.ResponseWriter, req *http.Request) {
...
+	stats.TaskStats = st.TaskStats
+	stats.NodeStats = st.NodeStats
 
-	err = s.finishedReplays.Create(fr)
+	r.ExecutionStats = stats
+	err = s.replays.Replace(r)
 	if err != nil {
 		return err
 	}
+
```
gpestana: removed

services/replay/service.go
gpestana: The reasoning is that 'could not find replay:  + id' would be enough to understand the problem. But I can add it back  since it may help to debug the cause of the problem.

services/replay/service.go
```diff
@@ -1233,6 +1296,7 @@ func (r *Service) doReplay(id string, task *kapacitor.Task, runReplay func(tm *k
 	if err != nil {
 		return errors.Wrap(err, "task master close")
 	}
+
```
gpestana: removed this line break too

nathanielc:
@gpestana I don't see the changes, you mention. Did you push them up already?

gpestana:
@nathanielc haven't pushed them upstream yet, sorry for the confusion. I've been using the comments as personal Todo points for tracking and to give visibility on what I'm working on. I'm also adding a couple of tests now and updating the docs in addition to the changes mentioned. 

I'll try to push the changes tomorrow or during the weekend he latest.

nathanielc:
@gpestana Ok, great. Thanks again.

client/API.md
```diff
@@ -1114,7 +1114,8 @@ The request returns once the replay is started and provides a replay ID and link
     "recording-time" : false,
     "status" : "running",
     "progress" : 0,
-    "error" : ""
+    "error" : "",
+    "stats": {},
```
nathanielc: No trailing comma in JSON :(

services/replay/service.go
```diff
@@ -766,6 +773,12 @@ func (s *Service) updateReplayResult(replay Replay, err error) {
 	}
 	replay.Progress = 1.0
 	replay.Date = time.Now()
+
+	r, err := s.replays.Get(replay.ID)
+	if err != nil {
+		s.diag.Error("failed to save replay results", err)
+	}
+	replay.ExecutionStats = r.ExecutionStats
```
nathanielc: I think this logic is going to be confusing and cause issues later on.

Looking at the code I think a small refactor is what we need here. 

Let's make the `doReplay` function call the `updateReplayResult` function and have the `updateReplayResult` function take the stats as an argument. Then remove the calls to updateReplayResult outside of `doReplay`.
This will move ownership of updating the replay data in the store to one location so we don't have to worry about overriding previous changes when replacing the replay data.

Thoughts?

client/API.md
gpestana: - [ ] remove trailing comma

gpestana:
@nathanielc I can get these changes in by tomorrow at around 11am UTC+2. If this is too late, please go ahead with the merge and make the necessary changes. Otherwise I can try to finish it up.

Thanks again for the comments!

nathanielc:
That should be fine. Thanks!

services/replay/service.go
```diff
@@ -1215,6 +1229,16 @@ func (r *Service) doReplay(id string, task *kapacitor.Task, runReplay func(tm *k
...
+	stats, err := tm.ExecutionStats(task.ID)
+	if err != nil {
+		return errors.Wrap(err, "getting executing stats replay")
+	}
+
+	// Update replay
+	err = r.updateReplayResult(id, stats, err)
+	if err != nil {
+		return errors.Wrap(err, "update finished replay")
+	}
```
gpestana: I wonder if this is the best way to handle this error after the refactoring. 

The problem might be that if `runReplay` or ` tm.ExecutionStats(task.ID)` calls fail, `updateReplayResult` will never be called and the replay will end up with status Running instead of Failed.

I will refactor the call to r.updateReplayResult as a function and pass it in the previous errors

client/API.md
```diff
@@ -1205,7 +1206,8 @@ All replays are assigned an ID which is returned in this format with a link.
     "recording-time" : false,
     "status" : "running",
     "progress" : 0.57,
-    "error" : ""
+    "error" : "",
+    "stats": {},
```
nathanielc: Trailing comma

client/API.md
```diff
@@ -1248,7 +1250,8 @@ GET /kapacitor/v1/replays/ad95677b-096b-40c8-82a8-912706f41d4c
...
 }
 ```
```
nathanielc: Trailing comma

client/API.md
```diff
@@ -1288,7 +1326,8 @@ GET /kapacitor/v1/replays/ad95677b-096b-40c8-82a8-912706f41d4c
     "recording-time" : false,
     "status" : "failed",
     "progress" : 1,
-    "error" : "error message explaining failure"
+    "error" : "error message explaining failure",
+    "stats": {},
```
nathanielc: Trailing comma

services/replay/service.go
nathanielc: Hmm, yeah this didn't work out very well, sorry about that.

Here is another idea, revert the updateReplayResult changes and just pass a pointer to the Replay to the doReplay function and have it set the stats on the Replay object if it finishes without error. Do not have the doReplay function actually store the replay, but leave that as part of the updateReplayResult function as it was before. This way updateReplayResult doesn't know or care if the stats were set. Does that seem like a better solution?

client/API.md
gpestana: 🤦‍♂️ all trailing commas removed form API.md now.

gpestana:
@nathanielc I reverted the previous changes and removed the trailing commas from the jsons in the API.md.

I tried to implement the latest suggestions but I couldn't make it work properly. Maybe I didn't understand exactly your suggestion. 

I believe the idea is to remove the code that gets the finished replay task's execution stats and stores it from `doReplay`. Apart from the previous commit approach, I tried to add that code to each and every `runReplay` functions. But all the values of the finished replay execution stats would always 0 no matter what data would be replayed.

I could use some some suggestions at this point 😄 Thanks!!

nathanielc:
@gpestana Attached is a quick patch based of your current PR, it compiles but I haven't tested it. Give it a try and see if it works :)


[1614-0.patch.txt](https://github.com/influxdata/kapacitor/files/1438287/1614-0.patch.txt)

gpestana:
@nathanielc did the changes based on your path code, it does seem clearer this way. Does it look enough to merge? Thanks for the help!!

nathanielc:
@gpestana Can you add a CHANGELOG entry? Once that is done I'll merge.

gpestana:
CHANGELOG entry added! 

Looking forward helping on another bug/feature or review. Any suggestion? 😄

gpestana:
ok, gotta resolve the conflict. 1sec

nathanielc:
> Looking forward helping on another bug/feature or review. Any suggestion? 😄

Likewise, there are a few issues that have the label `help-wanted`, but really any issue is free game. If your unit test framework needs any other features that would also be a good place to start.

Thanks again.

PR #1619, opened 2017-10-13
Add example of all nodes in load tasks style dir

desa:
This PR adds simple tickscripts for each node type in kapacitor and will serve as the basis for a burn-in test.

### DIFF
--- a/examples/nodes/tasks/combine.tick
+++ b/examples/nodes/tasks/combine.tick
@@ -1,0 +1,8 @@
+dbrp "telegraf"."autogen"
+
+stream
+    |from()
+        .measurement('cpu')
+        .groupBy(*)
+    |combine(lambda: "cpu" == 'cpu-total', lambda: TRUE)
+        .as('total', 'cpu')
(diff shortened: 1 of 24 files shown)


examples/nodes/tasks/combine.tick
```diff
@@ -0,0 +1,10 @@
+dbrp "telegraf"."autogen"
+
+stream
+    |from()
+        .measurement('system')
+        .groupBy(*)
+    |combine(lambda: isPresent("load1"), lambda: isPresent("uptime"))
```
nathanielc: maybe this one could compare the cpu-total to all other cpus?

```
 |combine(lambda: "cpu' == 'cpu-total', lambda: TRUE)
     .as('total', 'cpu')
```

desa: fixed

desa:
for some reason my new commit isn't showing up here https://github.com/influxdata/kapacitor/compare/md-all-nodes#diff-c69abc1e4574e1e93c4df0d65c4c0ace

PR #1625, opened 2017-10-18
Add JSON Marshaling of Pipeline and AST; Add conversion of Pipeline to AST

goller:
- [x] Rebased/mergable
- [x] Tests pass
- [ ] CHANGELOG.md updated

This PR supports converting a pipeline back into an AST.  ASTs can be transformed into TICKScript. This allows programatic building of Pipelines to generate Tickscripts (pipeline -> AST -> TICKScript)

Additionally, Pipelines are now able to be marshaled to and from JSON.  This will be used by chronograf to take in JSON descriptions of Pipelines and convert them to TICKScripts.


pipeline/tick/from_test.go
```diff
@@ -0,0 +1,67 @@
...
+        .groupByMeasurement()
+        .round(1s)
+        .truncate(1s)
+        .where(lambda: lambda: "cpu" != 'cpu-total' AND lambda: "host" =~ /logger\d+/)
+        .groupBy('this', 'that', 'these', 'those')
+`
+	if got != want {
+		t.Errorf("TestFrom = %v, want %v", got, want)
+		fmt.Println(got) // print is helpful to get the correct format.
+	}
```
nathanielc: I think it would be valuable to wrap this test logic in a helper method. Something like 

```
PipelineTickTestHelper(t *testing.T, pipeline *Pipeline, want string) {
   t.Helper() //mark test as helper so its excluded from the test stack trace.
   got, err := PipelineTick(pipe)
   if err != nil {
       t.Fatalf("Unexpected error building pipeline %v", err)
   }

      if got != want {
         t.Errorf("TestFrom = %v, want %v", got, want)
         fmt.Println(got) // print is helpful to get the correct format.
     }
}
```

nathanielc: Does using `t.Log()` instead of `fmt.Println` here server the same purpose? If so that I recommend we use `t.Log()` as it will ensure the output is correctly organized with the test PASS/FAIL output.

pipeline/window_test.go
```diff
@@ -0,0 +1,141 @@
...
+		t.Run(tt.name, func(t *testing.T) {
+			w := &WindowNode{}
+			err := json.Unmarshal([]byte(tt.input), w)
+			if (err != nil) != tt.wantErr {
+				t.Errorf("WindowNode.UnmarshalJSON() error = %v, wantErr %v", err, tt.wantErr)
+				return
+			}
+			if !reflect.DeepEqual(w, tt.want) {
+				t.Errorf("WindowNode.UnmarshalJSON() =\n%#+v\nwant\n%#+v", w, tt.want)
+			}
```
nathanielc: Could each of these marshaling tests be wrapped into a test helper as well?

goller: Fixed in a4c189c

PR #1637, opened 2017-10-24
Add sideload node and service

nathanielc:
Adds a sideload node that acts similar to the default node but allows the data to be loaded from external sources.

TODO
- [x] Rebased/mergable
- [x] Tests pass
- [x] CHANGELOG.md updated
- [x] Trigger source update over API or signal
- [x] Check only files below the source can be used.
- [x] Cache source objects so they are reused per unique source
- [x] Get feedback on `{tag}` syntax.
- [x] Figure out what to do about conflicting ToRow usage.
- [x] Update API docs
- [ ] Update CLI to trigger a sideload reload (Decided not to do this)


sideload.go
```diff
@@ -0,0 +1,240 @@
...
+	s          *pipeline.SideloadNode
+	source     sideload.Source
+	orderTmpls []orderTmpl
+
+	order []string
+
+	bufferPool *bufpool.Pool
+}
+
+// Create a new  SideloadNode which shifts points and batches in time.
```
desa: > which shifts points and batches in time.

Is this right?

sideload.go
```diff
@@ -0,0 +1,240 @@
...
+func newSideloadNode(et *ExecutingTask, n *pipeline.SideloadNode, d NodeDiagnostic) (*SideloadNode, error) {
+	sn := &SideloadNode{
+		node:       node{Node: n, et: et, diag: d},
+		s:          n,
+		bufferPool: bufpool.New(),
+		order:      make([]string, len(n.OrderList)),
+		orderTmpls: make([]orderTmpl, len(n.OrderList)),
+	}
+	u, err := url.Parse(n.Source)
```
desa: Should lines 40-49 go into the et.tm.SideloadService.Source method?

sideload.go
```diff
@@ -0,0 +1,240 @@
...
+func (n *SideloadNode) doSideload(p edge.FieldsTagsTimeSetter) {
+	for i, o := range n.orderTmpls {
+		p, err := o.Path(p.Tags())
+		if err != nil {
+			n.diag.Error("failed to evaluate order template", err, keyvalue.KV("order", o.raw))
+			return
+		}
+		n.order[i] = p
+	}
+	if len(n.s.Fields) > 0 {
```
desa: Sucks that the logic for fields and tags can't be pulled into a common method without adding reflection.

desa: :(

sideload.go
nathanielc: Nope,:)

sideload.go
nathanielc: Yeah, that would make more sense.

PR #1645, opened 2017-10-30
Stop writing HTTP header for sessions service.

### LINKED ISSUE #1641: Logs API writes multiple http headers
```
curl -X GET  "http://localhost:9092/kapacitor/v1/logs"
ts=2017-10-26T10:49:54.150-05:00 lvl=info msg="created log session" service=sessions id=f4c7ff11-37b5-457c-9e8e-49b1d01c5a6c content-type=
ts=2017-10-26T10:49:54.150-05:00 lvl=error msg="2017/10/26 10:49:54 http: multiple response.WriteHeader calls\n" service=http service=httpd_server_errors
```

desa:
Fixes #1641 

Writing headers is handled by the chunked HTTP response writer. Writing
headers manually results in an HTTP error log.

### DIFF
--- a/services/diagnostic/api.go
+++ b/services/diagnostic/api.go
@@ -104,13 +104,13 @@ func (s *SessionService) handleSessions(w http.ResponseWriter, r *http.Request) 
 
 	contentType := r.Header.Get("Content-Type")
 
-	session := s.SessionsStore.Create(w, contentType, level, tags)
-	defer s.SessionsStore.Delete(session)
-
 	header := w.Header()
 	header.Add("Transfer-Encoding", "chunked")
 	header.Add("Content-Type", r.Header.Get("Content-Type"))
 	w.WriteHeader(http.StatusOK)
+
+	session := s.SessionsStore.Create(w, contentType, level, tags)
+	defer s.SessionsStore.Delete(session)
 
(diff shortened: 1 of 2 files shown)


services/diagnostic/api.go
```diff
@@ -110,7 +110,6 @@ func (s *SessionService) handleSessions(w http.ResponseWriter, r *http.Request)
 	header := w.Header()
 	header.Add("Transfer-Encoding", "chunked")
 	header.Add("Content-Type", r.Header.Get("Content-Type"))
-	w.WriteHeader(http.StatusOK)
```
nathanielc: Hmm, how was this getting called multiple times?

desa: I'm trying to find the exact code, but there's a race between the call to `w.WriteHeader` and the first call to `w.Flush` which checks internally whether headers have been written already or not.

My thoughts were to let the internal HTTP logic handle this instead of writing the header ourselves. Thoughts?

nathanielc: If there is a race with `w.Flush` then there is likely a race with `w.Header()`. We need to fix the race, which probably just means moving lines 107-108 down after anything that interacts with `w`.

PR #1666, opened 2017-11-08
Move logs API underneath /v1preview

desa:


desa:
@nathanielc Do you think this needs a changelog entry?

Issue #1130, opened 2017-01-17
OpsGenie alerts description is not the AlertNode Details


fdhex:
Using the current version of the OpsGenie service for alerting, I assumed that the Description field within the alert created on OpsGenie would be the Details property of the alert node (event.State.Details), but it is in fact hardcoded to event.Data.Result, which is the unmodifiable result data of the InfluxQL query which resulted in the alert's creation.

You can see it there: https://github.com/influxdata/kapacitor/blob/master/services/opsgenie/service.go#L224

It seems more intuitive to me as well as more inline with the other alerting services to link the description of the alert in OpsGenie, which accepts HTML code, to the customizable Details property. A basic implementation can be found here: https://github.com/fdhex/kapacitor/commit/95d60595dea8c524d451ed9f43df6271a57e9b4e

I would propose this implementation to be the new default, but I think it would be important to keep the previous functionality as an optional choice. Pinging the original author of the service @ericiles for opinion here.

nathanielc:
@fdhex Thanks for the detailed write up, I agree leaving the original behavior in tack makes sense as well as enabling the new behavior. I'll defer to @ericiles on further input but the request seems reasonable to me.

ericiles:
I agree that if the details property is used, it should definitely take priority over the current functionality, falling back to the current functionality otherwise. This will allow those that upgrade in the future to not have to modify their OpsGenie rules unless they choose to use the details property to customize what gets sent to OpsGenie.

fdhex:
@ericiles An issue I can see is that Details has a default value (of {{ .json }}), so I'm not sure how to make that transparent so that old rules still get the same output, if they are not modified at all.

ericiles:
@fdhex if that is the case, and the output contains the same information, just in a slightly different format, perhaps this simply needs to be a breaking change, and as such, rules on the OpsGenie side will need to be modified accordingly. Otherwise, the only other solution would be to require an additional property be set on the rule in order to use the details property, which doesn't seem optimal, as it would be completely different from the other services available.Thoughts @nathanielc ?

nutztherookie:
Hi all!

may I inquire to the status here? :)
I agree with @fdhex 's approach but don't understand the issue with Default currently being `{{ .json }}`. This can still be true for OpsGenie if people don't set `|alert().details("...")`. 

Frankly I think people will always prefer having the option to configure the Details rather than just have json in an interface that does not particularly endorse it.

nutztherookie:
[...]

PR #1695, opened 2017-11-27
[services:OpsGenie] changed Description field

andreasnuesslein:
`.opsGenie()` will now use `|alert()`s `.details()` field for the description.
This breaks the old behavior of using `event.Data.Result` (which could not be changed) and draws level with the other alerting services.

Ticket addressing this: https://github.com/influxdata/kapacitor/issues/1130

- [x] Rebased/mergable


nathanielc:
@nutztherookie This is great! Thanks.

I think it can be quite simple to make this not a breaking change. Adding a simple configuration option to the opsgenie config to enable this behavior would work. This way its opt-in to the new behavior but once its turned on, its on for all ops-genie alerts.

See https://github.com/influxdata/kapacitor/pull/1686 for an example of a very similar change to the VictorOps integration.

nutztherookie:
@nathanielc  just so you know; the opsGenie recovery did not work (anymore). At least not for me. When I started looking into it, I realized the service is using the old v1 API; opsGenie has v2 now. 
It's set up a wee bit differently (the way the API-URLs are structured) than e.g. PagerDuty I think.

Long story short: I'm restructuring the file a tad, I hope that's fine with you as long as the result is pretty again? :)

nathanielc:
@nutztherookie Sounds good, maybe we can just set a version option in the config which will toggle all these updates.

nutztherookie:
@nathanielc quick question: there is no way to save a return value in an alert is there?
OpsGenie returns a unique ID for the alert by which it could be addressed later on. It's also fine to use the alias but the other way would be clearer.

nutztherookie:
@nathanielc 
- I didn't include the returned ID for now and I'm using the `alias` for everything. 
- I did also not do the config-version-thingy. The only reason for that would be to keep the .json-output-compliance, which I would rather integrate if really necessary. I don't think anybody would benefit from just keeping the old (deprecated) API.
- The new API changed the authentication-method: Instead of the api-key being in the payload, it's now a header `Authorization: GenieKey eb243592-faa2-4ba2-a551q-1afdf565c889` (https://docs.opsgenie.com/v1.0/docs/alert-api)  .. I updated most tests accordingly but the one remaining I can't figure out and was hoping you could help? :) 

thanks

nutztherookie:
@nathanielc 
I found the last bug, :)
please let me know if you want a backwards compatibility flag for the old `.details()` behavior.

happy new year!

fdhex:
@nutztherookie thanks for your work! Looking forward to have this merged.

yonido:
Can we expedite this PR for the next release? this is very important : ) +1 !

nathanielc:
@andreasnuesslein @yonido We have added an opsgenie2 implementation in this PR #1823 which should address most of these issues and uses the new v2 API. Closing this PR to keep the old handler backwards compatible.

nathanielc:
Further explanation of why this PR was closed since it was not quite clear with my initial comment. 

This PR represents a set of new features added to the OpsGenie alert integration for Kapacitor. These change are backwards incompatible with the current integration. In order to accept these changes we need to apply them to a new handler. The work in #1823 set the ground work for these new changes, primarily it created a new OpsGenie handler called opsgenie2 and it uses the v2 API.

The next steps are to apply the changes this PR made to the original opsgenie code to the new opsgenie2 code instead.

abadyan-vonage:
Guys, is there a plan to merge these sweet improvements to opsgenie2?
@andreasnuesslein @nathanielc

andreasnuesslein:
@abadyan-vonage sorry not from my side

fdhex:
@nathanielc I was pretty happy to see my code suggestions from back then were going to be merged as we're still using them in prod! Unfortunate that we are stuck with the previous behavior...

Issue #1697, opened 2017-11-28
kapacitord config command panics


karel-rehor:
[kapacitor.conf.txt](https://github.com/influxdata/kapacitor/files/1509385/kapacitor.conf.txt)
In cmd/kapacitord/main.go:132 it would appear that the command 'config' should print out a configuration to the console, as is the case with influxd and telegraf.  However, when I try this on the latest nightly build, it panics as follows. 

panic: unexpected reflect.Kind: func [recovered]

Console messages from latest attempts:
```
karl@algonquin:~/Downloads/influxdata$ kapacitord version
Kapacitor 1.5.0~n201711280812 (git: master a1a60f2e560fd89f011aabc0e5eb82e277b668a4)
karl@algonquin:~/Downloads/influxdata$ kapacitord config
2017/11/28 10:55:54 Merging with configuration at: /etc/kapacitor/kapacitor.conf
panic: unexpected reflect.Kind: func [recovered]
    panic: unexpected reflect.Kind: func

goroutine 1 [running]:
github.com/influxdata/kapacitor/vendor/github.com/BurntSushi/toml.(*Encoder).safeEncode.func1(0xc420641ba0)
    /root/go/src/github.com/influxdata/kapacitor/vendor/github.com/BurntSushi/toml/encode.go:101 +0xa9
panic(0x1d00b40, 0xc420537a40)
    /usr/local/go/src/runtime/panic.go:491 +0x283
github.com/influxdata/kapacitor/vendor/github.com/BurntSushi/toml.tomlTypeOfGo(0x1dac900, 0xc420337e98, 0x193, 0xb, 0x1dac900)
    /root/go/src/github.com/influxdata/kapacitor/vendor/github.com/BurntSushi/toml/encode.go:420 +0x547
github.com/influxdata/kapacitor/vendor/github.com/BurntSushi/toml.(*Encoder).eStruct.func1(0x33ac5e0, 0x20c1980, 0x20c1980, 0xc420337e00, 0x199, 0x0, 0x0, 0x0)
    /root/go/src/github.com/influxdata/kapacitor/vendor/github.com/BurntSushi/toml/encode.go:338 +0x168
github.com/influxdata/kapacitor/vendor/github.com/BurntSushi/toml.(*Encoder).eStruct(0xc420641cb0, 0xc4205377a0, 0x1, 0x1, 0x20c1980, 0xc420337e00, 0x199)
[...]

PR #1698, opened 2017-11-28
Fix panic with MQTT toml configuration generation

desa:
Previously, the `NewClientF` being exported was causing the our toml
package to panic.

Fixes https://github.com/influxdata/kapacitor/issues/1697

- [x] Rebased/mergable
- [x] Tests pass
- [x] CHANGELOG.md updated

### DIFF
--- a/services/mqtt/config.go
+++ b/services/mqtt/config.go
@@ -29,8 +29,15 @@ type Config struct {
 	Username string `toml:"username" override:"username"`
 	Password string `toml:"password" override:"password,redact"`
 
-	// NewClientF is a function that returns a client for a given config.
-	NewClientF func(c Config) (Client, error) `toml:"-" override:"-"`
+	// newClientF is a function that returns a client for a given config.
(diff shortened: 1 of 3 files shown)


services/mqtt/config.go
```diff
@@ -29,8 +29,15 @@ type Config struct {
...
-	// NewClientF is a function that returns a client for a given config.
-	NewClientF func(c Config) (Client, error) `toml:"-" override:"-"`
+	// newClientF is a function that returns a client for a given config.
+	// It is used exclusively for testing.
+	newClientF func(c Config) (Client, error) `override:"-"`
+}
+
+// SetNewClientF sets the newClientF on a Config.
+// It is used exclusively for testing.
+func SetNewClientF(cfg *Config, fn func(c Config) (Client, error)) {
```
nathanielc: Why not make this a method of the `Config` type? Seems like it would be easier to find that way.

desa:
@nathanielc should be good to go.

PR #1703, opened 2017-12-01
Fix log API accept <-> content-type header issue

nathanielc:
- [x] Rebased/mergable
- [x] Tests pass
- [x] CHANGELOG.md updated


nathanielc:
@desa We forgot to change this. I just remembered so I put up this PR. 

Although this probably means we just broke Chronograf.

@cryptoquick Does Chronograf set the `Accept` header to get the logs back in JSON? If not it will need to with this change.

cryptoquick:
@nathanielc Noted. Will submit a PR for this change. Hopefully adding this header won't break the current version of Kapacitor.

PR #1715, opened 2017-12-06
Remove references to actions from API docs

### LINKED ISSUE #1713: Update Alerts API docs
The API docs still refer to the `actions` concept of alert handlers which was removed in  v1.3 we need to update the docs.

All references to `actions` need to be updated.

desa:
Fix #1713


nathanielc:
@desa Can you create the docs PR to update the API with this new version?

PR #1719, opened 2017-12-07
Fix oddly generated TOML for mqtt & httppost

desa:
Previously `kapacitord config` generated TOML that looked like

```toml
httppost = []
mtqq = []
...
```

rather than

```toml
[[httppost]]
  ...

[[mqtt]]
  ...
```

this was due to `server.NewConfig` returning an empty slice of `Config`s
for httppost and mqtt.

- [ ] Tests pass
- [ ] CHANGELOG.md updated

### DIFF
--- a/services/httppost/config.go
+++ b/services/httppost/config.go
@@ -41,6 +41,13 @@ type Config struct {
 	RowTemplateFile   string            `toml:"row-template-file" override:"row-template-file"`
 }
 
+func NewConfig() Config {
+	return Config{
+		Endpoint: "example",
+		URL:      "http://example.com",
+	}
+}
+
 // Validate ensures that all configurations options are valid. The Endpoint,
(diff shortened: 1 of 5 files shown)


services/httppost/config.go
```diff
@@ -41,6 +41,13 @@ type Config struct {
 	RowTemplateFile   string            `toml:"row-template-file" override:"row-template-file"`
 }
 
+func NewConfig() Config {
+	return Config{
+		Endpoint: "default",
+		URL:      "http://localhost:9092",
```
nathanielc: This is a potentially confusing default. Maybe http://example.com?

nathanielc:
@desa Looks good. Can you add a CHANGELOG entry?

desa:
@nathanielc should be good to go now.

PR #1750, opened 2018-01-10
Update agent module for py3 encoding issues

rca:
When using stdin and stdout use the buffer attribute in order to be able
to send binary strings rather than text.

- [ ] Rebased/mergable
- [ ] Tests pass
- [ ] CHANGELOG.md updated

### DIFF
--- a/udf/agent/py/kapacitor/udf/agent.py
+++ b/udf/agent/py/kapacitor/udf/agent.py
@@ -12,6 +12,10 @@ try:
 except ImportError:
     from Queue import Queue
 
+# Check for python3
+# https://stackoverflow.com/a/38939320/703144
+PY3K = sys.version_info >= (3, 0)
+
 import io
 import traceback
 import socket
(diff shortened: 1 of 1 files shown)


nathanielc:
@rca Thanks for the PR, look good. I'll test this out locally in a bit.

rca:
Yep, no problem.  As mentioned in #1355, I still can't send out data points, but perhaps it's an issue with my Kapacitor config?  I noticed my `[[influxdb]]` configuration was pointing to localhost, but influx is not running alongside kapacitor (docker container).

shantanoo-desai:
# Local Check

@rca 's patch in `agent.py` works for me on a local Windows 10 machine for a custom UDF. Refer to #2170

shantanoo-desai:
As a Proof of Concept I have documented the complete process on a [Repository](https://github.com/shantanoo-desai/kapacitor-udf-python) @rca

N-Coder:
This works for me under Python 3.6.8.
When merging this (which already includes #1355), you'll probably want to also include the `from __future__ import absolute_import` line from #1799. That PR is otherwise identical to #1355, but this import ensures backwards compatibility to python 2.7 (which will [retire soon](https://pythonclock.org/), anyways).

PR #1778, opened 2018-01-27
Remove reference to WARN log level

gunnaraasen:
- [x] Rebased/mergable
- [x] Tests pass
- [x] CHANGELOG.md updated

The `WARN` log level was not being used and removed in the process of adding the logging API features in Kapacitor 1.4.0. Note that it is still possible to set warn thresholds on alert nodes, this only affects Kapacitor internal logging.


rbetts:
There is similar code in fermium that would need a corresponding PR (if one hasn't been made already).

PR #1795, opened 2018-02-07
Support PagerDuty API v2

sebito91:
- [x] Rebased/mergable
- [x] Tests pass
- [x] CHANGELOG.md updated

The PagerDuty API has stepped to v2 and it's a pretty serious refactor of the endpoint. While the v1 endpoint is still supported for some time, it's in deprecated mode and will ultimately be replaced soon(ish).

https://v2.developer.pagerduty.com/docs/events-api-v2
https://v2.developer.pagerduty.com/docs/getting-started

### DIFF
--- a/server/server_test.go
+++ b/server/server_test.go
@@ -22,6 +22,7 @@ import (
 	"testing"
 	"time"
 
+	"github.com/davecgh/go-spew/spew"
 	"github.com/dgrijalva/jwt-go"
 	iclient "github.com/influxdata/influxdb/client/v2"
 	"github.com/influxdata/influxdb/influxql"
@@ -46,6 +47,8 @@ import (
 	"github.com/influxdata/kapacitor/services/opsgenie2/opsgenie2test"
 	"github.com/influxdata/kapacitor/services/pagerduty"
 	"github.com/influxdata/kapacitor/services/pagerduty/pagerdutytest"
+	"github.com/influxdata/kapacitor/services/pagerduty2"
+	"github.com/influxdata/kapacitor/services/pagerduty2/pagerduty2test"
 	"github.com/influxdata/kapacitor/services/pushover/pushovertest"
 	"github.com/influxdata/kapacitor/services/sensu/sensutest"
 	"github.com/influxdata/kapacitor/services/slack/slacktest"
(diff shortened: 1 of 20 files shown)


services/pagerduty2/service.go
```diff
@@ -0,0 +1,328 @@
...
+// API entry point is now https://events.pagerduty.com/v2/enqueue
+type PDCEF struct {
+	Summary       string
+	Source        string
+	Severity      string
+	Timestamp     time.Time
+	Class         string
+	Component     string
+	Group         string
+	CustomDetails string
```
nathanielc: Based on the above it looks like this field can be arbitrary JSON, is that correct? If so the type could be `map[string]interface{}` or possibly just `interface{} `something and we could add the `details` JSON as this field.

See the VictorOps handler for how the details can be treated as rich data instead of just a string https://github.com/influxdata/kapacitor/blob/master/services/victorops/service.go#L208.  The `models.Result` type support JSON marshalling so it possibly could be set as the CustomDetails directly.

services/pagerduty2/service.go
```diff
@@ -0,0 +1,328 @@
...
+	eventType := "trigger"
+
+	switch level {
+	case alert.Warning:
+		severity = "warning"
+	case alert.Critical:
+		severity = "critical"
+	case alert.Info:
+		severity = "info"
+	case alert.Error:
```
nathanielc: How does the alert.Error level get set?

sebito91:
Thanks for the feedback! Still working to get the tests to pass fully before checking once more (comments in-line).

services/pagerduty2/service.go
sebito91: Was following the existing `pagerduty` description which actually has this as a standard string. A local commit type set `CustomDetails` to the `map[string]interface{}` you called out because that does seem to be the most flexible.

https://github.com/influxdata/kapacitor/blob/master/services/pagerduty/service.go#L155

Do we want to change this one too?

services/pagerduty2/service.go
sebito91: Good catch, this is my reading of the API v2 docs and not the internal representation from the `kapacitor` side...I'll remove.

sebito91:
@nathanielc mind taking another look? I've incorporated your feedback (hopefully)...

sebito91:
Not sure where this test is failing, but still looking into the fix for this one...
```
8432 === RUN   TestService_GetConfig
8433 --- FAIL: TestService_GetConfig (0.04s)
8434     service_test.go:1076: expected to get an response on errC
8435 FAIL
8436 FAIL    github.com/influxdata/kapacitor/services/config 1.072s
```

services/pagerduty2/service.go
```diff
@@ -0,0 +1,388 @@
...
+	} else {
+		ap.RoutingKey = serviceKey
+	}
+
+	ap.Client = "kapacitor"
+	ap.ClientURL = s.HTTPDService.URL()
+	ap.DedupKey = incidentKey
+	ap.EventAction = eventType
+
+	ap.Payload.CustomDetails = make(map[string]interface{})
```
nathanielc: I think it would be simpler and still just as useful if the CustomDetails where just the data.Result object. The Result object serializes to JSON in a standard format and contains all relevant data to the alert. Will it still work for your use case to just ignore the `details` string and pass in the Result directly? This way we can avoid any edge cases with trying to unmarshal the details strings since its opaque to the code at this point. Thoughts?

To be clear I am thinking something like this

```
ap.Payload.CustomDetails = make(map[string]interface{})
ap.Payload.CustomDetails["result"] = data.Result
```

Since the details strings can only be constructed using the data already in data.Result, then this is a superset of the data being provided. The only question is if the data is still useful if we do it this way.

services/pagerduty2/service.go
```diff
@@ -0,0 +1,388 @@
...
+}
+
+// Handle is a bound method to the handler that processes a given alert
+func (h *handler) Handle(event alert.Event) {
+	if err := h.s.Alert(
+		h.c.ServiceKey,
+		event.State.ID,
+		event.State.Message,
+		event.State.Level,
+		event.State.Details,
```
nathanielc: As per the above comment we can probably just remove passing the details down if they are not going to be used.

services/pagerduty2/service.go
sebito91: Ah, that makes sense. WIll update and get back to you ASAP!

server/server_test.go
```diff
@@ -8986,6 +9086,62 @@ func TestServer_AlertHandlers(t *testing.T) {
...
+								},
+							},
+							Timestamp: "1970-01-01T00:00:00.000000001Z",
+						},
+						RoutingKey: "service_key",
+					},
+				}}
+
+				fmt.Printf("DEBUG -- got: %v\n", got[0].PostData.Payload)
+				fmt.Printf("DEBUG -- exp: %v\n", exp[0].PostData.Payload)
```
nathanielc: Can you make these either `t.Log` or remove them?

sebito91:
@nathanielc apologies for the noise on this PR but I believe I've finally covered the updates you were looking for. Thanks again for your patience!

alert/types.go
```diff
@@ -125,6 +125,7 @@ const (
 	Warning
 	Critical
 	maxLevel
+	Error
```
nathanielc: This is unused correct?

services/pagerduty2/service.go
```diff
@@ -0,0 +1,378 @@
...
+	ap.Payload.CustomDetails = make(map[string]interface{})
+	ap.Payload.CustomDetails["result"] = data.Result
+
+	// The API doesn't explicitly mention a requirement for nanosecond resolution but payloads seem to
+	// fail if we don't include it (even zeroes). This hack is not graceful, but adds a negligible
+	// nanosecond resolution to our timestamp
+	m := timestamp.Format(time.RFC3339Nano)
+	if timestamp.Nanosecond() == 0 {
+		m = time.Unix(timestamp.Unix(), 1).In(timestamp.Location()).Format(time.RFC3339Nano)
+	}
```
nathanielc: This is just trying to ensure that the timestamps are correctly padded with zeros? And it does so by adding a single nanosecond? If so there is another way to do this. Go's time formatting docs don't make it super clear but in the format string if you use the digit `0` as a placeholder then it will correctly pad the string to be a fixed width.

From the Go source:

```
RFC3339Nano = "2006-01-02T15:04:05.999999999Z07:00"
                            
```
If you use this format instead, then zero padding will be added.

```
"2006-01-02T15:04:05.000000000Z07:00"
```

See https://play.golang.org/p/87-3z6vnUnc

alert/types.go
sebito91: I'll remove this one for now, but it is technically an option for the v2 API.

https://v2.developer.pagerduty.com/docs/send-an-event-events-api-v2

Since I'm not explicitly using it in the service then it's fine to ignore.

nathanielc: Yeah, eventually Kapacitor may support custom alert levels, right now it tries to keep things simple by implementing what is common between most alert services. 

Once there is room for customization here, we can enable the pagerduty2 to take advantage of it.

Thanks.

sebito91:
@nathanielc feedback incorporated, PR rebased and squashed!

PR #1818, opened 2018-02-27
Added oss to the appropriate locations to identify the oss platform

aanthony1243:
- [X] Rebased/mergable
- [X] Tests pass
- [ ] CHANGELOG.md updated  ??

### DIFF
--- a/cmd/kapacitord/main.go
+++ b/cmd/kapacitord/main.go
@@ -198,7 +198,7 @@ func (cmd *VersionCommand) Run(args ...string) error {
 	}
 
 	// Print version info.
-	fmt.Fprintf(cmd.Stdout, "Kapacitor %s (git: %s %s)\n", version, branch, commit)
+	fmt.Fprintf(cmd.Stdout, "Kapacitor oss version %s (git: %s %s)\n", version, branch, commit)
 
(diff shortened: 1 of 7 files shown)


server/vars/vars.go
```diff
@@ -23,7 +23,7 @@ const (
 	UptimeVarName = "uptime"
 
 	// The name of the product
-	Product = "kapacitor"
+	Product = "kapacitor oss"
```
nathanielc: This will break things. the product is used when making calls to various third parties, if we change the name we risk breaking their integrations.

cmd/kapacitord/main.go
```diff
@@ -198,7 +198,7 @@ func (cmd *VersionCommand) Run(args ...string) error {
 	}
 
 	// Print version info.
-	fmt.Fprintf(cmd.Stdout, "Kapacitor %s (git: %s %s)\n", version, branch, commit)
+	fmt.Fprintf(cmd.Stdout, "Kapacitor oss version %s (git: %s %s)\n", version, branch, commit)
```
nathanielc: Should oss be capitalized since its an acronym OSS?

PR #1823, opened 2018-02-28
Update OpsGenie to use v2 API

### LINKED ISSUE #1708: OpsGenie API v1 is deprecated: support non-deprecated API
{"code":32,"error":"DeprecatedApiException[API v1 is deprecated. Please refer https://docs.opsgenie.com/docs/dashboard/rest-api for details of the new version]"}

### LINKED ISSUE #1819: OpsGenie v2 Support
Comparing https://docs.opsgenie.com/docs/alert-api to https://docs.opsgenie.com/docs/alert-api-deprecated it looks like there are a few difference:

* The authentication method changes from a property of the JSON body to an HTTP header
* The team/recipients properties are combined into a respondents  list. Teams are `{"name":"team name","type":"team"}` and recipients are `{"username": "user name","type":"user"}`.

At this point I believe we can switch to using the new API without needing to create a new config or handler. Simply updating the existing code should be enough.

### LINKED ISSUE #1821: OpsGenie - Report tags as extra properties (Feature request)
Currently the only extra fields reported to opsgenie are level and monitoring tool (which is always Kapacitor). All other data is reported as a big a$$ json. OpsGenie doesn't support searching inside this json so that information has very limited usability.
Looking at the code, it doesn't look very hard to dynamically turn all tags of the alerts into extra properties, making it much easier to use. I can help with a pull request if you think this is a good idea.

nathanielc:
Fixes #1821 
Fixes #1819 
Fixes #1708 

- [x] Rebased/mergable
- [x] Tests pass
- [x] CHANGELOG.md updated

TODO
- [x] The RecoveryURL allowed user to customize recovery behavior, not sure how to update this in a non breaking way.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,10 +2,16 @@
 
 ## v1.5.0 [unreleased]
 
+### Features
+
+- [#1823](https://github.com/influxdata/kapacitor/pull/1823): Update OpsGenie integration to use the v2 API.
+    To upgrade to using the new API simply update your config and TICKscripts to use opsgenie2 instead of opsgenie.
+    If your `opsgenie` config uses the `recovery_url` option, for `opsgenie2` you will need to change it to the `recovery_action` option.
+    This is because the new v2 API is not structured with static URLs, and so only the action can be defined and not the entire URL.
+- [#1690](https://github.com/influxdata/kapacitor/issues/1690): Add https-private-key option to httpd config.
+
 ### Bugfixes
 
-### Features
-- [#1690](https://github.com/influxdata/kapacitor/issues/1690): Add https-private-key option to httpd config.
 
 ## v1.4.0 [2017-12-08]
 
--- a/alert.go
+++ b/alert.go
@@ -21,6 +21,7 @@ import (
 	"github.com/influxdata/kapacitor/services/httppost"
 	"github.com/influxdata/kapacitor/services/mqtt"
 	"github.com/influxdata/kapacitor/services/opsgenie"
+	"github.com/influxdata/kapacitor/services/opsgenie2"
 	"github.com/influxdata/kapacitor/services/pagerduty"
 	"github.com/influxdata/kapacitor/services/pushover"
 	"github.com/influxdata/kapacitor/services/sensu"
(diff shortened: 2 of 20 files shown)


services/opsgenie/service.go
```diff
@@ -122,71 +133,111 @@ func (s *Service) Alert(teams []string, recipients []string, messageType, messag
...
-	ogData["note"] = ""
-	ogData["monitoring_tool"] = "kapacitor"
+	u := c.URL
+	switch level {
+	case alert.OK:
+		up, err := url.Parse(c.URL)
+		if err != nil {
+			return nil, err
+		}
+		up.Path = path.Join(up.Path, entityID, "close")
```
abadyan-vonage: does this mean that close will always be the behavior on alert.OK? I think in the current API you can set whether this closes the alert or does some other action.

nathanielc: No, just a first pass implementation. See my note above about how to make the change be backwards compatible since the user can no longer just specify the url directly as the URL needs to be constructed dynamically.

abadyan-vonage: cool, thanks.

nathanielc:
OK, this should be ready to go.

I ended up just using the recovery_url last path element to indicate which kind of action should be taken on recovery. This is an odd way to do the configuration but it works in a backwards compatible way.

Also I discovered that the OpsGenie v2 API has a regression where if your entity IDs contains any slashes `/`, with their new API its now impossible to close or add notes to that alert.

nathanielc:
I found a way around the `/` issue. At this point the change is backwards compatible.

yonido:
@nathanielc Thanks. Two things:
1. In https://github.com/influxdata/kapacitor/pull/1695 nutztherookie added some very nice things, such as updating an existing alert with its new priority to support escalated alerts, why aren't these included?

2. If you're at it, can you PLEASE add support for multi apikey integrations? 
There's an issue with escalating alerts (warn>critical) under the same "alias" -
I'm in deep touch with the Opsgenie team and they said this the right way to tackle this is to have multiple integrations (multi api-keys), each one handling a different level.

nathanielc:
@yonido @andreasnuesslein We will still accept those changes once they are updated. The #1695 PR could not be accepted as it was because it would represent a backwards incompatible change. We have solved that part of the problem by creating the opsgenie2 handler.  At this point if either @yonido or @andreasnuesslein (or anyone else) wants to create a new PR with the changes from #1695 but applied against the opsgenie2 handler code we would gladly accept the changes. 

I probably should have communicated that better when I closed the PR, I'll post another comment on that PR to help clear that up.

yonido:
@nathanielc Thanks. What about multi apikey integration? ie - sending the same alert to multiple OpsGenie "accounts"? (with support via 'topics)

yonido:
@nathanielc ?

bogski87:
Have any of the changes discussed been implemented? Because i can tell for sure that the alert nodes .details() does NOT populate the description field. 

Using "details" in OpsGenie itself adds "extraProperties" which are great but contain none of the fields from the alert data, just tags. 

Also, none of those extra properties are sent in the "close" message, so building an alias  using the information contained in extra properties fails. When the alert closes the alias is not matched because the data from teh tags isn't included.

PR #1825, opened 2018-03-01
define properly assigns task name 

aanthony1243:
- [X] Rebased/mergable
- [X] Tests pass
- [X] CHANGELOG.md updated

### DIFF
--- a/cmd/kapacitor/main.go
+++ b/cmd/kapacitor/main.go
@@ -661,14 +661,15 @@ Options:
 }
 
 func doDefine(args []string) error {
-	defineFlags.Parse(args)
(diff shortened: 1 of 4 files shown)


cmd/kapacitor/main.go
```diff
@@ -661,11 +661,13 @@ Options:
 }
 
 func doDefine(args []string) error {
```
nathanielc: Can you update the usage comment for the define command to indicate that the id is optional if a `-file` is provided?

aanthony1243: it's not optional. we agreed that it should match what's on the cmd line.  I'll update in that context, though.

PR #1831, opened 2018-03-06
added a health check path

### LINKED ISSUE #1830: Add a health check endpoint
Hi,

We are using Kapacitor on an AWS environment with deployment tooling that puts our Kapacitor instance behind a load balancer (this makes dns settings and deployments much easier, as well as instance resilience).
One of the derivatives of this is that Kapacitor is being pinged with a health check on the root path.
This returns a 404 and makes the health check fail.
There's no documented routes but looking at `handler.go` I have indeed found the routes that the built-in httpd uses.
Would be nice to have a `/health-check` endpoint that returns an HTTP `200` (AWS's health check expect a `200` for a healthy instance).

justelad:
- [X] Rebased/mergable
- [ ] Tests pass
- [x] CHANGELOG.md updated

resolves #1830


nathanielc:
@elad-nach Thanks for the PR! Kapacitor already has a health check endpoint `/ping`. Since the AWS ELB requires the response to be a 200 OK. I think the best change is to have the `/ping` endpoint return a 200 OK instead of a 204 No Content. That way the ELB will still work and we don't have two endpoints for the same purpose.

Can you update this PR to modify the `/ping` endpoint instead of adding a new endpoint?

elad-nach:
@nathanielc thanks for the quick review. I didn't touch the `/ping` endpoint due to the assumption that this endpoint is already being used, maybe by influx or other daemons on the stack and this might turn out to be a breaking change (if someone's using it - they're probably also asserting the `204` and the assertion would now break).
Will adjust the PR nevertheless.

justelad:
closing this. not sure why the tests are still failing after my adjustments but the overhead of dealing with this is already past my expectations

PR #1833, opened 2018-03-06
[WIP] slack multi-config

aanthony1243:
Needs new tests, but all existing tests are passing with a multi-config instead of a single element config for slack. 

- [ ] Rebased/mergable
- [ ] Tests pass
- [ ] CHANGELOG.md updated

### DIFF
--- a/pipeline/alert.go
+++ b/pipeline/alert.go
@@ -1256,7 +1256,7 @@ type PushoverHandler struct {
 //         |alert()
 //             .slack()
 //
-// Send alerts to Slack channel in the configuration file.
+// Send alerts to the default worskace Slack channel in the configuration file.
 //
 // Example:
 //    stream
@@ -1264,7 +1264,7 @@ type PushoverHandler struct {
 //             .slack()
 //             .channel('#alerts')
 //
-// Send alerts to Slack channel '#alerts'
+// Send alerts to the default workspace with Slack channel '#alerts'
 //
 // Example:
 //    stream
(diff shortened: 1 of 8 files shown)


pipeline/alert.go
```diff
@@ -1337,13 +1337,23 @@ type PushoverHandler struct {
 //
 // Send alert to user '@jsmith'
 //
+// Example:
+// stream
+//      |alert()
+//          .slack('opencommunity')
```
nathanielc: This example should be:
```
    .slack()
         .workspace('opencommunity')
```

services/slack/config.go
```diff
@@ -52,3 +57,37 @@ func (c Config) Validate() error {
 	}
 	return nil
 }
+
```
nathanielc: The NewConfig() method is now not sufficient to set defaults.

 Because its possible that there are multiple config structs created, the toml parser will create a new one using reflection which will not call the NewConfig method.

The config package we use checks for config structs that implement an `ApplyDefaults` method and calls it on newly created config structs.

For example:

```
func (c *Config) ApplyDefaults() {
     if c.Username == "" {
        c.Username = DefaultUsername
    } 
}
```

It would be best to use the ApplyDefaults method within the NewConfig method to ensure they are the same.

services/slack/service.go
```diff
@@ -128,15 +168,15 @@ func (s *Service) Test(options interface{}) error {
...
 
-func (s *Service) Alert(channel, message, username, iconEmoji string, level alert.Level) error {
-	url, post, err := s.preparePost(channel, message, username, iconEmoji, level)
+func (s *Service) Alert(workspace, channel, message, username, iconEmoji string, level alert.Level) error {
+	url, post, err := s.preparePost(workspace, channel, message, username, iconEmoji, level)
 	if err != nil {
 		return err
 	}
-	client := s.clientValue.Load().(*http.Client)
+	client := s.clients[workspace]
```
nathanielc: How do you protect access to the clients dict here?

services/slack/service.go
```diff
@@ -184,6 +226,7 @@ func (s *Service) preparePost(channel, message, username, iconEmoji string, leve
 		Mrkdwn_in: []string{"text"},
 	}
 	postData := make(map[string]interface{})
+	postData["workspace"] = c.Workspace
```
nathanielc: Is this part of the slack API?

services/slack/service.go
nathanielc: It might be worth creating a Workspace struct that contains both a Config and http.Client so you do not have to maintain and protect access to two maps.

services/slack/service.go
```diff
@@ -23,31 +24,95 @@ type Diagnostic interface {
...
+	s := &Service{
+		diag:       d,
+		workspaces: make(map[string]*Workspace),
+	}
+
+	for _, c := range confs {
+
+		if c.InsecureSkipVerify {
+			s.diag.InsecureSkipVerify()
+		}
```
nathanielc: We also want this check to be in the `Service.Update` method.

nathanielc:
Oh and it looks like there is one more `Fatal -> Fatalf` change that needs to be made.

PR #1842, opened 2018-03-08
Add Alert Inhibitors

### LINKED ISSUE #1734: [Feature Request] Add Alert inhibitions
We have Kapacitor configured to send alerts for a bunch of measurements in multiple databases. All of these alerts have their own deadman component.

1) What we need is a configuration for a global outage across measurements and databases. Say our influxdb itself is down. Instead of Kapacitor firing an alert for each measurement in each db, can Kapacitor send out one alert?
 
2) And if say Kapacitor sends out an alert if a service is down on a node, consequently Kapacitor would send out alerts SEPARATELY for each node rather than a collective alert. Even if we develop logic for a combined alert, the individual node alerts will still fire since they're configured too. So if we have 20 nodes on which we collect and store data, and the database goes down, we need a single global alert rather than 20 separate alerts firing. However if say 2 nodes aren't reporting data, we want them to fire alerts independently. 

Are there possible configurations for the problems.

nathanielc:
Fixes #1734 

- [x] Rebased/mergable
- [x] Tests pass
- [x] CHANGELOG.md updated
- [x] Document feature


nathanielc:
Thanks, I have added a docstring to the Inhibit property method.

PR #1844, opened 2018-03-09
Node to detect changes between consecutive values in a series field

aanthony1243:
- [X] Rebased/mergable
- [X] Tests pass
- [X] CHANGELOG.md updated

### DIFF
--- a/change_detect.go
+++ b/change_detect.go
@@ -1,0 +1,139 @@
+package kapacitor
+
+import (
+	"fmt"
+	"time"
+
+	"github.com/influxdata/kapacitor/edge"
+	"github.com/influxdata/kapacitor/keyvalue"
+	"github.com/influxdata/kapacitor/models"
+	"github.com/influxdata/kapacitor/pipeline"
+)
+
+type ChangeDetectNode struct {
+	node
+	d *pipeline.ChangeDetectNode
+}
+
+// Create a new changeDetect node.
+func newChangeDetectNode(et *ExecutingTask, n *pipeline.ChangeDetectNode, d NodeDiagnostic) (*ChangeDetectNode, error) {
+	dn := &ChangeDetectNode{
+		node: node{Node: n, et: et, diag: d},
+		d:    n,
+	}
+	// Create stateful expressions
+	dn.node.runF = dn.runChangeDetect
+	return dn, nil
+}
+
+func (n *ChangeDetectNode) runChangeDetect([]byte) error {
+	consumer := edge.NewGroupedConsumer(
+		n.ins[0],
+		n,
+	)
+	n.statMap.Set(statCardinalityGauge, consumer.CardinalityVar())
+	return consumer.Consume()
+}
+
+func (n *ChangeDetectNode) NewGroup(group edge.GroupInfo, first edge.PointMeta) (edge.Receiver, error) {
+	return edge.NewReceiverFromForwardReceiverWithStats(
+		n.outs,
+		edge.NewTimedForwardReceiver(n.timer, n.newGroup()),
+	), nil
+}
+
+func (n *ChangeDetectNode) newGroup() *changeDetectGroup {
+	return &changeDetectGroup{
+		n: n,
+	}
+}
+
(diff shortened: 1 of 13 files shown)


change_detect.go
```diff
@@ -0,0 +1,139 @@
...
+type changeDetectGroup struct {
+	n        *ChangeDetectNode
+	previous edge.FieldsTagsTimeGetter
+}
+
+func (g *changeDetectGroup) BeginBatch(begin edge.BeginBatchMessage) (edge.Message, error) {
+	if s := begin.SizeHint(); s > 0 {
+		begin = begin.ShallowCopy()
+		begin.SetSizeHint(s - 1)
+	}
```
nathanielc: This size hinting is not relevant to the change detect node, meaning its now know that the batch size will be `s - 1`, but rather its unknown. As such the size hint should be set to 0 since an unknown number of points will be emitted.

change_detect.go
```diff
@@ -0,0 +1,139 @@
...
+	)
+	if store {
+		g.previous = p
+	}
+	if !emit {
+		return false
+	}
+
+	fields := n.Fields().Copy()
+	fields[g.n.d.Field] = value
```
nathanielc: Looking over the code I believe this can be simplified. The `value` here is guaranteed to be the existing value for the field correct? Meaning that all the logic to mutate the point can be removed, since changeDetect is just a special case of a filter node.

That means that `doChangeDetect` need only accept a `edge.FieldTagsTimeGetter` and does not need the setter. 

If this is not the case, what code path actually changes the value of the field?

change_detect.go
```diff
@@ -0,0 +1,139 @@
...
+	return b, nil
+}
+func (g *changeDetectGroup) DeleteGroup(d edge.DeleteGroupMessage) (edge.Message, error) {
+	return d, nil
+}
+
+// changeDetect calculates the changeDetect between prev and cur.
+// Return is the resulting changeDetect, whether the current point should be
+// stored as previous, and whether the point result should be emitted.
+func (n *ChangeDetectNode) changeDetect(prev, curr models.Fields, prevTime, currTime time.Time) (interface{}, bool, bool) {
```
nathanielc: It looks like `{prev,curr}Time` are unused, we should remove them from the function args.

integrations/streamer_test.go
```diff
@@ -144,6 +144,57 @@ stream
 	testStreamerWithOutput(t, "TestStream_InfluxQLNodeMissingValue", script, 15*time.Second, er, false, nil)
 }
 
+func TestStream_ChangeDetect(t *testing.T) {
+
+	var script = `
+stream
+	|from().measurement('packets')
+	|changeDetect('value')
+    |window()
```
nathanielc: nit, could you fix the whitespace difference here?

pipeline/change_detect.go
```diff
@@ -0,0 +1,75 @@
...
+import (
+	"encoding/json"
+	"fmt"
+)
+
+// Compute the changeDetect of a stream or batch.
+// The changeDetect is computed on a single field,
+// discarding consecutive duplicate values, detecting
+// detects the points at which the series field
+// changes from one value to another.
```
nathanielc: I think it would be good to mention the output of the changeDetect node, i.e. that it just passes the data through unmodified and therefore acts like a special kind of filter.

nathanielc: FYI, the Kapacitor docs [here](https://docs.influxdata.com/kapacitor/v1.4/nodes/) are autogenerated from these comments on the types in the pipeline package.

pipeline/tick/change_detect_test.go
```diff
@@ -0,0 +1,23 @@
...
+	d.As = "very important"
+	d.Unit = time.Hour
+	d.NonNegative()
+
+	want := `stream
+    |from()
+    |changeDetect('work')
+        .as('very important')
+        .unit(1h)
+        .nonNegative()
```
nathanielc: This test needs to be updated.

PR #1858, opened 2018-03-15
Add kafka alert handler

### LINKED ISSUE #1826: Add support for sending Alerts to Kafka queue
connect https://github.com/influxdata/kapacitor/pull/956

This feature would add an alert node that could be configured to send alerts to a kafka queue.

related ui issue https://github.com/influxdata/chronograf/issues/2904

nathanielc:
Fixes #1826 

NOTE: This PR adds a new dependency on https://github.com/segmentio/kafka-go, which itself does not have any dependencies. 

- [x] Rebased/mergable
- [x] Tests pass
- [x] CHANGELOG.md updated

### DIFF
--- a/services/kafka/service.go
+++ b/services/kafka/service.go
@@ -1,0 +1,283 @@
+package kafka
+
+import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"fmt"
+	"sort"
+	"sync"
+	"text/template"
+
+	"github.com/influxdata/kapacitor/alert"
+	"github.com/influxdata/kapacitor/keyvalue"
+	"github.com/pkg/errors"
+	kafka "github.com/segmentio/kafka-go"
+)
+
+type Diagnostic interface {
+	WithContext(ctx ...keyvalue.T) Diagnostic
+	InsecureSkipVerify()
+	Error(msg string, err error)
+}
+
+type Cluster struct {
+	mu  sync.RWMutex
+	cfg Config
+
+	writers map[string]*kafka.Writer
+}
+
+func NewCluster(c Config) *Cluster {
+	return &Cluster{
+		cfg:     c,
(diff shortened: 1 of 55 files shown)


services/kafka/service.go
```diff
@@ -0,0 +1,284 @@
...
+
+	cluster  *Cluster
+	topic    string
+	template *template.Template
+
+	diag Diagnostic
+}
+
+func (s *Service) Handler(c HandlerConfig, ctx ...keyvalue.T) (alert.Handler, error) {
+	cluster, ok := s.Cluster(c.Cluster)
```
aanthony1243: Do we want to allow for a default cluster/topic?  So they can do 
alert() 
  .kafka()  

as opposed to 
alert() 
  .kafka() 
  .cluster("default")
  .kafkaTopic("q1")

The slack update does that, but that's mostly to preserve backward compatibility with the old config format.  Probably better to just require them to be explicit

services/kafka/service.go
```diff
@@ -0,0 +1,284 @@
...
+}
+
+func (c *Cluster) writer(topic string) (*kafka.Writer, error) {
+	c.mu.RLock()
+	w, ok := c.writers[topic]
+	c.mu.RUnlock()
+	if !ok {
+		c.mu.Lock()
+		defer c.mu.Unlock()
+		w, ok = c.writers[topic]
```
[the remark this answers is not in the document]
nathanielc: My assumption is that acquiring a write lock may be an order of magnitude slower than acquiring a read lock.  Based on that assumption it makes sense to first try with the read lock then if that fails to attempt again with the write lock. This is a common pattern.

services/kafka/service.go
nathanielc: At this point I'd prefer them to be explicit. It simplifies the code quite a bit. And there is no existing behavior to preserve.

PR #1913, opened 2018-05-04
Fix Kafka configuration

jaredscheib:
- [x] Rebased/mergable
- [x] Tests pass
- [x] CHANGELOG.md updated

### The Problem
- Kafka was not being returned among the sections of the config in Kapacitor 1.5 out of the box.
- Two of the default Kafka config's duration values were missing quotes and caused errors when used as-is.

### The Solution
- Add a function to Kafka config to generate a new config, and make sure server generates that new config on startup.
- Properly quote the two durations in the default Kafka config.

### DIFF
--- a/services/kafka/config.go
+++ b/services/kafka/config.go
@@ -45,9 +45,7 @@ type Config struct {
 }
 
 func NewConfig() Config {
-	return Config{
-		Enabled: false,
-	}
+	return Config{}
 }
 
 func (c Config) Validate() error {
--- a/services/kafka/config.go
+++ b/services/kafka/config.go
@@ -44,6 +44,12 @@ type Config struct {
 	InsecureSkipVerify bool `toml:"insecure-skip-verify" override:"insecure-skip-verify"`
 }
 
+func NewConfig() Config {
+	return Config{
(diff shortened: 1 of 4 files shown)


services/kafka/config.go
```diff
@@ -44,6 +44,12 @@ type Config struct {
 	InsecureSkipVerify bool `toml:"insecure-skip-verify" override:"insecure-skip-verify"`
 }
 
+func NewConfig() Config {
+	return Config{
+		Enabled: false,
+	}
+}
```
nathanielc: Why do we want a default enabled kakfa config?

jaredscheib: Do you mean, why do we want a default disabled kafka config?

jaredscheib: To be sure, it's `Enabled: false`. I could also just do `return Config{}`, I suppose!

nathanielc: yes, just return an empty `Config{}`... and apparently I can't read :(

thanks

jaredscheib: No worries! Sounds good -- done.

PR #1980, opened 2018-06-22
fix: multi same-type discovery configs

flisky:
The LAST one will be registered multi-times.

See [Range Gotcha](https://github.com/golang/go/wiki/Range#gotchas) :)


flisky:
Testcase & changelog added. Also fix a `scraper.Close` bug.

@nathanielc, can you please take a look? Thanks!

PS. the ci error seems irrelative.

flisky:
Anyone here? This bug is annoying, and we have to fork our repository and run multi processes.

docmerlin:
@flisky sorry it has taken so long to get to this PR.  Thanks for adding it!

PR #1989, opened 2018-06-28
Improve Kafka alert throughput

nathanielc:
With this change the Async mode is used to write messages to Kafka. This mode allows for greater throughput at the cost of not blocking on the acknowledgment of the  writes. As such new statistics are added for each configured kafka cluster and topic. 

* write_errors - Reports the number of errors writing to Kafka for a given topic and cluster.
* write_messages - Reports the number of messages written to Kafka for a given topic and cluster.

- [x] Rebased/mergable
- [x] Tests pass
- [x] CHANGELOG.md updated

### DIFF
--- a/services/kafka/service.go
+++ b/services/kafka/service.go
@@ -6,13 +6,22 @@ import (
 	"encoding/json"
 	"fmt"
 	"sort"
+	"strconv"
 	"sync"
+	"sync/atomic"
 	"text/template"
+	"time"
 
 	"github.com/influxdata/kapacitor/alert"
 	"github.com/influxdata/kapacitor/keyvalue"
+	"github.com/influxdata/kapacitor/server/vars"
 	"github.com/pkg/errors"
 	kafka "github.com/segmentio/kafka-go"
+)
+
+const (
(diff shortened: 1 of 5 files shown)


nathanielc:
@stevebang This PR update the docs for the Kafka alert handler, can you incorporate these changes into the docs site?

services/kafka/service.go
```diff
@@ -60,7 +150,14 @@ func (c *Cluster) writer(topic string) (*kafka.Writer, error) {
...
 			wc.Topic = topic
-			w = kafka.NewWriter(wc)
+			kw := kafka.NewWriter(wc)
+			// Create new writer
+			w = &writer{
+				kafka:   kw,
+				cluster: c.cfg.ID,
+				topic:   topic,
+			}
+			w.Open()
```
aanthony1243: do we need a routine to cleanup and close all the writers at some point?

nathanielc: Writers already had logic to close them. This change just creates the companion Open method since now the initialization is more complex than creating the type.

PR #2070, opened 2018-10-02
Resolves #2069

### LINKED ISSUE #2069: Dangling link to udf.proto
At https://github.com/influxdata/kapacitor/blob/master/udf/agent/README.md there is a link to `udf.proto` with an absolute target of <https://github.com/influxdata/kapacitor/blob/master/udf/udf.proto> - but that gives a 404.

It looks like the correct link target should be <https://github.com/influxdata/kapacitor/blob/master/udf/agent/udf.proto>.  I believe that a relative link using Markdown `[udf.proto]()` would achieve this.

sdomino:
- [x] Rebased/mergable
- [x] Tests pass
- [x] CHANGELOG.md updated

### DIFF
--- a/udf/agent/README.md
+++ b/udf/agent/README.md
@@ -8,7 +8,7 @@ You can find those agent implementations in this directory and subdirectories ba
 
 Example uses of the agents can be found in the `examples` directory.
 These examples are working examples and are executed as part of the testing suite,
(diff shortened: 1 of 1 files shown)


sdomino:
Would it be possible to have the CI tests "kicked" manually to see if they go through? Looking at this change, and the CI error, I'm not really seeing how they're related. It appears to be external to the PR changes.

PR #2101, opened 2018-11-21
Add multiple field support to change detect

nathanielc:
This change also updates to Go 1.11 and runs `go fmt ./...`
- [x] Rebased/mergable
- [x] Tests pass
- [x] CHANGELOG.md updated


change_detect.go
```diff
@@ -106,21 +106,19 @@ func (g *changeDetectGroup) DeleteGroup(d edge.DeleteGroupMessage) (edge.Message
...
-			fmt.Errorf("expected field %s not found", n.d.Field),
-			keyvalue.KV("field", n.d.Field))
-		return false
-	}
-	if prev[n.d.Field] == value {
-		return false
+	for _, field := range n.d.Fields {
+		value, ok := curr[field]
+		if !ok {
+			n.diag.Error("Invalid field in change detect",
```
desa: Do you think that this might be a bit noisy? I'm imagining a case where you want to change detect on `field_a` or `field_b`, but 90% of the contains data that is written is for only `field_a`. Wont this create a log for every point that comes through (provided that there is no change detected each time?

nathanielc: Good question. All nodes now have the `.quiet` option to tell them to not log anything. So I think it if fine to leave as the default then in that special case the user can explicitly quite the specific changeDetect node.

Issue #532, opened 2016-05-10
x509: certificate signed by unknown authority For POST messages


fcastello:
I am getting this error

[cpu_alert:alert2] 2016/05/10 17:39:50 E! failed to POST batch Post https://user:pass@myurl/submit_result/: x509: certificate signed by unknown authority

I know the server needs some tweaking in the configuration but is it possible to skip ssl verification.

There is very little info about the post method in alert node. Some pointer will be helpful. 
Also can you point me in the code where the post is so I can look at it and understand what has been done

nathanielc:
@fcastello Currently it is not possible to ignore SSL cert errors.

Code locations:

* Struct that defines available options: https://github.com/influxdata/kapacitor/blob/master/pipeline/alert.go#L397
* Method that posts the alert data:  https://github.com/influxdata/kapacitor/blob/master/alert.go#L742

> There is very little info about the post method in alert node.  

This is in part because it does very little.

cryptokush:
tr := &http.Transport {
                TLSClientConfig: &tls.Config{InsecureSkipVerify: true},
        }
        client := &http.Client{Transport: tr}
        _, err := client.Post(post.URL, "application/json", buf)

        if err != nil {
                a.logger.Println("E! failed to POST batch", err)
        }

quick hack, no config

PR #2154, opened 2019-01-24
Add ability to skip ssl verification with an alert post node

itsHabib:
Fixes: https://github.com/influxdata/kapacitor/issues/532

- [x] Rebased/mergable
- [x] Tests pass
- [x] CHANGELOG.md updated

### DIFF
--- a/services/httppost/service.go
+++ b/services/httppost/service.go
@@ -329,8 +329,9 @@ func (h *handler) Handle(event alert.Event) {
 		req = req.WithContext(ctx)
 	}
 
+	httpClient := &http.Client{}
+
 	// Skip SSL verification?
-	httpClient := http.DefaultClient
 	if h.skipSSLVerification {
 		httpClient.Transport = &http.Transport{
 			TLSClientConfig: &tls.Config{
--- a/services/httppost/service.go
+++ b/services/httppost/service.go
@@ -2,6 +2,7 @@ package httppost
 
 import (
 	"bytes"
+	"crypto/tls"
 	"encoding/json"
 	"fmt"
 	"io"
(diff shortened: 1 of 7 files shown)


itsHabib:
Hello! There's a few comments I have regarding this issue. The first is that there is a test failing for me but I'm not sure I don't think it s due to my changes. The `server_test.go` fails when I'm on `master` or on `itshabib-skip-ssl-ver`.  Here is the output:

The other  is that I would like to add tests to actually send a POST with TLS so that I can assert using the skipSSLVerification property will prevent a panic. The only issue is that I'm not really sure where that should be placed. The three places I see where I can potentially add a test like that is `integrations/{streamer,batcher}_test.go` and `server/server_test.go`. Do you have any recommendations?

itsHabib:
Huh but looking at the CI tests there was no errors. Can it be possible its just my system?

services/httppost/service.go
```diff
@@ -323,8 +329,18 @@ func (h *handler) Handle(event alert.Event) {
 		req = req.WithContext(ctx)
 	}
 
+	// Skip SSL verification?
+	httpClient := http.DefaultClient
+	if h.skipSSLVerification {
+		httpClient.Transport = &http.Transport{
```
nathanielc: This needs to create a new client. It cannot modify the default client safely. 

```
httpClient = &http.Client{
    Transport: //...
}
```

CWoople:
I created a stream tick script with an alert that is posting to my express project from a kapacitor docker container. When the stream alert is triggered I'm getting this error. 
x509: cannot validate certificate  because it doesn't contain any IP SANs

I am using the .skipSSLVerification method within the tick script. Any ideas?
[example.txt](https://github.com/influxdata/kapacitor/files/4013302/example.txt)

PR #2167, opened 2019-03-07
fix(http): use default transport consitently

nathanielc:
The http transport used in most places did not have sane defaults. This change consolidates creation of a new http transport. 

These changes only effect outgoing HTTP connections from Kapacitor, which tend to be low volume.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,14 @@
 
 ## unreleased
 
+### Features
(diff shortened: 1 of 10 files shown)


nathanielc:
This PR should wait for https://github.com/influxdata/kapacitor/pull/2154 to be merged and then update the one instance of using creating a transport there.

PR #2181, opened 2019-03-27
fix: eliminate deadlocks in the barrier node when delete is used

### LINKED ISSUE #2144: Kapacitor leaks sockets when using barrier nodes with delete
Hi, 

kapacitor version: 1.5.2
distribution: Ubuntu xenial 16.04.3
installed from official prebuilt deb downloaded from influxdata website.

We upgraded to kapacitor 1.5.2 to take advantage of the delete feature of barrier nodes as our stream kapacitor tasks (mostly rollups) deal with ephemeral series for which memory was previously not released.

We have two sets of kapacitor nodes 
* type1: few rollup tick scripts (high cardinality data)
* type2: Around 228 tick scripts with a mix of alerts and rollups

After upgrading to version 1.5.2.  We added barrier nodes with delete before all window and union nodes in our tick scripts.

With type1, we have had only two instances of kapacitor being unresponsive (tasks don't proceed and kapacitor api is unresponsive; only way to resolve is to restart kapacitor).  We've been running with this  change for ~2 weeks and have had good uptime and it does release memory as expected.

With type2, we run into "unresponsive" kapacitor quite regularly (~3hrs).  When kapacitor gets unresponsive, we observed that the machine had 64K tcp connections to our influxdb cluster in CLOSED_WAIT and exhausted all file descriptors (the systemd unit raises it to 64K).  This does not happen when barrier nodes are added without the delete option.

Attached is a stacktrace obtained with `SIGQUIT` when the kapacitor processs was hung.  Also attaching graphs of growth in `CLOSED_WAIT` in comparision to a sibling kapacitor machine without barrier nodes and a sample tick script.

closed_wait_conns

[stacktrace.txt](https://github.com/influxdata/kapacitor/files/2734609/stacktrace.txt)
[rollup-infra-cassandra.tick.txt](https://github.com/influxdata/kapacitor/files/2734624/rollup-infra-cassandra.tick.txt)

jsternberg:
Modified two different parts of the barrier node to prevent any
deadlocks. There were two possible methods to deadlock with one of these
found to be hit by people using these nodes.

The first was when a point was received, `Point` would be called and a
message would be sent to reset the timer. If the message queue was
backed up, then it was possible for `emitBarrier` to be called and for
it to try and forward a message to itself. But, the goroutine that was
reading the points from the edge was attempting to reset the timer so
the delete group message was never able to be consumed and the timer was
never able to be reset.

This first one was fixed by adding a buffer of size one to that channel
and making the reset non-blocking when the channel was full. This
guarantees that the message will be sent and prevents it from blocking
if the reset message is sent twice in quick succession.

The second is if the delete group message was received. This would call
`Stop`. If `emitBarrier` was called in between having the delete group
message, then it could deadlock in the same way as the above because the
delete group message needs emit to be called and emit is blocked on
waiting for the idle handler to exit.

The stop method was also not thread safe and it was easy enough to make
it thread safe since it could potentially be called from a different
thread.

This is fixed by changing the stop channel to have a buffer of 1 and
making it so stopping was sending a goroutine to avoid a potential
double close of the goroutine. When the stop message is sent, it will be
done non-blocking to prevent a deadlock since only one message should be
received anyway. The call to `Stop()` in the `idleHandler` is changed to
a non-blocking version and does not wait for the goroutine to exit while
the other stop will wait for the goroutine to exit.

Fixes #2144.

### DIFF
--- a/barrier.go
+++ b/barrier.go
@@ -109,8 +109,8 @@ func newIdleBarrier(name string, group edge.GroupInfo, in edge.Edge, idle time.D
 		lastBarrierT: atomic.Value{},
 		wg:           sync.WaitGroup{},
 		outs:         outs,
-		stopC:        make(chan struct{}),
-		resetTimerC:  make(chan struct{}),
+		stopC:        make(chan struct{}, 1),
+		resetTimerC:  make(chan struct{}, 1),
 		del:          del,
 	}
 
@@ -128,11 +128,17 @@ func (n *idleBarrier) Init() {
 }
 
 func (n *idleBarrier) Stop() {
+	n.stop()
+	n.wg.Wait()
+}
+
+func (n *idleBarrier) stop() {
+	// Send a stop signal at least once to the stop channel.
+	// The stop channel has a buffer of size one and only the
+	// first stop signal matters.
 	select {
-	case <-n.stopC:
+	case n.stopC <- struct{}{}:
 	default:
-		close(n.stopC)
-		n.wg.Wait()
 	}
 }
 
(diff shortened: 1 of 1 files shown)


jsternberg:
I did not try. Let me give it a try. It should be pretty simple to replicate.

PR #2198, opened 2019-04-18
Add TLS configuration options

gunnaraasen:
Fixes https://github.com/influxdata/kapacitor/issues/2193

This PR ports the TLS configuration functionality from https://github.com/influxdata/influxdb/pull/10150 to Kapacitor.

A new `[tls]` section is added to the configuration that allows a user to set:

- Which cipher suites to use
- A minimum supported TLS version
- A maximum supported TLS version

The TLS configuration is only wired up to the HTTP service in this PR.

- [x] Rebased/mergable
- [x] Tests pass
- [x] CHANGELOG.md updated

### DIFF
--- a/tlsconfig/config.go
+++ b/tlsconfig/config.go
@@ -6,6 +6,8 @@ import (
 	"errors"
 	"fmt"
 	"io/ioutil"
+	"sort"
+	"strings"
 )
 
 // Create creates a new tls.Config object from the given certs, key, and CA files.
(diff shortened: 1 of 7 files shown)


tlsconfig/config.go
```diff
@@ -42,3 +44,131 @@ func Create(
...
 	return t, nil
 }
+
+// Update stores with into the tls config pointed at by into but only if with is not nil
+// and into is nil. Think of it as setting the default value.
+func Update(into **tls.Config, with *tls.Config) {
+	if with != nil && into != nil && *into == nil {
+		*into = with
+	}
+}
```
nathanielc: This function is not used, can you remove it?

PR #2202, opened 2019-04-22
Add templating for MQTT topics

fhriley:
This adds the ability to use templates when specifying the MQTT
topic. This is much more useful than static topics since you
can create topics like this:

switches/{{ index .Tags "agent_host" }}/ports/{{ index .Tags "port_num" }}

- [x] Rebased/mergable
- [x] Tests pass
- [x] CHANGELOG.md updated

### DIFF
--- a/services/mqtt/service.go
+++ b/services/mqtt/service.go
@@ -1,9 +1,12 @@
 package mqtt
 
 import (
+	"bytes"
 	"fmt"
 	"log"
 	"sync"
+	text "text/template"
+	"time"
 
 	"github.com/influxdata/kapacitor/alert"
 	"github.com/influxdata/kapacitor/keyvalue"
@@ -64,6 +67,7 @@ const (
 type Service struct {
 	diag Diagnostic
 
+	bufPool sync.Pool
 	mu      sync.RWMutex
 	clients map[string]Client
 	configs map[string]Config
@@ -93,7 +97,12 @@ func NewService(cs Configs, d Diagnostic) (*Service, error) {
 	}
 
 	return &Service{
-		diag:              d,
+		diag: d,
+		bufPool: sync.Pool{
+			New: func() interface{} {
+				return new(bytes.Buffer)
+			},
+		},
 		configs:           configs,
 		clients:           clients,
 		defaultBrokerName: defaultBrokerName,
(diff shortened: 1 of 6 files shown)


fhriley:
Is anyone looking at these?

fhriley:
@nathanielc @aanthony1243 Can someone look at this? We are using this in production, but we prefer not to have to maintain a custom build of kapacitor. The static topics that kapacitor supports without this change are essentially useless.

services/mqtt/service.go
```diff
@@ -233,11 +242,81 @@ type handler struct {
...
+
+	// Alert Level, one of: INFO, WARNING, CRITICAL.
+	Level string
+
+	// Time
+	Time time.Time
+
+	// Duration of the alert
+	Duration time.Duration
+}
```
nathanielc: These structures should not be copied. Use the `TemplateData` function on the alert Event in order to use consistent template data for the templates.

See https://github.com/influxdata/kapacitor/blob/master/services/pagerduty2/service.go#L402 as an example.

services/mqtt/service.go
```diff
@@ -233,11 +242,81 @@ type handler struct {
...
+		Duration: event.State.Duration,
+	}
+
+	topicBuf := h.s.bufPool.Get().(*bytes.Buffer)
+	defer func() {
+		topicBuf.Reset()
+		h.s.bufPool.Put(topicBuf)
+	}()
+
+	topicTmpl, err := text.New("topic").Parse(topic)
```
nathanielc: The topic template doesn't change for a given handler, can you udpate this to parse the template during handler construction? Then you can eval the template for each eval without parsing it. This also provides the user with earlier feedback if they have a typo in their template.

services/mqtt/service.go
fhriley: Done.

services/mqtt/service.go
fhriley: Done.

nathanielc:
@fhriley LGTM, can you fix the conflicts and then I can merge.

fhriley:
Merge conflict resolve.d

fhriley:
Test added.

PR #2246, opened 2019-09-05
Allow filling alert details template with HTML fields/tags

SuddenSelect:
Alerts' details (email body) can currently be enhanced by using HTML tags, however embedding there Field or Tag value (ex. _{{ index .Fields "someField" }}_ ) automatically triggers html escaping on said value (as per [example here](https://golang.org/pkg/html/template/#hdr-Introduction))
While this is a sensible default behavior, it unnecessarily limits Users flexibility when creating HTML-rich email notifications, and from a User's perspective seems counter-intuitive - some email body elements can have HTML and some cannot.

Proposed solution introduces a new function to template function map for encapsulation of given string in a type html/template/HTML [as suggested here](https://golang.org/pkg/html/template/#hdr-Typed_Strings).

Added tests are based on TestStream_AlertEmail.

- [ ] Rebased/mergable
- [ ] Tests pass
- [ ] CHANGELOG.md updated

### DIFF
--- a/alert.go
+++ b/alert.go
@@ -126,6 +126,9 @@ func newAlertNode(et *ExecutingTask, n *pipeline.AlertNode, d NodeDiagnostic) (a
 
 			return html.JS(tmpBuffer.String())
 		},
+		"safeHtml": func(v interface{}) html.HTML {
+			return html.HTML(v.(string))
+		},
 	}).Parse(n.Details)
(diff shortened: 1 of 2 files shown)


SuddenSelect:
2 out of 3 containers in CircleCI job seem to time-out on apt-get setup

PR #2253, opened 2019-09-18
Upgrade the kafka library to set the timestamp correctly

jsternberg:


jsternberg:
It's because the `Gopkg.toml` just specifies constraints and `Gopkg.lock` locks the revisions. I updated the version of `github.com/segmentio/kafka-go` using `dep ensure -update github.com/segmentio/kafka-go`.

The `Gopkg.lock` may have changed because I used a different version of `dep` than was originally used. I couldn't identify the version and it may have been last generated with a specific revision between `0.4.1` and `0.5.0`.

jsternberg:
I fixed the unit and integration tests and it also now checks that the timestamp is being set.

PR #2268, opened 2019-12-19
Update RPM specific packaging to match RPM naming convention

8none1:
fix: Update RPM specific packaging to match RPM naming convention

RPM naming convention says that:
 1. We should use dashes not underscores in the package filename
 2. We should include a package version (iteration) in the package name

This change removes the code which swapped underscores for dashes in the RPM specific code. 
It also moves the code which removes the package_iteration in to an else block and adds an rpm specific work around to not do that.  This was done so that non-rpm builds are not affected by this change.

Fixes: #95

- [ ] Rebased/mergable
- [ ] Tests pass
- [ ] CHANGELOG.md updated

### DIFF
--- a/build.py
+++ b/build.py
@@ -772,9 +772,10 @@ def package(build_output, pkg_name, version, nightly=False, iteration=1, static=
                                 outfile = new_outfile
                             else:
                                 if package_type == 'rpm':
-                                    # rpm's convert any dashes to underscores
-                                    package_version = package_version.replace("-", "_")
-                                new_outfile = outfile.replace("{}-{}".format(package_version, package_iteration), package_version)
+                                    # rpm naming convention requires us to include the package_iteration value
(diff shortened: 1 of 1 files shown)


build.py
```diff
@@ -772,9 +772,10 @@ def package(build_output, pkg_name, version, nightly=False, iteration=1, static=
                                 outfile = new_outfile
                             else:
                                 if package_type == 'rpm':
-                                    # rpm's convert any dashes to underscores
-                                    package_version = package_version.replace("-", "_")
```
nathanielc: hmmm this is not the solution I expected. If it works great, it been a while since I touched this code. I believe the reason the `-` are being replaced is so that what comes after the `-` doesn't get understood as the `release` part of the version. 

I am likely wrong here, but testing is king.

PR #2270, opened 2019-12-24
Change CLI to support JSON formatted for service

pnikos:
- [X] Rebased/mergable
- [X] Tests pass
- [ ] CHANGELOG.md updated

### DIFF
--- a/cmd/kapacitor/main.go
+++ b/cmd/kapacitor/main.go
@@ -701,9 +701,19 @@ func doDefine(args []string) error {
 			return errors.Wrapf(err, "failed to open file %s", *dvars)
 		}
 		defer f.Close()
-		dec := json.NewDecoder(f)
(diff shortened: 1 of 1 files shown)


pnikos:
I use Ubuntu 18 64bit

build.sh run succesfully
test.sh run and in tests-logs the results (success and failing) were the same as what I get with master)

I tested the compiled kapacitor in my local installation with both JSON formats

cmd/kapacitor/main.go
```diff
@@ -701,9 +701,19 @@ func doDefine(args []string) error {
 			return errors.Wrapf(err, "failed to open file %s", *dvars)
 		}
 		defer f.Close()
-		dec := json.NewDecoder(f)
+		data, err := ioutil.ReadAll(f)
```
nathanielc: Can you add a check for `err != nil` here?

PR #2289, opened 2020-02-11
Update build.py to support arm64

povlhp:
Currently arm64 aka aarch64 results in armhf binaries. Not ARM64 binaries. 
Modern arm chips (like those in TV boxes) does not support armhf instruction set.  exec format error.

The first change will detect aarch64 and set arch to arm64, which is supported.

Cleaned up code  around line 540 to store arch in fullarch for later use in if statement below that sets GOARM. Line 543 would overwrite the value tested in the if statement in 546 and down.
According to this: https://github.com/golang/go/wiki/GoArm GOARM should not be used for arm64

- [x] Rebased/mergable
- [x] Tests pass
- [ ] CHANGELOG.md updated

### DIFF
--- a/build.py
+++ b/build.py
@@ -388,6 +388,8 @@ def get_system_arch():
         arch = "amd64"
     elif arch == "386":
         arch = "i386"
+    elif arch == "aarch64":
+        arch = "arm64"
     elif 'arm' in arch:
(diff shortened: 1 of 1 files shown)


povlhp:
Same style patch applied to Chronograf.

timhallinflux:
thanks @povlhp -- appreciate your contribution.

PR #2301, opened 2020-02-28
RFC: Feature. Allow for overriding OpsGenieV2's alert recovery action

zabullet:
Allows the addition of the _.RecoveryAction(...)_ method to the chain of the opsGenie2 service. Historically the recovery action was either _close_|_notes_ set in the global config which didn't allow uses cases where specific scripts wanted to deviate from the default behavior.

RFC: I'm not sure I fully grokked the model and pattern for config in kapacitor so would be good to make get some input on this.

TBC: Update documentation

- [X] Rebased/mergable
- [x] Tests pass
- [X] CHANGELOG.md updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## unreleased
 
 ### Features
+- [#2301](https://github.com/influxdata/kapacitor/pull/2301): Allow for overriding OpsGenieV2's alert recovery action in tickSCRIPT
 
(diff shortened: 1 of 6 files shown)


zabullet:
Any reason PRs are languishing?

docmerlin:
@zabullet sorry it has taken so long, this is next on my list after I finish fixing a panic, and have a little bit of enterprise edition work to do.

PR #2319, opened 2020-04-15
fix(kafka): update kafka lib; make kafka errors unsilent

docmerlin:
Errors on sending to kafka when they failed, used to fail silently. This
lead to user frustration.  This updates the kafka client and injects a
logger so they now fail loudly.


docmerlin:
@nathanielc yes I have.

GeorgeJahad:
This is good news.  We have been running into these problems.  Is it possible to create a docker image?: https://hub.docker.com/_/kapacitor

PR #2322, opened 2020-04-20
Add TLS1.3 support

docmerlin:
This adds support for TLS 1.3


docmerlin:
@gunnaraasen we can't build off the most recent 1.14.2 till we remove support for SSL 3.0, but I agree.

Issue #2321, opened 2020-04-20
Depreciate ssl v3.0


docmerlin:
```
SSLv3 has been irreparably broken since the POODLE attack 5 years ago
and RFC 7568 (f.k.a. draft-ietf-tls-sslv3-diediedie) prohibits its use
in no uncertain terms.
```

https://tools.ietf.org/html/rfc7568

SSL v3.0 was removed from mainline go (in 1.14.2) in a pr as of august of 2019.  If we wish to continue using go past 1.14.2 we will should, ideally remove ssl v3.0 support.

psteinbachs:
This has been done as part of moving to go 1.15

PR #2331, opened 2020-04-30
depreciate: depreciate ssl 3.0 support

docmerlin:
- [x] Rebased/mergable
- [x] Tests pass
- [x] CHANGELOG.md updated

I didn't use a formal logger because I didn't want to re-architect the objects for a change that will be short lived (we intend to remove ssl 3.0 support in the next version).

We have to do this, if we want to be able to continue upgrading to newer versions of go.

Closes https://github.com/influxdata/kapacitor/issues/2321

### DIFF
--- a/tlsconfig/config.go
+++ b/tlsconfig/config.go
@@ -6,6 +6,7 @@ import (
 	"errors"
 	"fmt"
 	"io/ioutil"
+	"os"
 	"sort"
 	"strings"
 )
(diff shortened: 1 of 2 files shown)


tlsconfig/config.go
```diff
@@ -99,6 +100,11 @@ func (c Config) Parse() (out *tls.Config, err error) {
 		out.MaxVersion = version
 	}
 
+	if out != nil && (out.MinVersion == tls.VersionSSL30 || out.MaxVersion == tls.VersionSSL30) {
+		_, _ = fmt.Fprintln(os.Stderr, "WARNING: SSL 3.0 is depreciated and support for it will be removed soon")
+		return out, nil
```
nathanielc, line 105: Does this `if` branch need its own return? It looks the same as the one right after the `if statement`.

docmerlin:
I'm closing this issue as stale.  We had to update the version of go to avoid a security problem, which decided this issue for us.

Issue #2281, opened 2020-01-24
"Deadman" interval = 0s will crash kapacitor


wcannon:
To reproduce:

In chronograf create a tick script that uses Deadman.

Set interval = 0s

Save tick script

In testing this using the kapacitor container the container will exit on crash.

Looks like it might be a divide by zero bug (e.g.  threshold / interval)

russorat:
@wcannon thanks for the issue. A deadman of 0s doesn't really make much sense but the error shouldn't happen either way.

sivascu:
I have a strange feeling I am having the same issue. Can you guys point me to some info on how to disable the check so Kapacitor would actually start again ?

My current error is:
```
May  6 12:13:06 kapacitor01 kapacitord: 2020/05/06 12:13:06 Using configuration at: /etc/kapacitor/kapacitor.conf
May  6 12:13:59 kapacitor01 kapacitord: panic: non-positive interval for NewTicker
May  6 12:13:59 kapacitor01 kapacitord: goroutine 67092 [running]:
May  6 12:13:59 kapacitor01 kapacitord: time.NewTicker(0x0, 0xed644a107)
May  6 12:13:59 kapacitor01 kapacitord: /usr/local/go/src/time/tick.go:23 +0x147
May  6 12:13:59 kapacitor01 kapacitord: github.com/influxdata/kapacitor.(*StatsNode).runStats(0xc1fef7c140, 0x0, 0x0, 0x0, 0x0, 0x0)
May  6 12:13:59 kapacitor01 kapacitord: /root/go/src/github.com/influxdata/kapacitor/stats.go:44 +0x307
May  6 12:13:59 kapacitor01 kapacitord: github.com/influxdata/kapacitor.(*node).start.func1(0xc1fef7c140, 0x0, 0x0, 0x0)
May  6 12:13:59 kapacitor01 kapacitord: /root/go/src/github.com/influxdata/kapacitor/node.go:186 +0x9d
May  6 12:13:59 kapacitor01 kapacitord: created by github.com/influxdata/kapacitor.(*node).start
May  6 12:13:59 kapacitor01 kapacitord: /root/go/src/github.com/influxdata/kapacitor/node.go:164 +0x5d
May  6 12:13:59 kapacitor01 systemd: kapacitor.service: main process exited, code=exited, status=2/INVALIDARGUMENT

```
I tried changing kapacitor.db with a boltdb browser, but at that point the error changes to:
```
[...]

PR #2335, opened 2020-05-11
fix: panic when setting a zero interval for ticker

docmerlin:
- [ ] Rebased/mergable
- [ ] Tests pass
- [ ] CHANGELOG.md updated

Closes https://github.com/influxdata/kapacitor/issues/2281

### DIFF
--- a/barrier.go
+++ b/barrier.go
@@ -59,7 +59,7 @@ func (n *BarrierNode) NewGroup(group edge.GroupInfo, first edge.PointMeta) (edge
 
 func (n *BarrierNode) newBarrier(group edge.GroupInfo, first edge.PointMeta) (edge.ForwardReceiver, func(), error) {
 	switch {
-	case n.b.Idle != 0:
+	case n.b.Idle > 0:
 		idleBarrier := newIdleBarrier(
 			first.Name(),
 			group,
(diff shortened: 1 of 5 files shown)


barrier.go
```diff
@@ -80,7 +80,7 @@ func (n *BarrierNode) newBarrier(group edge.GroupInfo, first edge.PointMeta) (ed
 		)
 		return periodicBarrier, periodicBarrier.Stop, nil
 	default:
-		return nil, nil, errors.New("unreachable code, barrier node should have non-zero idle or non-zero period")
+		return nil, nil, errors.New("unreachable code, barrier node should have positive idle or positive period")
```
wolffcm, line 83: I'm curious, what is the behavior supposed to be if `period` and `idle` are both set? Is that possible?

wolffcm: The docs say it's based on one of them. I wonder if you should return an error if both are set?

pipeline/json.go
```diff
@@ -447,6 +448,9 @@ func unmarshalStats(data []byte, parents []Node, typ TypeOf) (Node, error) {
 	}
 	child := nodeParent.Stats(0)
 	err := json.Unmarshal(data, child)
+	if child.Interval == 0 {
+		return nil, errors.New("zero is an invalid stats interval")
```
wolffcm, line 452: This code is for serializing kapacitor pipelines, isn't it? Doesn't this check belong in `stats.go`?

barrier.go
docmerlin: That is probably a good idea.  I think it is outside the scope of this PR though.

wolffcm: Yeah, that makes sense.

pipeline/json.go
docmerlin: I added one to stats.go

PR #2380, opened 2020-08-06
Add variadic options to udf

random-char:
- [+] Rebased/mergable
- [+] Tests pass
- [-] CHANGELOG.md updated
- [-] [TICKscript Spec](https://github.com/influxdata/kapacitor/blob/master/tick/TICKscript.md) updated

Want to use lists in UDF method calls. As lists contain unknown amount of items and are passed not as a single list argument, but as X arguments, where X is list's length, it is currently impossible to add UDF method which works with lists of undefined length. To allow this, I've edited ``agent.OptionInfo`` struct to have a ``IsVariadic`` bool value. If it is set to true, one can pass any amount of values of any type into UDF method.

### DIFF
--- a/pipeline/udf.go
+++ b/pipeline/udf.go
@@ -125,8 +125,10 @@ func (u *UDFNode) Property(name string) interface{} {
 func (u *UDFNode) SetProperty(name string, args ...interface{}) (interface{}, error) {
 	opt, ok := u.options[name]
 	if ok {
-		if got, exp := len(args), len(opt.ValueTypes); got != exp {
-			return nil, fmt.Errorf("unexpected number of args to %s, got %d expected %d", name, got, exp)
+		if !opt.IsVariadic {
+			if got, exp := len(args), len(opt.ValueTypes); got != exp {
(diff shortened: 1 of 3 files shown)


danxmoran:
Hi @random-char, apologies for the long delay in reviewing this. I don't see your GitHub handle in our CLA-tracker. If you're still interested in contributing this patch, could you please sign the CLA [here](https://influxdata.com/community/cla/)?

random-char:
> Hi @random-char, apologies for the long delay in reviewing this. I don't see your GitHub handle in our CLA-tracker. If you're still interested in contributing this patch, could you please sign the CLA [here](https://influxdata.com/community/cla/)?

Done

danxmoran:
Thanks @random-char! I'm having a tough time understanding how this change will be used in practice. Could you please:
1. Rebase your changes onto the latest `master` (to fix the conflict)
2. Add a new test case that uses your new `IsVariadic` field?

danxmoran:
Closing due to inactivity

PR #2388, opened 2020-08-21
feat: httpPost url templating

docmerlin:
This adds templating to the url for httpPost.

- [x] Rebased/mergable
- [x] Tests pass
- [ ] CHANGELOG.md updated

### DIFF
--- a/services/httppost/config.go
+++ b/services/httppost/config.go
@@ -32,7 +32,7 @@ func (b BasicAuth) validate() error {
 // configuration file.
 type Config struct {
 	Endpoint          string            `toml:"endpoint" override:"endpoint"`
-	URL               string            `toml:"url" override:"url"`
+	URLTemplate       string            `toml:"url" override:"url"`
 	Headers           map[string]string `toml:"headers" override:"headers"`
 	BasicAuth         BasicAuth         `toml:"basic-auth" override:"basic-auth,redact"`
 	AlertTemplate     string            `toml:"alert-template" override:"alert-template"`
@@ -43,8 +43,8 @@ type Config struct {
 
 func NewConfig() Config {
 	return Config{
-		Endpoint: "example",
-		URL:      "http://example.com",
+		Endpoint:    "example",
+		URLTemplate: "http://example.com",
 	}
 }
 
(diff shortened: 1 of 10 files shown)


services/httppost/service.go
```diff
@@ -85,14 +89,24 @@ func (e *Endpoint) RowTemplate() *template.Template {
...
+	return &e.urlTemplate
+}
+
+func (e *Endpoint) NewHTTPRequest(body io.Reader, tmplCtx interface{}) (req *http.Request, err error) {
 	e.mu.RLock()
 	defer e.mu.RUnlock()
 	if e.closed {
 		return nil, errors.New("endpoint was closed")
 	}
+	eURL := &bytes.Buffer{}
```
ayang64, line 104: would a `strings.Builder{}` be better than a `bytes.Buffer{}`?

services/httppost/config.go
```diff
@@ -55,12 +55,22 @@ func (c Config) Validate() error {
...
 		return errors.New("must specify url")
 	}
-
-	if _, err := url.Parse(c.URL); err != nil {
-		return errors.Wrapf(err, "invalid URL %q", c.URL)
+	urlTemplate, err := GetTemplate(c.URLTemplate, "")
+	if err != nil {
+		return errors.Wrapf(err, "invalid URL template syntax %q", c.URLTemplate)
+	}
+	buf := &bytes.Buffer{}
```
ayang64, line 65: would a strings.Builder{} be more appropriate than a bytes.Buffer{}?

services/httppost/service.go
```diff
@@ -85,14 +89,24 @@ func (e *Endpoint) RowTemplate() *template.Template {
 	return e.rowTemplate
 }
 
-func (e *Endpoint) NewHTTPRequest(body io.Reader) (req *http.Request, err error) {
+func (e *Endpoint) URL() *template.Template {
```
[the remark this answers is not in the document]
docmerlin: It should be safe to do this in our case, furthermore, as we are storing a `template.Template`, so the address won't change relative to the object's address.

PR #2395, opened 2020-09-02
fix: .Details should be avaliable inside AlertTemplates

docmerlin:
This adds .Details to AlertTemplates, where it should have been.


ayang64 (approved):
LGTM

PR #2409, opened 2020-10-01
use kapacitor alert details as opsgenie description text

JamesClonk:
This is a _breaking_ change in the OpsGenie plugin behaviour of Kapacitor. The modification will use alert/event state details as description when opening OpsGenie alerts instead. This allows much more descriptive alerts in OpsGenie. Alert details in OpsGenie are still filled with the same key/values as before.

The current behaviour of simply encoding the event details as description is imho not very useful as a value for an OpsGenie alert description, Kapacitor alert details can provide much more information and are more in line with OpsGenie alert descriptions.

Would such a change make sense, or is it too breaking in current behaviour?

- [x] Rebased/mergable
- [x] Tests pass
- [ ] CHANGELOG.md updated

### DIFF
--- a/etc/kapacitor/kapacitor.conf
+++ b/etc/kapacitor/kapacitor.conf
@@ -294,6 +294,9 @@ default-retention-policy = ""
     #     * close -- close the alert
     #
     recovery_action = "notes"
+    # If true then all alerts will use their event details
+    # as OpsGenie alert description.
+    details = false
     # If true then all alerts will be sent to OpsGenie
     # without explicitly marking them in the TICKscript.
     # The team and recipients can still be overridden.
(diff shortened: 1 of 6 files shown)


docmerlin:
@JamesClonk With this kind of breaking change, it would be a good idea to add a method to the node that tells it to do this behavior, so the new behavior is optional.

JamesClonk:
@docmerlin Makes sense, I'll have a look at it.

docmerlin (changes requested):
We need a method on the node so you can select the breaking behavior while defaulting to current behavior.

JamesClonk:
@docmerlin I'm trying to figure out how to do that exactly. Is it enough to just add a field to the Config struct, does that automatically also add a method on the OpsGenie event handler node (It seems that way but I'm not 100% sure)?

Issue #2304, opened 2020-03-06
BigPanda Integration


y-ard:
Hello!

It'll be nice to have a preconfigured integration  with BigPanda, which will simplify upstream alert integration. This is simple HTTPS RESTful API, and to have it out of the box will be great.

Thanks,

~yuri

russorat:
@y-ard thanks! Have you tried configuring the Http POST endpoint to talk to BigPanda? If so, do you mind sharing the config?

y-ard:
Not yet. We still waiting on the environment 😊 nothing is flying this days 😊

y-ard:
Hi Russ,

Finally I got the test environment and I did. So far it doesn’t work ☹.

Yes can provide you with our configurations, and the way how is alert should be configured

Thanks,

~yuri

y-ard:
Well so far no success. At least major one :(
even if i am sending legit json BigPanda responds with:
err="<!DOCTYPE html>\n<html lang=\"en\">\n<head>\n<meta charset=\"utf-8\">\n<title>Error</title>\n</head>\n<body>\n<pre>Bad Request</pre>\n</body>\n</html>\n" code=400

if the same body submitted with curl command it makes it through.

plus we have some set of required fields, like "host", "cluster", etc. But i did not figured out yet how to populate it, with "alert-template".

timhallinflux:
While the event handler has been created and it represents a good first pass.  It does not currently expose the mechanism to populate the `tags` parameter. https://docs.bigpanda.io/reference#alert-object

The is a critical part of the contextual information which needs to be passed upstream to BigPanda.  

Need to extend the existing integration to allow for this to be passed.  Further, examples should be documented about how to construct the various key value pairs within a TICKscript to build up the appropriate context.

rhajek:
Alerts Object with the `tags` attribute is related to the `GET` method https://api.bigpanda.io/resources/v1.0/incidents/{incident ID}/alerts/ used to read the schema of the alert incident.

We use `POST` method https://api.bigpanda.io/data/v2/alerts for posting alerts.

BigPanda handler automatically adds all tags presented in `alert.EventData` .
Note: If the tick script contains `groupBy` operator, then resulting alert event contains only tags specified in by the `groupBy` expression.

#### Example for global tags:

telegraf global tags config:

```
[global_tags]
   dc = "us-east-1"
   rack = "1a"
   user = "$USER"%        
```

Tick Script
```
var data = stream
  |from()
[...]

PR #2414, opened 2020-10-13
Feat/Big Panda Event Handler

rhajek:
Here is initial implementation of BigPanda event handler #2304. It is not tested with on production `https://api.bigpanda.io/data/v2/alerts` yet because we don't have BigPanda account.

Example of tick script:

```
stream
  |from()
    .measurement('cpu')
  |alert()
    .warn(lambda: "usage_idle" < 20)
    .stateChangesOnly()
    .message('Hey, check your CPU')
    .bigPanda()
```

kapacitor.conf

```
[bigpanda]
  enabled = true
  app-key = "my-app-key"
  token = "my-token-123"
  state-changes-only = false
  insecure-skip-verify = true
```

- [ ] Rebased/mergable
- [ ] Tests pass
- [ ] CHANGELOG.md updated


services/alert/service.go
```diff
@@ -763,17 +767,17 @@ func (s *Service) createHandlerFromSpec(spec HandlerSpec) (handler, error) {
 		keyvalue.KV("topic", spec.Topic),
 	}
 	switch spec.Kind {
-	case "aggregate":
-		c := newDefaultAggregateHandlerConfig(s.EventCollector)
```
docmerlin, line 767: Er why was this removed?

docmerlin (changes requested):
Can you ping me when it is passing tests?

services/alert/service.go
rhajek: This was mistake, thanks for pointing that!

rhajek:
@docmerlin, tests are now passing in run 0 and 2 but I don't understand why is the TestService_GetConfig failing in run 1. On local machine TestService_GetConfig is also passing. What is the reason to have these 3 runs, I don't see any differences between them? Thanks.

CI link:
https://app.circleci.com/pipelines/github/influxdata/kapacitor/94/workflows/60e3e4e8-9c7c-462f-a57d-9fd1881fa285/jobs/4274/parallel-runs/1?filterBy=ALL

image

rhajek:
It looks like flaky test, after rerun it is all green. Now I need real BigPanda testing account to verify implementation.

docmerlin:
@rhajek working on getting us an account so we can do some manual testing.

docmerlin (changes requested):
This is a marker so we can remember we need to test against actual BP.

timhallinflux:
I'm working on getting us a test account with the BigPanda team.

AlienHoboken:
Hey guys, couple quick comments on the default `kapacitor.conf` you proposed.

1. Do you expect any of the properties on the alert to change after the initial trigger (eg. `CRITICAL` status)? If not then I would suggest setting `state-changes-only = true`. BigPanda has [deduplication logic](https://docs.bigpanda.io/docs/alert-correlation-logic#event-deduping) which will drop these events if the timestamp hasn't changed (or merge if timestamp has but status hasn't) so there may not be value in repeatedly sending the alert, it'll appear as a single alert in BigPanda anyways.

2. I'd suggest setting `insecure-skip-verify = false`. Our API `https://api.bigpanda.io/data/v2/alerts` should have a valid TLS certificate unless something on your side necessitates this feature (proxy, etc).

rhajek:
Hi @AlienHoboken 

1. `state-changes-only` should behave the same in all service integrations, common use case is that user specify deduplication directly in tick script.

```
stream
  |from()
    .measurement('cpu')
  |alert()
    .warn(lambda: "usage_idle" < 20)
    .stateChangesOnly()
    .message('Hey, check your CPU')
    .bigPanda()
```

2. I agree, `insecure-skip-verify = false` should be the default, insecure https is only for testing and non production use cases

Kapacitor Event can also contain map of custom tags and fields, that depends on data (user defined query). All additional tags are added to BP alert as `Additional attributes`, for additional fields I am not sure if they are important in BP, so they are not included.

AlienHoboken:
@rhajek Thanks, #1 makes sense to me 👍 

Awesome that you are sending the additional tags. For additional fields could you possibly share an example of what data that might contain? I could advise if it makes sense to send also.

rhajek:
First kapacitor alert in bigpanda:

### kapacitor.conf:
```
[bigpanda]
  enabled = true
  app-key = "8f57...e4"
  token = "1d9...cec"
```
### tickscript:
```
stream
  |from()
    .measurement('cpu')
  |alert()
    .warn(lambda: "usage_idle" < 20)
    .stateChangesOnly()
    .message('Hey, check your CPU')
    .bigPanda()
``` 

### bigpanda ui
image

### For cpu monitoring example kapacitor alert data looks like this:
image

timhallinflux:
any other changes requested @AlienHoboken or @docmerlin ?  Otherwise, let's also land this for 1.5.8...

services/bigpanda/service.go
```diff
@@ -0,0 +1,281 @@
...
+	}
+
+}
+
+func (s *Service) Test(options interface{}) error {
+	o, ok := options.(*testOptions)
+	if !ok {
+		return fmt.Errorf("unexpected options type %T", options)
+	}
+	//c := s.config()
```
docmerlin, line 116: delete this comment

docmerlin (changes requested):
This looks good, only really one thing to clean up (the forgotten commented out code).  Please fix it before merging.

services/bigpanda/service.go
rhajek: fixed

docmerlin (approved):

AlienHoboken:
Looks great to me @timhallinflux . It is possible customers _might_ want to include some information from the fields section, not only tags, in their alerts but we'll see that in feedback probably.

PR #2425, opened 2020-11-02
test: use non-nil parent context in tests to avoid panics

### LINKED ISSUE #2424: streamer_test.go panics on Go 1.15
Filing to hopefully save others some debugging time.

There are a few uses of `Context.WithValue` in `streamer_test.go` that pass `nil` as its 1st argument. On Go 1.13 this is fine, but on 1.15 (and maybe 1.14? haven't checked) the tests panic because the function returns the error `"cannot create context from nil parent"`

danxmoran:
Fixes #2424.

Go 1.15 doesn't allow nil values for the 1st arg to context.WithValue.

- [x] Rebased/mergable
- [x] Tests pass
- [ ] CHANGELOG.md updated


docmerlin (approved):
LGTM

PR #2441, opened 2020-12-02
perf: preallocate GroupID

docmerlin:


models/point.go
```diff
@@ -87,6 +87,17 @@ func ToGroupID(name string, tags map[string]string, dims Dimensions) GroupID {
 		return NilGroup
 	}
 	var buf strings.Builder
+	l := 0
+	if dims.ByName {
+		l += len(name) + 1
```
danxmoran, line 92: I was confused about why these values were used until I looked below at the `WriteString` / `WriteRune` calls. Would it be overkill to add comments here that show the corresponding values added to `buf`? i.e.:
```go
if dims.ByName {
  // add capacity for (name + \n)
  l += len(name) + 1
}
```

danxmoran (approved):

models/point.go
docmerlin: Good idea.

PR #2442, opened 2020-12-02
chore: update go to 1.15.5 and switch to go mod

docmerlin:
update go to 1.15.5 and switch to go mod


udf/agent/udf.pb.go
```diff
@@ -1,50 +1,29 @@
-// Code generated by protoc-gen-go.
+// Code generated by protoc-gen-go. DO NOT EDIT.
```
danxmoran, line 1: Did we bump the protoc-gen-go version and trigger this change? Or is this a gofmt thing?

danxmoran (approved):
✨ Blessed PR ✨

udf/agent/udf.pb.go
docmerlin: I had to mess with the versions a little bit due to some broken things.

PR #2448, opened 2020-12-10
fix: topic-handler duration() -> alertDuration()

docmerlin:
the topic handler match function `duration()` was colliding with the lambda function type conversion function `duration`.  This removes that collision by changing the name of the topic handler match function to `alertDuration`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Bugfixes
+- [#2448](https://github.com/influxdata/kapacitor/pull/2448): Changes the alert-handler match function duration() to be alertDuration() to avoid name collision with the type conversion function of the same name.
+
+
 ### Features
 - [#1839](https://github.com/influxdata/kapacitor/pull/1839): Add Subscription path configuration option to allow Kapacitor to run behind a reverse proxy, thanks @aspring
 - [#2055](https://github.com/influxdata/kapacitor/pull/2055): Add support for correlate in the Alerta AlertNode, thanks @nermolaev!
--- a/services/alert/handlers.go
+++ b/services/alert/handlers.go
@@ -373,7 +373,7 @@ const (
 	levelFunc    = "level"
 	nameFunc     = "name"
 	taskNameFunc = "taskName"
-	durationFunc = "duration"
+	durationFunc = "alertDuration"
 )
 
 var matchIdentifiers = map[string]interface{}{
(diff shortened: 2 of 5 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Bugfixes
+- [#2448](https://github.com/influxdata/kapacitor/pull/2448): Changes the alert-handler match function duration() to be alertDuration() to avoid name collision with the type conversion function of the same name.
```
danxmoran, line 6: Is this something we can change in a patch release? Or is it not really a breaking change because the collision was preventing all uses of the match function form?

services/alert/service.go
```diff
@@ -1017,10 +1017,20 @@ func (s *Service) createHandlerFromSpec(spec HandlerSpec) (handler, error) {
...
 	}
+	if h == nil && err != nil {
+		return handler{}, err
+	}
 	if spec.Match != "" {
 		// Wrap handler in match handler
 		handlerDiag := s.diag.WithHandlerContext(ctx...)
-		h, err = newMatchHandler(spec.Match, h, handlerDiag)
+		if h == nil {
+			panic("handler is nil, this should not happen")
```
danxmoran, line 1027: Does this one need to be a panic? Could we return an `err` instead?

services/alert/handlers.go
```diff
@@ -440,6 +439,21 @@ func newMatchHandler(match string, h alert.Handler, d HandlerDiagnostic) (*match
...
+				h.diag.Error("recovered from panic", errors.New(r))
+			case error:
+				h.diag.Error("recovered from panic", r)
+			}
+		}
+	}()
+
+	if h.h == nil {
+		panic("h.h is nil")
+	}
```
danxmoran, line 455: This looks like a debug check, do we want to leave it in?

danxmoran (commented):
Are there any low-hanging automated tests we could add?

CHANGELOG.md
docmerlin: Yah, it isn't a breaking change because the old `duration()` function never actually worked.

services/alert/handlers.go
docmerlin: Doh, I thought I removed it.

services/alert/service.go
docmerlin: It is a state that should never happen and should panic.  The problem was it was panic-ing in a very distant place and made it hard to debug.

danxmoran (approved):

PR #2454, opened 2020-12-15
feat: add fields to BigPanda alert

rhajek:
This PR improves BigPanda alerting:
-  by default BP Alert contains all field values and tags
-  BigPanda primary and secondary property is configurable in tick script or in topic handler

```tick
alert
  .bigPanda()
  .appKey('fc39458f98e91eb0310258c3b725d643')
  .primaryProperty('device')
  .secondaryProperty('sensor_name')
  .topic('bigpanda-topic')
```
or in topic hadler
```yaml
id: bigpanda-cpu-alert
topic: bigpanda-topic
kind: bigpanda
options:
  app-key: '8f5795cada387b9f310f41fc369c88e4'
  primary-property: 'host'
  secondary-property: 'check'
  url: 'http://localhost:8080/data/v2/alerts'
```

- [ ] Rebased/mergable
- [ ] Tests pass
- [ ] CHANGELOG.md updated

### DIFF
--- a/services/bigpanda/config.go
+++ b/services/bigpanda/config.go
@@ -26,16 +26,17 @@ type Config struct {
 	// Only applies if global is also set.
 	StateChangesOnly bool `toml:"state-changes-only" override:"state-changes-only"`
 
-	// Whether to skip the tls verification of the alerta host
+	// Whether to skip the tls verification
 	InsecureSkipVerify bool `toml:"insecure-skip-verify" override:"insecure-skip-verify"`
 
-	// BigPanda Alert webhook URL,
-	// https://api.bigpanda.io/data/v2/alerts
+	//BigPanda Alert api URL, if not specified https://api.bigpanda.io/data/v2/alerts is used
 	URL string `toml:"url" override:"url"`
 }
 
 func NewConfig() Config {
-	return Config{}
+	return Config{
+		URL: defaultBigPandaAlertApi,
+	}
 }
 
 func (c Config) Validate() error {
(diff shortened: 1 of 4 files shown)


docmerlin (approved):

services/bigpanda/config.go
```diff
@@ -29,17 +29,20 @@ type Config struct {
...
+	// BigPanda Alert webhook URL,
+	// https://api.bigpanda.io/data/v2/alerts
 	URL string `toml:"url" override:"url"`
 }
 
 func NewConfig() Config {
-	return Config{
-		URL: defaultBigPandaAlertApi,
-	}
+	return Config{}
```
docmerlin, line 38: Why this change?

rhajek: I think that the dafault url https://api.bigpanda.io/data/v2/alerts should not be hardcoded. For now we use the generic REST API, but we want BigPanda to add our integration among the supported official integrations. Instead of a generic REST API url, we will get a special webhook url like https://inbound.bigpanda.io/influxdata/alerts, which will be generated in the bigpanda admin console.

docmerlin: Reasonable defaults are a good idea for users, this was a specific ask, for a previous PR.

services/bigpanda/service.go
```diff
@@ -215,24 +218,45 @@ func (s *Service) preparePost(appKey, id string, message string, details string,
...
 		return nil, err
 	}
 
-	alertUrl, err := url.Parse(c.URL)
+	var bpUrl string
+	if hc.URL != "" {
+		bpUrl = hc.URL
+	} else {
+		bpUrl = c.URL
+	}
```
docmerlin, line 258: Since strings in go are immutable, we can do this which is a bit cleaner.

```go
    bpUrl := hc.URL
    if bpUrl == "" {
        bpUrl = c.URL
}
```

docmerlin (approved):

PR #2456, opened 2020-12-18
Gzip data by default that is sent to influxdb.

docmerlin:
This gzips data by default that is sent to influxdb.


danxmoran (approved):
:shipit:

PR #2479, opened 2021-02-12
fix: influx gzip writes

docmerlin:
This is a fix for a data race in kapacitor 1.5.8 where large batch writes were not completely writing to influxdb.


influxdb/client.go
```diff
@@ -571,6 +568,33 @@ func (p Point) Bytes(precision string) []byte {
 	return bytes
 }
 
+func (p Point) BytesWithLineFeed(precision string) []byte {
+	key := imodels.MakeKey([]byte(p.Name), imodels.NewTags(p.Tags))
```
docmerlin, line 572: This is a copy of the `func (p Point) Bytes` method, only it adds a line feed to the end of the line to avoid the allocation of doing it afterwards.

influxdb/client_test.go
```diff
@@ -196,6 +197,59 @@ func TestClient_Write(t *testing.T) {
 	}
 }
 
+func TestClient_WriteLarge(t *testing.T) {
+	expected := &bytes.Buffer{}
```
docmerlin, line 201: This writes a million points to an httptest server and checks to make sure that it is delivered properly.  The million points are gzipped.

influxdb/point_test.go
```diff
@@ -78,5 +78,10 @@ func TestPoint_Bytes(t *testing.T) {
 			t.Errorf("%s: Bytes() mismatch:\n actual:	%v\n exp:		%v",
 				test.name, got, exp)
 		}
+		if got, exp := string(p.BytesWithLineFeed(test.precision)), test.exp+"\n"; got != exp {
+			t.Errorf("%s: Bytes() mismatch:\n actual:	%v\n exp:		%v",
```
docmerlin, line 82: We want to test not just the Point.Bytes() but also Point.BytesWithLineFeed()

influxdb/client.go
```diff
@@ -337,28 +336,26 @@ func (c *HTTPClient) Write(bp BatchPoints) error {
...
-
-	}
-
-	for _, p := range bp.Points() {
-		if _, err := b.Write(p.Bytes(precision)); err != nil {
+		if err := bodyWriteCloser.Close(); err != nil {
 			return err
 		}
 
-		if err := b.WriteByte('\n'); err != nil {
```
lesam, line 356: It's very surprising to me that this would be allocating - what is the allocation caused by?

davidby-influx (approved):
Makes sense to me.  I prefer the synchronous copy to the go proc pattern unless you are using a channel to handle the EOF.

influxdb/client.go
docmerlin: That line doesn't allocate, but the gzip writer doesn't have a WriteByte, it only has Write, which means allocating a single element byte slice.

PR #2488, opened 2021-02-26
fix: function node name for ServiceNow handler

alespour:
Fixes generated function node name for _ServiceNow_ event handler.

- [x] Rebased/mergable
- [x] Tests pass


sranka (approved):
thank you

PR #2489, opened 2021-02-26
fix: memory leaks.

docmerlin:
There were a number of memory leaks of the form of `a = a[i:]`.
These were replaced with grow-able circular queue.


lesam:
Could we extract the pattern? i.e. 

```
a.PushBack(<slice>)
a.PopFront(<slice>)
a.Peek(index) <slice>
```

We already have codegen tools for the various slice types and it would make me feel better.

nathanielc (approved):
LGTM, in short these buffers grow indefinitely so when we reslice them to only the end of the buffer we need to shift all the data to the front of the buffer otherwise Go cannot free the unused space. 

These appends therefore do not do any new allocations rather move the data to the beginning of the slice

lesam:
Also this repo should be review-protected!

lesam (changes requested):
see comment

lesam:
Mostly I'm worried about someone relying on these slices not being mutated, if we had the pattern extracted that would prevent future errors. i.e. these two are not equivalent:

```
a := []int{1,2,3}
b := a[:1]
fmt.Println(a[0])
```

```
a := []int{1,2,3}
b := append(a[0:], a[:1]...)
fmt.Println(a[0])
```

lesam:
Possibly a better interface:

a.PushBack(<slice>)
a.PopFront()
a.PeekFront() <value>

circularqueue.gen_test.go
```diff
@@ -0,0 +1,218 @@
...
+		q.tail += n
+		q.data = q.data[:q.tail] // resize q.data to where it should be
+	case cap(q.data) < len(v)+q.l: // we need to grow
+		if q.tail > q.head { // where the current buffer is filled with continuous data
+			// we need to slice up v by how much we need to grow added to any free space left at the end of the buffer
+			// which is (len(v)+q.l-cap(q.data))+(cap(q.data)- q.tail)
+			endL := len(v) + q.l - q.tail
+			q.data = append(q.data[:0], append(v[endL:], append(q.data[q.head:q.tail], v[:endL]...)...)...)
+			q.tail += endL
+			break
```
docmerlin, line 154: This may be a bit overengineered, I ended up not needing more than 1 enqueue at a time.  I am not sure if I should remove the ability to enqueue multiple items at once.

circularqueue_test.go
```diff
@@ -0,0 +1,180 @@
...
+	for i := 0; i < 20; i++ {
+		expectedFinalizedItems = append(expectedFinalizedItems, i)
+	}
+
+	q := newCircularIntPtrCircularQueue(nil)
+	for i := 0; i < len(expectedFinalizedItems); i++ {
+		i := i // make i a local object
+		item := circularIntPtr(&i)
+		runtime.SetFinalizer(item, func(q circularIntPtr) {
+			// if the finalizer is called, that means that the GC believes the objects should be freed
```
docmerlin, line 140: We use `runtime.SetFinalizer` to test if the GC is seeing the objects for deletion.

join.go
```diff
@@ -207,11 +206,6 @@ func (n *JoinNode) matchPoints(p srcPoint) {
 				break
 			}
 		}
-		if n.allReported {
-			// Can't trust lowMark until all parents have reported.
-			// Remove any unneeded match points.
-			n.matchGroupsBuffer[groupId] = matches[i:]
-		}
```
docmerlin, line 214: This block above,  looks like it creates duplicates when stuff hasn't reported.  I removed it, it could be related to an issue that has been seen.

circularqueue.gen_test.go
```diff
@@ -0,0 +1,218 @@
+// Generated by tmpl
```
lesam, line 1: did you mean to call this file `gen_test` ?

circularqueue.gen_test.go
lesam: I vote for single enqueue - feel free to save the multi-enqueue version somewhere in case we need it in the future. Much easier to test single enqueue.

circularqueue.gen_test.go
lesam: Oh I see this is just for testing.

join_circularqueues.gen.go
```diff
@@ -0,0 +1,218 @@
...
+	data    []srcPoint
+	head    int
+	tail    int
+	l       int
+	oldHead int
+}
+
+// newSrcPointconstructs a Circular Queue
+// with a given buffer buf.  It is ok for buf to be nil.
+func newSrcPointCircularQueue(buf ...srcPoint) *srcPointCircularQueue {
```
lesam, line 21: Looks like this is never called?

circularqueue.gen_test.go
lesam: Easier for me to review as well 😄

docmerlin: Backed up to: https://gist.github.com/docmerlin/b29b9a36c120e239b1e6c348d09dc0c1

circularqueue.gen.go.tmpl
```diff
@@ -0,0 +1,111 @@
...
+	if q.l==0{
+		q.head = 0
+		q.tail = 0
+	}
+	return
+}
+
+// Peek peeks i ahead of the current head of queue.  It should be used in conjunction with .Len() to prevent a panic.
+func (q *{{ $k }}CircularQueue) Peek(i int) {{ $k }} {
+	p := q.head + i
```
lesam, line 95: Should we bounds check to make sure we don't get garbage?

circularqueue.gen.go.tmpl
```diff
@@ -0,0 +1,111 @@
...
+		if q.tail == len(q.data){
+			q.tail = 0
+		}
+		q.data[q.tail] = v
+	}else{ // we need to grow
+		buf := make([]{{ $k }}, cap(q.data)*2)
+		if q.head < q.tail{
+			copy(buf, q.data[q.head:q.tail])
+		} else {
+			copy(buf[copy(buf, q.data[q.head:]):], q.data[:q.tail])
```
lesam, line 44: This might be easier to understand written on more than one line:

```
partialLength := copy(buf, q.data[q.head:])
copy(buf[partialLength):], q.data[:q.tail])
```

circularqueue_test.go
```diff
@@ -0,0 +1,249 @@
...
+		t.Errorf("expected peeking, we would see %v, got %v", exp, peekRes)
+
+	}
+	peekRes = []circularInt{}
+	for i := 0; i < q.Len(); i++ {
+		peekRes = append(peekRes, q.Peek(i))
+
+	}
+
+	//if q.Dequeue(1); q.Peek(-1) != q.Val() {
```
lesam: commented code ?

circularqueue.gen.go.tmpl
docmerlin: It's intended to be used in conjuction with .Len to avoid that.

docmerlin: Also it is bound checked.

circularqueue_test.go
```diff
@@ -0,0 +1,249 @@
...
+			}
+			if !reflect.DeepEqual(res, c.expected) {
+				t.Errorf("expected %v, got %v", c.expected, res)
+			}
+		})
+
+	}
+}
+
+func Test_leakCircularBuf(t *testing.T) {
```
lesam, line 189: It might be easier to use Benchmarks for this sort of test, it gives you the number of allocations. (for the future, this is fine as-is)

circularqueue.gen.go.tmpl
```diff
@@ -0,0 +1,114 @@
+//lint:file-ignore U1000 this is generated code
+package kapacitor
+
+{{with $types := .}}{{range $k := $types}}
+
+// {{ $k }}CircularQueue defines a circular queue, always use the contructor to create one.
+type {{ $k }}CircularQueue struct {
+	data     []{{ $k }}
+	head     int
+	tail     int
```
lesam, line 10: invariant: `tail = (head + l) % len(data)` right? Is it worth keeping it separately?

circularqueue.gen.go.tmpl
```diff
@@ -0,0 +1,114 @@
...
+	q.l-=n
+	if q.l==0{
+		q.head = 0
+		q.tail = 0
+	}
+	return
+}
+
+// Peek peeks i ahead of the current head of queue.  It should be used in conjunction with .Len() to prevent a panic.
+func (q *{{ $k }}CircularQueue) Peek(i int) {{ $k }} {
```
lesam, line 94: This might be better named `At` since we're only showing one value at an index, Peek sounds like it should return a slice. Not a blocker.

circularqueue.gen.go.tmpl
docmerlin: Yes, its mostly so we don't have to do a mod over and over again.

circularqueue_test.go
docmerlin: This isn't for testing allocations, but for testing to make sure things get properly de-allocated (a test against leaks).

circularqueue.gen.go.tmpl
docmerlin: I think I will keep it this same.

circularqueue.gen.go.tmpl
docmerlin: Ok added the bounds check.

lesam (approved):

PR #2492, opened 2021-03-05
fix: tests weren't running properly in CI

docmerlin:
Tests were not properly running in CI.  This makes them run properly.


influxdb/client_test.go
```diff
@@ -213,10 +219,14 @@ func TestClient_WriteLarge(t *testing.T) {
 			t.Errorf("expected gzip Content-Encoding but got %s", r.Header.Get("Content-Encoding"))
 		}
 		if string(bod) != expected.String() {
-			t.Errorf("unexpected send, expected '%s', got '%s'", expected, string(bod))
+			t.Errorf("unexpected send:\n%s",
```
danxmoran, line 222: Does `Errorf` halt the test? (Is there a chance the `wg.Done()` below will never be called?)

influxdb/client_test.go
```diff
@@ -243,11 +253,12 @@ func TestClient_WriteLarge(t *testing.T) {
 	if err != nil {
 		t.Errorf("unexpected error.  expected %v, actual %v", nil, err)
 	}
-
+	wg.Add(1)
 	err = c.Write(bp)
```
danxmoran, line 257: Is `Write` async here, and that's why we need the wait group?

server/server_test.go
```diff
@@ -6310,6 +6315,80 @@ func TestServer_UpdateConfig(t *testing.T) {
...
+					mCopy[k] = vCopy.Interface()
+				}
+			}
+			return mCopy
+		}
+	}
+	if !cmp.Equal(deepCopyMapWithReplace(defMap)(), defMap) {
+		t.Fatalf("deepCopyMapWithReplace is broken expected %v, got %v", defMap, deepCopyMapWithReplace(defMap)())
+	}
+	{ // new block to keep vars clean
```
danxmoran, line 6384: IMO the extra indentation from this makes it less readable, but not the hill I'm going to die on 😄

udf/agent/io_test.go
```diff
@@ -2,9 +2,12 @@ package agent_test
 
 import (
 	"bytes"
-	"reflect"
 	"testing"
 
+	"github.com/google/go-cmp/cmp/cmpopts"
+
+	"github.com/google/go-cmp/cmp"
+
```
danxmoran, line 10: Formatting needs cleaned up here, I think

danxmoran (commented):
Some small stuff

influxdb/client_test.go
docmerlin: `Errorf` does not halt the test, just logs and marks the test as having failed.

influxdb/client_test.go
docmerlin: It isn't, but there was a bug in v1.5.8 that made it async (and fail for large payloads), this was added as a regression test around that bug.

danxmoran (approved):

PR #2493, opened 2021-03-05
feat(kafka): route alerts to partitions by ID, allow for configuring the hashing strategy

### LINKED ISSUE #2413: Kafka messages with the same key are not written to the same partition
### The Problem
The Kafka doc says "Events with the same event key (e.g., a customer or vehicle ID) are written to the same partition, and Kafka guarantees that any consumer of a given topic-partition will always read that partition's events in exactly the same order as they were written.". See: https://kafka.apache.org/documentation/#intro_concepts_and_terms

However, when writing Kapacitor events to Kafka, that is not the case.

### Possible Solution
I assume that is because the _Balancer_ in _WriterConfig_ is not set. The _Balancer_ defaults to the _RoundRobinGroupBalancer_ as you can see here: https://godoc.org/github.com/segmentio/kafka-go#WriterConfig

To fix this issue one would have to set _Hash_ as _Balancer_. That ensures that messages with the same key are written to the same partition. See: https://godoc.org/github.com/segmentio/kafka-go#Hash.Balance

### Bug Report Infos
Operating System: 64-Bit RHEL 8
Kapacitor Version: Kapacitor OSS 1.5.6 (git: HEAD ee3f609cd1c96f7c8f4eea924278db1fb2a96f0b)
installed using a pre-built package

### Test Case That Demonstrates The Issue
Write a streaming task that sends events with the same key to a Kafka topic with 3 partitions. To set the key you have  to set the id. For example:

dbrp "telegraf"."autogen"

stream
    |from()
        .measurement('cpu')
    |alert()
        .crit(lambda: int("usage_idle") <  70)
        .id('foo')
        .kafka()
            .cluster('my-cluster')
            .kafkaTopic('my-topic')

Now Kafka messages with the key "foo" will be sent to Kafka. They should all be written to the same partition but they are not.

danxmoran:
Closes #2413

The intended default behavior is to partition-by-id, using the crc32 algorithm (to match confluent, who are pretty dominant in the space).

Not sure where the documentation lives that will need to be updated to cover the new options.

- [ ] Rebased/mergable
- [ ] Tests pass
- [ ] CHANGELOG.md updated


go.mod
```diff
@@ -82,7 +82,7 @@ require (
 	github.com/russross/blackfriday v0.0.0-20151110051855-0b647d0506a6 // indirect
 	github.com/samuel/go-zookeeper v0.0.0-20161028232340-1d7be4effb13 // indirect
 	github.com/sean-/seed v0.0.0-20170313163322-e2103e2c3529 // indirect
-	github.com/segmentio/kafka-go v0.3.5
+	github.com/segmentio/kafka-go v0.3.10
```
danxmoran, line 85: Not sure if needed, just seemed like a good idea to upgrade to the latest patch. Can revert.

docmerlin: Nah, sounds good.

server/server_test.go
```diff
@@ -8812,9 +8812,13 @@ func TestServer_ListServiceTests(t *testing.T) {
...
 				Options: client.ServiceTestOptions{
 					"cluster": "example",
-					"topic":   "test",
 					"key":     "key",
 					"message": "test kafka message",
+					"target": map[string]interface{}{
+						"PartitionAlgorithm": "",
+						"PartitionById":      true,
+						"Topic":              "test",
+					},
```
danxmoran, line 8821: @docmerlin will this schema change break people when they upgrade? I could rework it so topic stays a top-level key, and the partitioning info is in its own struct

docmerlin (approved):
Approved, just needs the CHANGELOG.md updated.

docmerlin (changes requested):
Yes, it will break upgrading

services/kafka/service.go
```diff
@@ -318,21 +315,23 @@ func (s *Service) Test(options interface{}) error {
 	if !ok {
 		return fmt.Errorf("unknown cluster %q", o.Cluster)
 	}
-	return c.WriteMessage(s.diag, o.Topic, []byte(o.Key), []byte(o.Message))
+	return c.WriteMessage(s.diag, WriteTarget{Topic: o.Topic}, []byte(o.Key), []byte(o.Message))
```
danxmoran, line 318: I figure people don't really need to specify the partitioning details when they're just testing the integration

docmerlin (approved):

PR #2498, opened 2021-03-11
fix: avoid infinite hang when closing Kakfa writer

### LINKED ISSUE #2497: Closing a Kafka writer hangs
* Our Kafka writers run a metric-collecting goroutine in the background.
* The goroutine `range`s over a `time.Ticker`'s channel, and calls `Done()` on a `sync.WaitGroup` after the `range` completes.
* `Close()` on the writer calls `Stop()` on the ticker, then `Wait()` on the wait group
* Calling `Stop()` on a ticker doesn't actually close its channel, so:
   * The `range` over the channel never completes
   * `Done()` is never called on the wait group
   * `Wait()` on the wait group never completes

danxmoran:
Fixes #2497

### DIFF
--- a/services/kafka/service.go
+++ b/services/kafka/service.go
@@ -16,7 +16,7 @@ import (
 	"github.com/influxdata/kapacitor/keyvalue"
 	"github.com/influxdata/kapacitor/server/vars"
 	"github.com/pkg/errors"
-	kafka "github.com/segmentio/kafka-go"
+	"github.com/segmentio/kafka-go"
 )
 
 const (
@@ -51,7 +51,11 @@ type writer struct {
 	wg sync.WaitGroup
 
 	statsKey string
-	ticker   *time.Ticker
+
+	// time.Ticker doesn't expose any way to close its internal channel,
+	// so we have to use a side-channel to signal when it's been stopped.
+	ticker     *time.Ticker
+	tickerDone chan struct{}
 }
 
 func (w *writer) Open() {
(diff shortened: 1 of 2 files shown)


services/kafka/service_test.go
```diff
@@ -0,0 +1,34 @@
...
+	diag := diagnostic.NewService(diagnostic.NewConfig(), os.Stderr, os.Stdin)
+	require.NoError(t, diag.Open())
+	defer diag.Close()
+
+	// Write a message to generate a writer.
+	require.NoError(t, cluster.WriteMessage(diag.NewKafkaHandler(), "testTopic", []byte{1, 2, 3, 4}, []byte{1, 2, 3, 4}))
+
+	// Update the config in a way that requires shutting down all active writers.
+	c.UseSSL = true
+	require.NoError(t, cluster.Update(c))
```
danxmoran, line 33: Without the changes in `service.go`, this test never completes.

services/kafka/service.go
```diff
@@ -71,6 +75,7 @@ func (w *writer) Open() {
 	statsMap.Set(statWriteMessageCount, writeMessages)
 
 	w.ticker = time.NewTicker(time.Second)
+	w.tickerDone = make(chan struct{}, 1)
```
docmerlin, line 78: this doesn't need to have a `1)` since its only ever used by closing.  It isn't that big a deal though.

docmerlin (approved):

PR #2501, opened 2021-03-18
fix: external Zenoss testability

alespour:
This PR
- fixes external testability of Zenoss handler (needed for Chronograf issue #2620)
- adds missing alert test
- adds missing `summary` field default value assignment to alert message (as described in documentation)

- [x] Rebased/mergable
- [x] Tests pass


services/zenoss/service.go
```diff
@@ -74,18 +74,32 @@ func (s *Service) StateChangesOnly() bool {
...
 type testOptions struct {
-	Action string `json:"action"`
-	Method string `json:"method"`
-	Type   string `json:"type"`
-	TID    int64  `json:"tid"`
+	AlertID       string                 `json:"alert_id"`
+	Level         alert.Level            `json:"level"`
+	Message       string                 `json:"message"`
+	Action        string                 `json:"action"`
+	Method        string                 `json:"method"`
```
sranka, line 81: Please use also the new properties in the Test function.

PR #2506, opened 2021-03-23
fix: build 1.6

docmerlin:
This should fix the failing protoc.


lesam (approved):
Some test skips to be fixed, and some module fixes. LGTM

PR #2530, opened 2021-04-13
feat: tricklenode

docmerlin:
This creates a node that converts from a batch to a stream, it could be thought of as the inverse of windownode.


pipeline/tick/tick_test.go
```diff
@@ -8,6 +8,8 @@ import (
 	"testing"
 	"time"
 
+	"github.com/google/go-cmp/cmp"
+
```
danxmoran, line 12: Needs reformatted

trickle.go
```diff
@@ -0,0 +1,66 @@
...
+
+type TrickleNode struct {
+	node
+	dims models.Dimensions
+	name string
+}
+
+func (n *TrickleNode) runTrickle(_ []byte) error {
+	consumer := edge.NewConsumerWithReceiver(
+		n.ins[0],
```
danxmoran, line 25: Do we enforce anywhere that `len(ins) == 1`?

trickle.go
```diff
@@ -0,0 +1,66 @@
...
+// BeginBatch sets some batch variables on the node, and isn't forwarded.
+func (n *TrickleNode) BeginBatch(b edge.BeginBatchMessage) error {
+	n.dims = b.Dimensions()
+	n.name = b.Name()
+	return nil
+}
+
+// BatchPoint forwards a PointMessage
+func (n *TrickleNode) BatchPoint(bp edge.BatchPointMessage) error {
+	return n.outs[0].Collect(edge.NewPointMessage(
```
danxmoran, line 40: Same question with `len(outs)`

integrations/batcher_test.go
```diff
@@ -2605,6 +2602,42 @@ data
...
+	var script = `
+	var data = batch
+	|query('SELECT value FROM "telegraf"."default"."cpu"')
+		.period(4s)
+		.every(4s)
+		.groupBy('host')
+data
+	|trickle()
+	|window().period(10s)
+	|httpOut('TestBatch_Trickle')`
```
danxmoran, line 2615: Checking my understanding: this test is showing that `trickle()` followed by `window` is a way to re-batch data? i.e. in this case, re-batching from 4s windows into 10s windows?

trickle.go
docmerlin: Yes, in the attachment, in the tick code.

trickle.go
docmerlin: Yes, in the attachment, in the tick code.  if len is not 1, we will have errors way before it gets to here.

danxmoran (approved):

Issue #2532, opened 2021-04-13
Update prometheus dependency


lesam:
The prometheus version Kapacitor depends on for scrapers is from 2017, we should update it. This will unblock pulling in other more modern packages (e.g. flux).

PR #2536, opened 2021-04-13
chore: update prometheus dependency for service discovery

lesam:
Closes: https://github.com/influxdata/kapacitor/issues/2532

There were significant reformatting changes in the prometheus service discovery
packages and their configuration, but it seems like the logic is mostly similar.

- [x] Rebased/mergable
- [x] Tests pass
- [x] CHANGELOG.md updated


lesam:
Do we want a changelog for this?

Dockerfile_build_ubuntu32
```diff
@@ -1,58 +0,0 @@
-FROM 32bit/ubuntu:16.04
```
lesam, line 1: Delete the 32 bit tests since we will soon be deleting the 32 bit build.

go.mod
```diff
@@ -3,112 +3,75 @@ module github.com/influxdata/kapacitor
...
 	github.com/onsi/ginkgo v1.14.2 // indirect
 	github.com/onsi/gomega v1.10.3 // indirect
-	github.com/pborman/uuid v0.0.0-20160209185913-a97ce2ca70fa // indirect
-	github.com/pkg/errors v0.8.1
-	github.com/prometheus/common v0.9.1
-	github.com/prometheus/prometheus v0.0.0-00010101000000-000000000000
+	github.com/pierrec/lz4 v2.5.2+incompatible // indirect
+	github.com/pkg/errors v0.9.1
+	github.com/prometheus/common v0.20.0
+	github.com/prometheus/prometheus v1.8.2-0.20210331101223-3cafc58827d1
```
lesam, line 52: This is actually tag v2.26.0 of prometheus, but https://github.com/prometheus/prometheus/issues/6048 prevents us from using the semver name, we have to specify the commit instead.

services/scraper/service.go
```diff
@@ -42,13 +49,23 @@ type Service struct {
 
 	discoverers []Discoverer
 
-	// TargetManager represents a scraping/discovery manager
-	mgr interface {
-		ApplyConfig(cfg *config.Config) error
-		Stop()
-		Start()
```
lesam, line 49: The prometheus fork we had was just to split `Run` into `Start` and `Wait`. But the latest version of prometheus looks like it should support calling `Stop` before `Run` in racy cases, so I don't think we need the fork any longer.

lesam:
Tested: https://gist.github.com/lesam/1120b58a5261a2b7eee8270ce9426727

services/scraper/service.go
docmerlin: awesome!

docmerlin (approved):

services/marathon/config.go
```diff
@@ -27,11 +28,11 @@ type Config struct {
 	SSLServerName string `toml:"ssl-server-name" override:"ssl-server-name"`
 	// Use SSL but skip chain & host verification
 	InsecureSkipVerify bool `toml:"insecure-skip-verify" override:"insecure-skip-verify"`
+	// This previously supported Timeout but it was removed, see https://github.com/prometheus/common/pull/123
```
lesam, line 31: This needs a docs change.

PR #2548, opened 2021-05-10
feat: client work for queryFlux

docmerlin:


influxdb/client.go
```diff
@@ -93,6 +106,7 @@ const (
 	NoAuthentication AuthenticationMethod = iota
 	UserAuthentication
 	BearerAuthentication
+	TokenAuthentication // like bearer authentication but with the word Token
```
lesam, line 109: Do we have code in the config to actually set this?

influxdb/client.go
```diff
@@ -281,19 +309,93 @@ func (c *HTTPClient) do(req *http.Request, result interface{}, codes ...int) (*h
...
-		err := d.Decode(result)
+		err = d.Decode(result)
 		if err != nil {
 			return nil, errors.Wrap(err, "failed to decode JSON")
 		}
 	}
 	return resp, nil
 }
 
+func (c *HTTPClient) doFlux(req *http.Request, codes ...int) (*Response, error) {
```
lesam, line 326: Since this is not implemented anyway - how about having this return a ReadCloser and we can have a QueryFluxResponse to parse and return a `Response`, and a QueryFlux to return a `ResultIterator` ?

We can do this post-merge, wondering what your preference on API is.

services/influxdb/service.go
```diff
@@ -506,7 +506,12 @@ func httpConfig(c Config) (influxdb.Config, error) {
 	}
 	tr := khttp.NewDefaultTransportWithTLS(tlsConfig)
 	var credentials influxdb.Credentials
-	if c.Username != "" {
+	if c.Token != "" {
+		credentials = influxdb.Credentials{
+			Method: influxdb.TokenAuthentication,
```
lesam, line 511: 👍

influxdb/client.go
docmerlin: I like this idea.

lesam:
Tests are still failing:
```
=== RUN   TestServer_UpdateConfig/influxdb/default-0
    server_test.go:9018: unexpected defaults for influxdb/default: influxdb: section element 0 are not equal: unexpected config option difference 
           map[string]interface{}{
          	... // 18 identical entries
          	"subscriptions-sync-interval": string("1m0s"),
          	"timeout":                     string("0s"),
        + 	"token":                       bool(false),
          	"udp-bind":                    string(""),
          	"udp-buffer":                  float64(1000),
          	... // 3 identical entries
          }
 ```

docmerlin:
These commits were merged as part of a different branch, so I am closing this branch.

PR #2549, opened 2021-05-12
fix: Zenoss config handling

alespour:
More fixes to Zenoss:
- `URL` and authentication info (`username`, `password`) removed from handler config
- `collector` property is configurable item as it is unlikely to be changed per alert
- external testing: configuration values are respected during test request handling (`TestOptions()`)

- [x] Rebased/mergable
- [x] Tests pass


services/zenoss/config.go
```diff
@@ -71,6 +71,8 @@ type Config struct {
 	Type string `toml:"type" override:"type"`
 	// Event TID.
 	TID int64 `toml:"tid" override:"tid"`
+	// Collector name.
```
sranka, line 74: Pls remove ServiceNow of `Username` and `Password` property comments

alespour: Done, thanks.

services/zenoss/service.go
```diff
@@ -199,20 +199,12 @@ type Event struct {
 	TID    int64                    `json:"tid"`
 }
 
-func (s *Service) preparePost(url string, state *alert.EventState, data *alert.EventData, hc *HandlerConfig) (string, io.Reader, error) {
+func (s *Service) preparePost(state *alert.EventState, data *alert.EventData, hc *HandlerConfig) (string, io.Reader, error) {
 	c := s.config()
 	if !c.Enabled {
 		return "", nil, errors.New("service is not enabled")
 	}
```
sranka, line 207: I think that it was correct to validate the URL return value in order to avoid a possible future panic, wasn't it? It would be also better to use `url.ParseRequestURI` so that relative URLs and even empty strings are not allowed. 

```suggestion

	if _, err := neturl.ParseRequestURI(c.URL); err != nil {
		return "", nil, err
	}
```

sranka (changes requested):

etc/kapacitor/kapacitor.conf
```diff
@@ -370,6 +370,8 @@ default-retention-policy = ""
   type = "rpc"
   # Temporary request transaction ID.
   tid = 1
+  # Collector name.
+  # collector = ""
```
sranka, line 374: A reasonable default could be provided
```suggestion
collector = "Kapacitor"
```

services/zenoss/service.go
alespour: Config is validated elsewhere, so I think it is not necessary to validate the URL here. I've replaced `Parse` with `ParseRequestURI`.
[services/zenoss/config.go#L101](https://github.com/bonitoo-io/kapacitor/blob/d7a0c037c82650fa6639ae0bcd2debc9a1198176/services/zenoss/config.go#L101)

sranka (approved):

sranka (changes requested):
Can you please also rename json properties in https://github.com/influxdata/kapacitor/blob/master/services/zenoss/service.go#L87-L88 , so that these props are always the same (evclasskey, evclass)

sranka (changes requested):
You also might consider changing "TID" to "tid" in

https://github.com/influxdata/kapacitor/blob/master/pipeline/tick/alert.go#L266

so it consistent with config and other namings of the same property

alespour:
@sranka json props names unified.

sranka (approved):

PR #2550, opened 2021-05-13
feat: Flux batch queries in TICKscripts

docmerlin:


batch.go
```diff
@@ -81,35 +90,67 @@ func (n *BatchNode) Count() int {
...
+
+		case *FluxQueryNode:
+			qs, err := qn.Queries(start, stop)
+			if err != nil {
+				return nil, err
+			}
+			queries[i] = BatchQueries{
+				FluxQueries: qs,
+				Cluster:     qn.Cluster(),
+				//GroupByMeasurement: qn.GroupByMeasurement(),
```
lesam, line 147: Commented code - what does this do?

batch.go
```diff
@@ -522,3 +563,225 @@ func (c *cronTicker) Stop() {
...
+		qstop := current.Add(-1 * n.b.Offset)
+		if qstop.After(now) {
+			break
+		}
+
+		q, err := n.query.Clone()
+		if err != nil {
+			return nil, err
+		}
+		q.Now = time.Now()
```
lesam, line 661: Should we use the same `now` as above?

batch.go
```diff
@@ -522,3 +563,225 @@ func (c *cronTicker) Stop() {
...
+	n.pointsQueried = &expvar.Int{}
+
+	n.statMap.Set(statsBatchesQueried, n.batchesQueried)
+	n.statMap.Set(statsPointsQueried, n.pointsQueried)
+
+	if n.et.tm.InfluxDBService == nil {
+		return errors.New("InfluxDB not configured, cannot query InfluxDB for batch query")
+	}
+
+	con, err := n.et.tm.InfluxDBService.NewNamedClient(n.b.Cluster)
```
lesam, line 680: Should we use the internal execution engine that tasks uses, or stick with sending to a remote influxdb?

cmd/kapacitor/main.go
```diff
@@ -348,11 +348,12 @@ var (
 	rbId             = recordBatchFlags.String("recording-id", "", "The ID to give to this recording. If not set an random ID is chosen.")
 
 	recordQueryFlags = flag.NewFlagSet("record-query", flag.ExitOnError)
-	rqQuery          = recordQueryFlags.String("query", "", "The query to record.")
-	rqType           = recordQueryFlags.String("type", "", "The type of the recording to save (stream|batch).")
-	rqCluster        = recordQueryFlags.String("cluster", "", "Optional named InfluxDB cluster from configuration.")
-	rqNowait         = recordQueryFlags.Bool("no-wait", false, "Do not wait for the recording to finish.")
-	rqId             = recordQueryFlags.String("recording-id", "", "The ID to give to this recording. If not set an random ID is chosen.")
+	// TODO: queryFlux recording
```
lesam, line 351: Are we fixing the TODO in this PR? Is there an issue for it?

fluxquery.go
```diff
@@ -0,0 +1,34 @@
...
+}
+
+// Deep clone this query
+func (q *QueryFlux) Clone() (*QueryFlux, error) {
+	q.queryMu.Lock()
+	defer q.queryMu.Unlock()
+	return &QueryFlux{
+		stmt:  q.stmt,
+		org:   q.org,
+		orgID: q.orgID,
```
lesam, line 27: Why don't we clone `Now` ?

influxdb/csv.go
```diff
@@ -0,0 +1,226 @@
...
+	defaultVals    []string
+	Err            error
+	buf            []imodels.Row
+	seriesBuf      *imodels.Row
+	tableNum       string
+}
+
+const skipNonDataFluxCols = 3
+
+func (q *queryCSVResult) response() (*Response, error) {
```
lesam, line 53: Why can't this be re-written as:

```
r := <original io.Reader in NewFluxQueryResponse>
dec := csv.NewMultiResultDecoder(csv.ResultDecoderConfig{})
fluxResultIterator := dec.Decode(r)
for fluxResultIterator.More() {
  res := results.Next()
  if err := res.Tables().Do(func(tbl flux.Table) error {
    // TODO: process flux.Table into Response, see https://pkg.go.dev/github.com/influxdata/flux#Table and https://github.com/influxdata/flux/blob/b9b7f50e098a28318b6dee5032a2da18d51eab9d/execute/format.go#L95
    return err
  }); err != nil {
    return nil, err
  }
}
```

If we can't use the existing parser, let's talk to the flux team about how to get some parts of this parser into the `flux` library - I don't want every application that integrates against flux to need to re-write a parser starting with `csv.Reader`.

node.go
```diff
@@ -205,6 +205,7 @@ func (n *node) Wait() error {
 	defer n.finishedMu.Unlock()
 	if !n.finished {
 		n.finished = true
+		// deadlocking here
```
lesam, line 208: Still?

docmerlin: No, thanks, forgot to remove the comment.

batch.go
docmerlin: Sorry that code was supposed to be deleted as GroupByMeasurement isn't meaningful for flux queries.

batch.go
docmerlin: Yes, yes we should.

cmd/kapacitor/main.go
docmerlin: https://github.com/influxdata/kapacitor/issues/2565

fluxquery.go
docmerlin: It shouldn't matter as that time never gets used, but you are right, for completeness and readability it is better to also clone the now time.

docmerlin (commented):
Your changes look good to me.

tools.go
```diff
@@ -8,6 +8,8 @@ package kapacitor
 import (
 	_ "github.com/benbjohnson/tmpl"
 	_ "github.com/golang/protobuf/protoc-gen-go"
-	_ "github.com/influxdata/pkg-config" // so we can use the rust dependencies of flux
+
+	// so we can use the rust dependencies of flux
+	_ "github.com/influxdata/pkg-config"
```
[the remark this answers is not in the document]
lesam: I just use `gci` which is the same formatter 2.x is using. https://github.com/daixiang0/gci

task.go
```diff
@@ -234,7 +234,9 @@ func (et *ExecutingTask) start(ins []edge.StatsEdge, snapshot *TaskSnapshot) err
 
 func (et *ExecutingTask) stop() (err error) {
 	close(et.stopping)
+	i := 0
```
lesam, line 237: This is unused?

integrations/helpers_test.go
```diff
@@ -20,6 +20,14 @@ import (
 	"github.com/influxdata/kapacitor/uuid"
 )
 
+func mustParseTime(s string) time.Time {
```
lesam, line 23: Looks like this is defined twice?

batch.go
```diff
@@ -522,3 +562,225 @@ func (c *cronTicker) Stop() {
...
+				Org:   n.query.org,
+				OrgID: n.query.orgID,
+				Now:   n.query.Now,
+			})
+			if err != nil {
+				n.diag.Error("error executing query", err)
+				n.timer.Stop()
+				break
+			}
+			_ = resp
```
lesam, line 708: dead?

services/influxdb/service_test.go
```diff
@@ -1263,19 +1260,42 @@ func (c influxDBClient) Ping(ctx context.Context) (time.Duration, string, error)
...
 }
 
+func (c influxDBClient) QueryFlux(q influxcli.FluxQuery) (flux.ResultIterator, error) {
+	if c.FluxQueryFunc != nil {
+		return c.FluxQueryFunc(c.clusterName, q)
+	}
+	return nil, nil
+}
+
+func (c influxDBClient) QueryFluxResponse(q influxcli.FluxQuery) (*influxcli.Response, error) {
```
lesam, line 1292: I don't see a real implementation of this?

task.go
docmerlin: yes

integrations/helpers_test.go
docmerlin: It exists in integrations test and in a different tests.  I went the coping route, rather than creating a helper package for one tiny function that only exists to make tests easier to read.

services/influxdb/service_test.go
docmerlin: It isn't part of this PR it was already in master.
https://github.com/influxdata/kapacitor/blob/226f1cad9207676c5fd86699b27620c31fbff6ae/influxdb/client.go#L616

batch.go
docmerlin: Yes.

batch.go
lesam: Answer: yes we should, but let's merge this first.

edge/messages.go
```diff
@@ -819,17 +819,21 @@ func ResultToBufferedBatches(res influxdb.Result, groupByName bool) ([]BufferedB
...
 				if c == "time" {
-					tStr, ok := v[i].(string)
-					if !ok {
-						return nil, fmt.Errorf("unexpected time value: %v", v[i])
-					}
-					var err error
-					t, err = time.Parse(time.RFC3339Nano, tStr)
-					if err != nil {
-						t, err = time.Parse(time.RFC3339, tStr)
+					//tStr, ok := v[i].(string)
```
lesam, line 822: Please fix this in the follow-up

lesam (approved):
lgtm

PR #2563, opened 2021-05-28
build: upgrade to go1.17

docmerlin:


lesam (approved):
lgtm

docmerlin:
I'm putting this on hold as it breaks cross builds for some reason.

Dockerfile_build_ubuntu64
```diff
@@ -1,5 +1,5 @@
 ARG GO_VERSION
-FROM quay.io/influxdb/cross-builder:go${GO_VERSION}-latest
+FROM quay.io/influxdb/cross-builder:go${GO_VERSION}-326be487b2cd7a2df71767dab51395feec33911e
```
danxmoran, line 2: NOTE: This is from a branch in the `edge` repo, pushed to confirm it's working as expected

.circleci/config.yml
```diff
@@ -1,92 +1,118 @@
...
-                  - setup_remote_docker:
-                        docker_layer_caching: true
-                  - run_tests
+  build:
+    executor: linux-amd64
+    parallelism: 2
+    working_directory: ~/kapacitor
+    steps:
+      - checkout
+      - upgrade_docker
```
lesam, line 71: I think you should be able to request a specific VM image and avoid the docker upgrade

build.py
```diff
@@ -213,7 +231,7 @@ def run_tests(race, parallel, timeout, no_vet):
             return False
     else:
         logging.info("Skipping 'go vet' call...")
-    test_command = "go test -v"
+    test_command = "go test"
```
danxmoran, line 234: Might want to add this back, but under a flag (it annoyed me locally but is useful in CI to see that it's not frozen)

.circleci/config.yml
danxmoran: :+1: the image I've pinned in the `executors` definition is the same as used in influxdb 2.x / influx-cli, and I remember its pre-installed Docker being pretty old. I'll check if there's a newer image out that we can try.

Dockerfile_build_ubuntu64
danxmoran: Will need updated after influxdata/edge#109 merges

build.py
lesam: I think we do want it back for CI, there I want all the info I can get after the fact

build.py
```diff
@@ -188,9 +188,33 @@ def check_nochanges():
...
+        download_cmd = ["go", "mod", "download"]
+        if verbose:
+            download_cmd.append("-x")
+        subprocess.check_output(download_cmd)
+    except subprocess.CalledProcessError as exc:
+        logging.error("Downloading dependencies failed, see logs for details")
+        logging.error("{}".format(exc.output))
+    # NOTE: We vendor deps here because TestPipelineImplemented in pipeline/tick/tick_test.go uses `importer.For` to
+    # load a few packages from source, and it recursively descends through dependencies. It's apparently not module-
+    # aware and only looks under $GOPATH/src and vendor/.
```
lesam, line 207: Looks like `ForCompiler` in that package is module-aware - can we have that added to the comment here so when we fix it we know where to look?

udf/agent/server.go
```diff
@@ -74,7 +74,7 @@ func (s *Server) StopOnSignals(signals ...os.Signal) {
 	s.wg.Add(1)
 	s.mu.Unlock()
 
-	c := make(chan os.Signal)
+	c := make(chan os.Signal, 1)
```
lesam, line 77: Why?

danxmoran: It failed `go vet`, reasoning explained in the release notes [here](https://golang.org/doc/go1.17#vet).

server/server_const_windows_test.go
```diff
@@ -1,3 +1,4 @@
+//go:build windows
```
lesam, line 1: Why the changed build tag format? Do we need both?

danxmoran: Yes, it's auto-generated by `go fmt` on 1.17 and `go vet` complains if you try to delete the old form. Notes start [here](https://golang.org/doc/go1.17#gofmt).

Dockerfile_build_ubuntu64
```diff
@@ -1,5 +1,5 @@
 ARG GO_VERSION
```
lesam, line 1: Excited about every change making this smaller. Looks like all the rest of this could be pushed up to the edge repo too - maybe in a follow-up PR?

danxmoran: For now, I'll delete all the things from the `apt-get install` that are already covered by the cross-builder.

lesam (approved):
LGTM, but would be nice to have the ForCompiler comment

PR #2564, opened 2021-06-01
fix: null pointer panic in scraper handler

docmerlin:


lesam (approved):

gwossum (approved):
Appears to follow pattern from other scrapers of making sure buf is allocated, so looks good to me

PR #2575, opened 2021-06-10
feat(alerta-service): Support the "attributes" attribute in Alerta node

sp1r:
Fix influxdata#2349

This commit adds support for custom Alerta alert attributes (see Alerta Alert Format for details).

Attributes are specified with "attribute" keyword in tick script, e.g.:

    |alert()
        // ...
        .alerta()
            // ...
            .attribute('customUrl', 'https://example.com/')
            .attribute('booleanAttribute', TRUE)
            .attribute('numericAttribute', 9001)
            .attribute('templatedAttribute', 'Templates are supported: {{ .Name }}')

- [x] Rebased/mergable
- [x] Tests pass
- [x] CHANGELOG.md updated

### DIFF
--- a/pipeline/alert.go
+++ b/pipeline/alert.go
@@ -1263,6 +1263,9 @@ type AlertaHandler struct {
 	// List of Correlated
 	Correlate []string `tick:"Correlated" json:"correlate"`
 
+	// Map of alert Attributes
+	Attributes map[string]interface{} `tick:"Attribute" json:"attributes"`
+
 	// Alerta timeout.
 	// Default: 24h
 	Timeout time.Duration `json:"timeout"`
@@ -1278,6 +1281,14 @@ func (a *AlertaHandler) Services(service ...string) *AlertaHandler {
 
 func (a *AlertaHandler) Correlated(correlate ...string) *AlertaHandler {
 	a.Correlate = correlate
+	return a
+}
+
+func (a *AlertaHandler) Attribute(k string, v interface{}) *AlertaHandler {
+	if a.Attributes == nil {
+		a.Attributes = make(map[string]interface{})
+	}
+	a.Attributes[k] = v
 	return a
 }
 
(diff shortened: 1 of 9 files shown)


sp1r:
Tested this changes with Alerta API 8.6.0 (latest)

pipeline/alert.go
```diff
@@ -1263,6 +1263,9 @@ type AlertaHandler struct {
 	// List of Correlated
 	Correlate []string `tick:"Correlated" json:"correlate"`
 
+	// Map of alert Attributes
+	Attributes map[string]string `tick:"Attribute" json:"attributes"`
```
docmerlin, line 1267: In alerta attributes can be of different types (string, number, bool, etc), so it would be better if it was `map[string]interface{}`

services/alerta/alertatest/alertatest.go
```diff
@@ -53,14 +53,15 @@ type Request struct {
...
-	Timeout     int64    `json:"timeout"`
+	Resource    string            `json:"resource"`
+	Event       string            `json:"event"`
+	Group       string            `json:"group"`
+	Environment string            `json:"environment"`
+	Text        string            `json:"text"`
+	Origin      string            `json:"origin"`
+	Service     []string          `json:"service"`
+	Correlate   []string          `json:"correlate"`
+	Attributes  map[string]string `json:"attributes"`
```
docmerlin, line 64: probably want `map[string]interface{}`

docmerlin (changes requested):
Thanks for the PR: The alerta API allows for other types than string in the attributes, this should really support that. 
Other than that, I really like this PR and would like to get it into 1.6.3

pipeline/alert.go
sp1r: Nice point! I've completely missed this behaviour in Alerta. On my way with appropriate changes.

sp1r:
Looks like builds on master branch are broken.

docmerlin (approved):

mustaqeem:
Hi @docmerlin, when can we expect the v1.6.3 release?

PR #2579, opened 2021-06-14
fix: cli auth and error handling for flux tasks

### LINKED ISSUE #2578: Kapacitor cli does not handle authentication properly with flux tasks
Authentication is not handled properly as we do not parse the URL properly to parse out and forward the `Basic` authentication, if present.

Additionally we do not handle authentication errors well in the flux CLI as it they do not match the OpenAPI schema.

lesam:
Closes #2578 

Before:

➜  ./kapacitor -url http://steve:pas@localhost:9092 flux task list
undefined response type
➜  ./kapacitor -url http://steve:pass@localhost:9092 flux task list
undefined response type 

After:

➜  ./kapacitor -url http://steve:pas@localhost:9092 flux task list
Error: authorization failed
➜  ./kapacitor -url http://steve:pass@localhost:9092 flux task list
ID                      Name    Organization ID Organization    Status  Every   Cron
079625bc3aeb8000        poster                                  active  3s

- [x] Rebased/mergable
- [ ] Tests pass
- [ ] CHANGELOG.md updated

### DIFF
--- a/services/httpd/handler.go
+++ b/services/httpd/handler.go
@@ -555,13 +555,17 @@ func serveExpvar(w http.ResponseWriter, r *http.Request) {
 
 // HttpError writes an error to the client in a standard format.
 func HttpError(w http.ResponseWriter, err string, pretty bool, code int) {
+	w.Header().Set("Content-Type", "application/json; charset=utf-8")
 	w.WriteHeader(code)
 
 	type errResponse struct {
+		// Kapacitor 1.5 used only 'error'
 		Error string `json:"error"`
+		// Kapacitor 1.6 also uses 'message' to match InfluxDB 2.x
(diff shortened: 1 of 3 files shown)


docmerlin (approved):

services/httpd/handler.go
```diff
@@ -555,13 +555,17 @@ func serveExpvar(w http.ResponseWriter, r *http.Request) {
 
 // HttpError writes an error to the client in a standard format.
 func HttpError(w http.ResponseWriter, err string, pretty bool, code int) {
+	w.Header().Set("Content-Type", "application/json")
```
docmerlin, line 558: This is unnecessary, there is something that is already taking care of this, and it collides with that.

PR #2589, opened 2021-07-06
feat: flag to allow blacklisting CIDR ranges

docmerlin:
This allows blacklisting CIDR ranges for various post/get operations.

### DIFF
--- a/http/transport.go
+++ b/http/transport.go
@@ -2,13 +2,28 @@ package http
 
 import (
 	"crypto/tls"
+	"fmt"
 	"net"
 	"net/http"
+	"net/url"
+	"strings"
+	"syscall"
 	"time"
+
+	furl "github.com/influxdata/flux/dependencies/url"
 )
 
 // NewDefaultTransport creates a new transport with sane defaults.
-func NewDefaultTransport() *http.Transport {
+func NewDefaultTransport(dialer *net.Dialer) *http.Transport {
+
+	if dialer == nil {
+		dialer = &net.Dialer{
+			Timeout:   30 * time.Second,
+			KeepAlive: 30 * time.Second,
+			DualStack: true,
+		}
+	}
+
 	// These defaults are copied from http.DefaultTransport.
 	return &http.Transport{
 		Proxy: http.ProxyFromEnvironment,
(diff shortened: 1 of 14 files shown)


http/transport.go
```diff
@@ -27,8 +42,100 @@ func NewDefaultTransport() *http.Transport {
...
+		return urlValidator.ValidateIP(ip)
+	}
+}
+
+// NewDefaultClientWithTLS creates a tls client with sane defaults.
+func NewDefaultClientWithTLS(tlsConfig *tls.Config, urlValidator furl.Validator) *http.Client {
+
+	// These defaults are copied from http.DefaultTransport.
+	return &http.Client{
+		Transport: NewDefaultTransport(&net.Dialer{
```
wbaker85, line 77: Is this supposed to use the `tlsConfig` ... like with `NewDefaultTransportWithTLS`?

http/transport.go
```diff
@@ -2,13 +2,28 @@ package http
...
+
+	furl "github.com/influxdata/flux/dependencies/url"
 )
 
 // NewDefaultTransport creates a new transport with sane defaults.
-func NewDefaultTransport() *http.Transport {
+func NewDefaultTransport(dialer *net.Dialer) *http.Transport {
+
+	if dialer == nil {
+		dialer = &net.Dialer{
```
wbaker85, line 20: Does this `dialer` get used anywhere else in this function? I was thinking it could be used down on line 30 instead of making a new one.

influxdb/client.go
```diff
@@ -243,6 +244,12 @@ func (c *HTTPClient) Update(new Config) error {
 		if tr == nil {
 			tr = old.Transport
 		}
+		tr.DialContext = (&net.Dialer{
+			Timeout:   new.Timeout, // I am not sure if this is the right value to set it to
+			KeepAlive: new.Timeout, // I am not sure if this is the right value to set it to
```
wbaker85, line 249: I'd tend to think that the `KeepAlive` should be set the same here as it is elsewhere, and not necessarily be the same as the timeout. Since it's being set to 30 sec everywhere else to match `http.DefaultClient`, it would make sense to me to have it set to that here as well.

Question: Is it possible to call this function with a new config that has a `nil` or zero-value duration for the timeout? I'm thinking that would make the timeout be unlimited for the dialer (and the client), and am wondering if that's desirable.

wbaker85 (approved):
lgtm!

codyshepherd (approved):

pierwill:
@docmerlin @codyshepherd Is there an example configuration file for this anywhere?

pierwill:
> @docmerlin @codyshepherd Is there an example configuration file for this anywhere?

nvmnd, https://github.com/influxdata/docs-v2/pull/2904 😇

PR #2591, opened 2021-07-07
Disable alert handlers with a flag

docmerlin:
This creates a flag that allows you to disable alert handlers.  This is useful for security reasons such as disabling exec on shared machines.


server/server.go
```diff
@@ -181,8 +184,12 @@ func New(c *Config, buildInfo BuildInfo, diagService *diagnostic.Service) (*Serv
 		tlsConfig = new(tls.Config)
 	}
 	d := diagService.NewServerHandler()
+	if disabledAlertHandlers == nil {
```
lesam, line 187: Just leaving it as nil would be fine too right? It's immutable?

lesam (approved):

server/server.go
docmerlin: Good point.

PR #2592, opened 2021-07-13
fix(bigpanda): fields of alerting data point are serialized to proper types

bednar:
Closes https://github.com/influxdata/EAR/issues/2365

This PR changes how are serialized `fields of alerting` data into BigPanda payload - use original type of field instead mapping a field value to `string`.

- [x] Rebased/mergable
- [x] Tests pass
- [x] CHANGELOG.md updated


idclark:
@bednar is there an instance of BigPanda I can use to test API calls against this build before handing it off to the customer?

docmerlin (approved):
LGTM

PR #2601, opened 2021-08-02
security: switch to github.com/golang-jwt/jwt

docmerlin:
This switches our jwt library because of a security bug in
github.com/dgrijalva/jwt-go


jdstrand (approved):
I've not tried to build this or otherwise look at the PR deeply (so may want to have someone else take a look), but it has the shape used in PRs for other projects and looks the way I would expect it to look. Thanks!

PR #2603, opened 2021-08-03
test: fix flaky TestService_GetConfig #1367

### LINKED ISSUE #1367: Fix TestService_GetConfig
```
-- FAIL: TestService_GetConfig (0.03s)
	service_test.go:1070: expected to get an response on errC
```

docmerlin:
Fix flaky test
Closes #1367

### DIFF
--- a/services/config/service_test.go
+++ b/services/config/service_test.go
@@ -1033,67 +1033,76 @@ func TestService_GetConfig(t *testing.T) {
 			},
 		},
 	}
-	for _, tc := range testCases {
+	for i, tc := range testCases {
 		updates := make(chan config.ConfigUpdate, len(testCases))
 		service, server := OpenNewSerivce(testConfig, updates)
 		defer server.Close()
 		defer service.Close()
 		basePath := server.Server.URL + httpd.BasePath + "/config"
-		// Apply all updates
-		for _, update := range tc.updates {
-			errC := make(chan error, 1)
-			go func() {
-				// Validate we got the update over the chan.
-				// This keeps the chan unblocked.
-				timer := time.NewTimer(100 * time.Millisecond)
-				defer timer.Stop()
-				select {
-				case cu := <-updates:
-					cu.ErrC <- nil
-					errC <- nil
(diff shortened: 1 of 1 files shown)


services/config/service_test.go
```diff
@@ -1033,67 +1033,76 @@ func TestService_GetConfig(t *testing.T) {
...
+				}
+				defer resp.Body.Close()
+				body, err := ioutil.ReadAll(resp.Body)
+				if err != nil {
+					t.Fatal(err)
+				}
+				if got, exp := resp.StatusCode, http.StatusNoContent; got != exp {
+					t.Fatalf("update failed: got: %d exp: %d\nBody:\n%s", got, exp, string(body))
+				}
+				timeOut := time.NewTimer(5 * time.Second)
```
davidby-influx, line 1075: I'm always leery of tests that depend on timers, because they operate differently in different environments, but I don't have any good way to avoid it here.

services/config/service_test.go
```diff
@@ -1033,67 +1033,76 @@ func TestService_GetConfig(t *testing.T) {
...
-				t.Fatalf("update failed: got: %d exp: %d\nBody:\n%s", got, exp, string(body))
-			}
-			select {
-			case err := <-errC:
-				if err != nil {
-					t.Fatal(err)
-				}
-			default:
-				t.Fatal("expected to get an response on errC")
+			defer resp.Body.Close()
```
davidby-influx, line 1092: Can the resp.Body.Close() fail? Should you capture the error and fail the test if so?

davidby-influx (approved):
Non-blocking  comments. LGTM

services/config/service_test.go
docmerlin: I don't either :-( This is just to reduce the fake-fail by letting it happen out of order sometimes.

services/config/service_test.go
docmerlin: Technically it can, but if it fails the problem isn't in the system we are testing, but in the underlying subsystem.

PR #2608, opened 2021-08-12
Allow for compact json in templates

docmerlin:
Some third party services reject standard json that has a carriage return at the end.  This allows you to use compact json in your templates.

Also, it compacts the details in the Big Panda alert, to avoid a bug in Panda's service.


alert.go
```diff
@@ -119,16 +119,36 @@ func newAlertNode(et *ExecutingTask, n *pipeline.AlertNode, d NodeDiagnostic) (a
...
-		"json": func(v interface{}) html.JS {
-
+		"jsonCompact": func(v interface{}) html.JS {
 			tmpBuffer := an.bufPool.Get().(*bytes.Buffer)
+			tmpBuffer2 := an.bufPool.Get().(*bytes.Buffer)
+
 			defer func() {
-				tmpBuffer.Reset()
-				an.bufPool.Put(tmpBuffer)
+				if tmpBuffer.Cap() < (2 << 19) { // only reuse the buffer if it is less than 500kb
```
danxmoran, line 127: Can you pull the `2 << 19` out into a `const` so it's in one place?

alert.go
```diff
@@ -119,16 +119,36 @@ func newAlertNode(et *ExecutingTask, n *pipeline.AlertNode, d NodeDiagnostic) (a
 	}
 
 	an.detailsTmpl, err = html.New("details").Funcs(html.FuncMap{
-		"json": func(v interface{}) html.JS {
-
+		"jsonCompact": func(v interface{}) html.JS {
```
danxmoran, line 122: Worth adding a test for this new case?

danxmoran (approved):

PR #2618, opened 2021-08-31
fix: switch flux formatter to one that preserves comments

danxmoran:
- [x] Rebased/mergable
- [x] Tests pass
- [x] CHANGELOG.md updated


build.py
```diff
@@ -213,6 +213,12 @@ def run_tests(race, parallel, timeout, no_vet):
             return False
     else:
         logging.info("Skipping 'go vet' call...")
+
+    # NOTE: TestPipelineImplemented in pipeline/tick/tick_test.go uses the deprecated importer.For to load
+    # packages from source, requiring either GOPATH to be set or deps to be vendored.
```
danxmoran, line 218: I have no idea how tests passed without this before. I see a few merges to our cross-builder image definition since the last merge to `master` here, but none of the x-builder changes look like they'd have broken things around GOPATH or vendoring deps.

Dockerfile_build_ubuntu64
```diff
@@ -36,7 +36,7 @@ RUN gem install fpm
 # Install protobuf3 protoc binary
 ENV PROTO_VERSION 3.11.1
 RUN wget -q https://github.com/google/protobuf/releases/download/v${PROTO_VERSION}/protoc-${PROTO_VERSION}-linux-x86_64.zip \
-    && unzip -j protoc-${PROTO_VERSION}-linux-x86_64.zip bin/protoc -d /bin \
+    && unzip -o -j protoc-${PROTO_VERSION}-linux-x86_64.zip bin/protoc -d /bin \
```
danxmoran, line 39: `-o` needed to overwrite the `protoc` binary that's included in the x-builder image. We could instead delete this whole block from the dockerfile here, but I was wary about switching from the pinned version here to whatever happens to get pulled from the apt repos.

build.sh
```diff
@@ -20,9 +20,6 @@ dataname=kapacitor-data-$BUILD_NUM
 # Build new docker image
 docker build -f Dockerfile_build_ubuntu64 -t $imagename --build-arg GO_VERSION=${GO_VERSION} $DIR
 
-# Build new docker image
-docker build -f Dockerfile_build_ubuntu64 -t influxdata/kapacitor-builder --build-arg GO_VERSION=${GO_VERSION} $DIR
```
danxmoran, line 24: Drive-by cleanup: image was built but never used.

task/options/options.go
```diff
@@ -45,7 +46,18 @@ type Duration struct {
 }
 
 func (a Duration) String() string {
-	return ast.Format(&a.Node)
+	// NOTE: This is a copy of `formatDurationLiteral` from the flux codebase.
```
skartikey, line 49: Hi Daniel, 
I am adding an exportable method that can be used for formatting Duration. Changes are in the review atm. https://github.com/influxdata/flux/pull/4009

build.py
lesam: Are we pinning the cross-builder image? If we're using 'latest' anywhere that could explain some of the weirdness here - if so we should pin it.

danxmoran: I think all the x-builder changes made to get plutonium building are/will be needed for kapacitor too, but I'll swap `latest` with the hash of the latest build.

Dockerfile_build_ubuntu64
lesam: I'm not too worried if we pin the build - it should be consistent since we pull the same cross build container every time now.

If we still want to pin it, why not in the parent image?

lesam (approved):

Dockerfile_build_ubuntu64
lesam: btw from the protobuf docs:

```
A new notion "syntax" is introduced to specify whether a .proto file
uses proto2 or proto3:

// foo.proto
syntax = "proto3";
message Bar {...}
If omitted, the protocol buffer compiler generates a warning and "proto2" is
used as the default. This warning will be turned into an error in a future
release.
```

Hopefully we use `syntax` everywhere but if we don't then it is still backward-compatible with newer versions of the protobuf compiler.

lesam:
LGTM but would be nice to at least move the protobuf stuff up to the edge container and pull in the Duration formatting fix.

Dockerfile_build_ubuntu64
danxmoran: I'll just use the version from the cross-builder

danxmoran: Ah wait, that is if it includes the python libraries... need to check

danxmoran: Scratch that, will update the cross-builder image to use the version pinned here & install the matching c++ runtime. This image will still have access to the PROTO_VERSION env var from the cross-builder, so it can use that to also download & install the python runtime.

PR #2621, opened 2021-09-06
feat(tasks): add template-id to task list

### LINKED ISSUE #2619: GET /kapacitor/v1/tasks does not return task template-id
__Actual behavior:__

`GET /kapacitor/v1/tasks` does not return `template-id` in the returned task list, it is not a supported field per https://github.com/influxdata/kapacitor/blob/v1.6.1/services/task_store/service.go#L487 . `template-id` is however properly returned when getting a particular task (https://github.com/influxdata/kapacitor/blob/v1.6.1/services/task_store/service.go#L1134). 

__Expected behavior:__
- `template-id` is OOTB returned in `GET /kapacitor/v1/tasks`
-  the kapactior swagger file adds (a read-only) `template-id` to the Task object, it should be already there since the `get task` operation returns it.

__Rationale:__
 Tasks that were created from a template must be read-only in chronograf  so that the users are not confused with something that cannot be effectively edited (in chronograf), this is already fixed in https://github.com/influxdata/chronograf/pull/5805 .  Additionally, templated tasks cannot be ever chronograf rules (their TICKScripts were not created by chronograf). This is already taken into account in the chronograf code, but it does not work prefectly because of the missing template-id in the task list. Chronograf now has to spend useless time with parsing and processing such templated tasks. If the task TICKScript can be parsed as a chronograf rule, the task then appears in the list of chronograf rules and can be clicked as a rule. The user is then however informed that it is not a chronograf rule that could be possibly managed. Fixing this issue would also fix such a weird user experience on the chronograf side.

sranka:
Closes #2619 

This PR adds template-id OOTB to the list of returned tasks. It includes also tests, an updated swagger file, and housekeeping of staticcheck errors on updated files.

Adding template-id to task list  helps chronograf with processing and visualization of templated tasks. Details are explained in  influxdata/chronograf#5807

- [x] Rebased/mergable
- [x] Tests pass
- [x] CHANGELOG.md updated

### DIFF
--- a/client/v1/client_test.go
+++ b/client/v1/client_test.go
@@ -833,7 +833,8 @@ func Test_ListTasks(t *testing.T) {
 		"type":"stream",
 		"dbrps":[{"db":"db","rp":"rp"}],
 		"status" : "disabled",
-		"executing" : false
+		"executing" : false,
+		"template-id": "myTemplate"
 	},
 	{
 		"link": {"rel":"self", "href":"/kapacitor/v1/tasks/t2"},
@@ -881,6 +882,7 @@ func Test_ListTasks(t *testing.T) {
 			Status:         client.Disabled,
 			Executing:      false,
 			ExecutionStats: client.ExecutionStats{},
+			TemplateID:     "myTemplate",
 		},
 		{
 			Link: client.Link{Relation: client.Self, Href: "/kapacitor/v1/tasks/t2"},
--- a/client/v1/client_test.go
+++ b/client/v1/client_test.go
@@ -899,7 +899,7 @@ func Test_ListTasks(t *testing.T) {
 					"throughput": 5.6,
 				},
 				NodeStats: map[string]map[string]interface{}{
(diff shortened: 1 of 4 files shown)


client/v1/client_test.go
```diff
@@ -7,10 +7,10 @@ import (
 	"io/ioutil"
 	"net/http"
 	"net/http/httptest"
-	"reflect"
 	"testing"
 	"time"
 
+	"github.com/google/go-cmp/cmp"
```
lesam, line 13: Why the library change?

lesam (approved):

client/v1/client_test.go
sranka: VS.Code linter warned about the usage of `reflect.DeepEqual`. It was probably a false warning since the `reflect.DeepEqual` caveats did not apply to the cases used in the test.  I appeased the linter, https://pkg.go.dev/github.com/google/go-cmp/cmp should be rather used in tests.

PR #2622, opened 2021-09-14
feat: auto-create 1.x DB or 2.x bucket for flux task logs

### LINKED ISSUE #2620: Support auto-creation of InfluxDB database for flux task logs from kapacitor
https://docs.influxdata.com/kapacitor/v1.6/administration/configuration/#flux-tasks

Kapacitor uses a remote influxdb bucket (or 1.x database) to store task run logs. Currently, users need to run a manual `CREATE DATABASE` or 2.x equivalent in order to set this up.

This is not great UX, ideally kapacitor would create the bucket/database automatically.

Moving parts:
* Retention: we can use ~1 month retention as a default EDIT: changed to infinite by default, can be changed any time by the user.
* Deal with edge cases:
  * Database already exists but retention policy is missing
  * Database already exists so we shouldn't create it
  * The kapacitor user doesn't have permissions to create databases

lesam:
Closes #2620

Flux task logs are stored in the "Analytic Store", which is backed
by an influxdb 1.x Database or 2.x Bucket.

Auto-creation gives an infinite retention policy, which can be adjusted later by the user.

Note that for 1.x, "database/rp" bucket names will not be autocreated, only
bare "database" bucket names. This could potentially be added in the future
but the assumption is if a user wants control over the retention policy name
they probably want to control the retention policy creation as well.

Tested manually against 1.x and 2.x.

- [x] Rebased/mergable
- [x] Tests pass
- [x] CHANGELOG.md updated

### DIFF
--- a/services/influxdb/service.go
+++ b/services/influxdb/service.go
@@ -563,6 +563,11 @@ func (c *influxdbCluster) Open() error {
 
 	c.watchSubs()
 
+	// Even if we're not linking subscriptions, validate the client to ensure influxdb is actually up
+	if err := c.validateClientWithBackoff(ctx); err != nil {
+		return errors.Wrap(err, "failed to validate influxdb client on startup")
+	}
+
 	if err := c.linkSubscriptions(ctx, c.subName); err != nil {
 		return errors.Wrap(err, "failed to link subscription on startup")
 	}
--- a/task/backend/analytical_storage.go
+++ b/task/backend/analytical_storage.go
@@ -6,6 +6,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"strings"
 	"time"
 
 	"github.com/influxdata/flux"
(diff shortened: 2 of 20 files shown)


task/backend/analytical_storage.go
```diff
@@ -29,6 +30,71 @@ const (
...
+				return err
+			}
+			for result.More() {
+				if err := result.Next().Tables().Do(func(table flux.Table) error {
+					return table.Do(func(col flux.ColReader) error {
+						return nil
+					})
+				}); err != nil {
+					return err
+				}
```
gwossum, line 57: If I'm reading this loop correctly, if we only succeed in creating the bucket on the last attempt, we will actually return an error even if the bucket was created. The bucket check and no error return is done at the top of the loop, and on the last iteration we can't get back to the top to check for success. Doesn't seem likely to happen in the real life, but might be confusing when it does.

task/config.go
```diff
@@ -54,8 +57,5 @@ func (c Config) Validate() error {
 	if len(c.TaskRunOrgID) > 0 && len(c.TaskRunOrg) > 0 {
 		return fmt.Errorf("only one of task-run-org and task-run-orgid should be set")
 	}
-	if len(c.TaskRunBucket) == 0 {
-		return fmt.Errorf("task-run-bucket is required")
-	}
```
gwossum, line 59: Why is this check no longer necessary? What if the config file had `task-run-bucket =""`?

lesam: In that case we want to use the default - otherwise people who generated a default config from 1.6.1 would not get auto-created buckets.

gwossum: That's what I thought, thanks for confirming

task/backend/analytical_storage.go
lesam: fixed

task/backend/analytical_storage.go
```diff
@@ -29,6 +30,79 @@ const (
...
+		if try > numTries {
+			break
+		}
+
+		// First check if we're reading any flux queries at all
+		err = queryForError("buckets()")
+		if err != nil {
+			log.Warn("InfluxDB instance for analytic store not processing flux queries", zap.Error(err))
+			continue
+		}
```
gwossum, line 81: From the comment, my impression is that if the query fails that means flux is not enabled, so we don't need to create the task run bucket. However, since we `continue` instead of `break`ing or `returning`ing, we'll do this check 4 more times, and increase the startup time by 4 seconds or so.

I'm guessing there could be some errors, like a timeout or connection refused, that warrant retrying this. Are there other errors that allow us to escape from the loop early?

lesam: If this check fails permanently, we fail kapacitor startup (we already fail kapacitor startup if the influxdb instance is not responding), so I think a bit of extra waiting before failing is ok.

services/influxdb/service.go
```diff
@@ -563,6 +563,11 @@ func (c *influxdbCluster) Open() error {
 
 	c.watchSubs()
 
+	// Even if we're not linking subscriptions, validate the client to ensure influxdb is actually up
+	if err := c.validateClientWithBackoff(ctx); err != nil {
```
lesam, line 567: Note this pings the configured influxdb and happens before we try to set up the flux task service.

task/backend/analytical_storage.go
gwossum: In that case, my interpretation is that InfluxDB is up and running, and this check tells us Flux is not enabled, so we should not attempt to create the default database. If that's true, wouldn't `return nil` be appropriate here to speed up starting Kapacitor when Flux is disabled on the InfluxDB instance? Or am I misinterpreting something? Would we never even call this function if Flux is disabled?

lesam: If you've told kapacitor to store its task logs in InfluxDB, flux needs to be enabled. Configuring Kapacitor to store task logs to an InfluxDB with flux disabled is a serious misconfiguration and we should refuse to start I think - certainly the task logs will not save correctly.

gwossum: In that case, should we return a more helpful `Flux must be enabled` error instead of `could not create database for analytic store`?

lesam: Both errors will just be logged (since the error will be returned to the startup code) - I think `log.Warn("InfluxDB instance for analytic store not processing flux queries", zap.Error(err))` is sufficient.

lesam: This looks like:

```
ts=2021-09-16T14:47:54.666-04:00 lvl=info msg="Error reading from bucket for task analytic storage, attempting to create it" service=fluxtask-analytical-store name=kapacitor_fluxtask_logs error="Flux query service disabled. Verify flux-enabled=true in the [http] section o
f the InfluxDB config." errorVerbose="Flux query service disabled. Verify flux-enabled=true in the [http] section of the InfluxDB config.\ngithub.com/influxdata/kapacitor/influxdb.(*HTTPClient).doHttpV2\n\t/Users/sarnold/influx/kapacitor/influxdb/client.go:495\ngithub.com
/influxdata/kapacitor/influxdb.(*HTTPClient).QueryFlux\n\t/Users/sarnold/influx/kapacitor/influxdb/client.go:731\ngithub.com/influxdata/kapacitor/task/backend.CreateDestination.func1\n\t/Users/sarnold/influx/kapacitor/task/backend/analytical_storage.go:42\ngithub.com/infl
uxdata/kapacitor/task/backend.CreateDestination\n\t/Users/sarnold/influx/kapacitor/task/backend/analytical_storage.go:65\ngithub.com/influxdata/kapacitor/task/backend.NewAnalyticalStorage\n\t/Users/sarnold/influx/kapacitor/task/backend/analytical_storage.go:141\ngithub.co
m/influxdata/kapacitor/services/fluxtask.(*Service).Open\n\t/Users/sarnold/influx/kapacitor/services/fluxtask/service.go:71\ngithub.com/influxdata/kapacitor/server.(*Server).startServices\n\t/Users/sarnold/influx/kapacitor/server/server.go:1143\ngithub.com/influxdata/kapa
citor/server.(*Server).Open\n\t/Users/sarnold/influx/kapacitor/server/server.go:1125\ngithub.com/influxdata/kapacitor/cmd/kapacitord/run.(*Command).Run\n\t/Users/sarnold/influx/kapacitor/cmd/kapacitord/run/command.go:156\nmain.(*Main).Run\n\t/Users/sarnold/influx/kapacito
r/cmd/kapacitord/main.go:82\nmain.main\n\t/Users/sarnold/influx/kapacitor/cmd/kapacitord/main.go:42\nruntime.main\n\t/Users/sarnold/.gvm/gos/go1.17/src/runtime/proc.go:255\nruntime.goexit\n\t/Users/sarnold/.gvm/gos/go1.17/src/runtime/asm_amd64.s:1581"
...
ts=2021-09-16T14:47:54.670-04:00 lvl=error msg="encountered error" service=run err="open server: open service *fluxtask.Service: creating analytical store: could not create database for analytic store, tried 5 times"
```

gwossum: I'll accept this

gwossum (approved):
Approved

PR #2632, opened 2021-09-24
build: Import the gpg key required for package signing

codyshepherd:
- [x] Rebased/mergable
- [ ] Tests pass

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -23,6 +23,10 @@ commands:
                     value: ${CIRCLE_TAG}
           steps:
             - run:
+                name: Import GPG key
+                command: |
+                  echo -e "$GPG_1X_PRIVATE_KEY" | gpg --import --
+            - run:
                 name: Deploy Release Packages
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -22,6 +22,11 @@ commands:
                     pattern: ".*rc[0-9]+$"
                     value: ${CIRCLE_TAG}
           steps:
+            - run:
+                name: Import GPG key
+                command: |
+                  echo -e "$GPG_1X_PRIVATE_KEY" > private.key
+                  gpg --batch --import private.key
```
docmerlin, line 29: this looks good, but it looks like it is being imported as a file into the build environment at the same level as kappa's folder, are we sure it won't leak into packages/containers that are released

pierwill: Could we add a step to `rm` the file, or pipe the key directly to `gpg`?

pierwill (approved):

PR #2640, opened 2021-10-04
perf(ingestion): switch from using formvalue to query params

docmerlin:
This cleans up ingestion by getting rid of the `http.Request.FormValue` usages and switching to just looking at query params.

This shouldn't affect perf very much, as  `http.Request.FormValue` short circuits to query params if the right content types are not set. Still this is clearer and avoids quite a few calls.


davidby-influx (approved):
LGTM

PR #2641, opened 2021-10-04
test(tick): remove depreciated importer.For

docmerlin:
This removes the depreciated importer.For.  This removal allows tests to be module aware.


codyshepherd (approved):
I don't really understand what's going on here, so I'm taking your word for it.

PR #2650, opened 2021-11-01
feat(alerts): templating for address in email alert

docmerlin:
This adds templating email addresses for email alerts.


integrations/streamer_test.go
```diff
@@ -10758,108 +10758,217 @@ stream
...
+				Enabled: true,
+				Host:    smtpServer.Host,
+				Port:    smtpServer.Port,
+				From:    "test@example.com",
+			}
+			smtpService := smtp.NewService(sc, diagService.NewSMTPHandler())
+			if err := smtpService.Open(); err != nil {
+				t.Fatal(err)
+			}
+			defer smtpService.Close()
```
davidby-influx, line 10941: Do we want to check the error here on `Close()`?

integrations/streamer_test.go
```diff
@@ -10758,108 +10758,217 @@ stream
...
+			}
 
-	tmInit := func(tm *kapacitor.TaskMaster) {
-		tm.SMTPService = smtpService
-	}
+			testStreamerNoOutput(t, "TestStream_Alert", script, 13*time.Second, tmInit)
 
-	testStreamerNoOutput(t, "TestStream_Alert", script, 13*time.Second, tmInit)
+			// Close both client and server to ensure all message are processed
+			smtpService.Close()
```
davidby-influx, line 10950: Check the error?

services/smtp/config.go
```diff
@@ -57,6 +59,8 @@ func (c Config) Validate() error {
 		return fmt.Errorf("invalid from email address: %q", c.From)
 	}
 	for _, t := range c.To {
+		// this is technically not correct, but I have only seen one email address without an `@` ever, in the modern day
+		// this is a safe, bet... no one uses bang paths anymore (also many libraries and services don't support them)
```
davidby-influx, line 63: How about using [mail.ParseAddress](https://pkg.go.dev/net/mail#ParseAddress) for more robust validation?

integrations/streamer_test.go
davidby-influx: I think a `Close` error should be fatal for a test.

davidby-influx (changes requested):
Check for errors in tests and fail if they are encountered.  Maybe add more robust email address validation, because it is a standard Go package and easy to use.

davidby-influx (approved):
LGTM

PR #2658, opened 2021-12-06
build: larger cgo stack

docmerlin:
This increases the cgo stack size of builds.

### DIFF
--- a/build.py
+++ b/build.py
@@ -535,24 +535,13 @@ def build(version=None,
             build_command += "-race "
         if len(tags) > 0:
             build_command += "-tags \"{}\" ".format(' '.join(tags))
-        if "1.4" in get_go_version():
-            if static:
-                build_command += "-ldflags=\"-s -X main.version {} -X main.branch {} -X main.commit {} -X main.platform OSS\" ".format(version,
-                                                                                                                  get_current_branch(),
-                                                                                                                  get_current_commit())
-            else:
-                build_command += "-ldflags=\"-X main.version {} -X main.branch {} -X main.commit {} -X main.platform OSS\" ".format(version,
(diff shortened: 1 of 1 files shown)


build.py
```diff
@@ -535,24 +535,13 @@ def build(version=None,
...
-                build_command += "-ldflags=\"-s -X main.version={} -X main.branch={} -X main.commit={} -X main.platform=OSS\" ".format(version,
-                                                                                                                  get_current_branch(),
-                                                                                                                  get_current_commit())
-            else:
-                build_command += "-ldflags=\"-X main.version={} -X main.branch={} -X main.commit={} -X main.platform=OSS\" ".format(version,
+        build_command += "-ldflags=\""
+        if static:
+            build_command +="-s "
+
+        build_command += "-extldflags \\\"-fno-PIC -Wl,-z,stack-size=8388608\\\"  -X main.version={} -X main.branch={} -X main.commit={} -X main.platform=OSS\" ".format(version,
```
lesam, line 544: This might be clearer with triple-quotes `"""python triple quotes for my raw string so I don't have to "escape" backslash \ ..."""`

lesam:
Did we validate that this is fixing the problem we saw with segfaults?

docmerlin:
We have not, I was unable to reproduce them however.

build.py
docmerlin: That is incorrect, in python to make it not need to escape you can use a an `r` before the string, example an `r'\'` is a slash.  Triple quotes just allow for multi-line strings, but they still have to be escaped.

lesam: Ah, TIL - my python is rusty. Thanks!

lesam (approved):

PR #2661, opened 2021-12-07
feat(topics): make topic queue length configurable

docmerlin:
This is a tuning param to make the topic queue length configurable.  This is useful if very large amounts of alerts must go off due to scheduled maintenance etc such as a windows update.

### DIFF
--- a/alert/topics.go
+++ b/alert/topics.go
@@ -11,19 +11,25 @@ import (
 )
 
 const (
-	// eventBufferSize is the number of events to buffer to each handler per topic.
-	eventBufferSize = 5000
+	// DefaultEventBufferSize is the default number of events to buffer to each handler per topic.
+	DefaultEventBufferSize = 5000
+	MinimumEventBufferSize = 1000
 )
 
 type Topics struct {
-	mu sync.RWMutex
-
-	topics map[string]*Topic
+	mu              sync.RWMutex
+	eventBufferSize int
+	topics          map[string]*Topic
 }
 
-func NewTopics() *Topics {
+// NewTopics creates a new Topics struct with a minimum bufferSize of 500.
+func NewTopics(bufferSize int) *Topics {
+	if bufferSize < MinimumEventBufferSize {
+		bufferSize = DefaultEventBufferSize
+	}
 	s := &Topics{
-		topics: make(map[string]*Topic),
+		eventBufferSize: bufferSize,
+		topics:          make(map[string]*Topic),
 	}
 	return s
 }
@@ -54,7 +60,7 @@ func (s *Topics) RestoreTopic(id string, eventStates map[string]EventState) {
 	defer s.mu.Unlock()
 	t, ok := s.topics[id]
 	if !ok {
-		t = newTopic(id)
+		t = s.newTopic(id)
 		s.topics[id] = t
 	}
 	t.restoreEventStates(eventStates)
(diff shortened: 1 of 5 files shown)


alert/topics.go
```diff
@@ -11,19 +11,23 @@ import (
...
-
-	topics map[string]*Topic
+	mu              sync.RWMutex
+	eventBufferSize int
+	topics          map[string]*Topic
 }
 
-func NewTopics() *Topics {
+func NewTopics(bufferSize int) *Topics {
+	if bufferSize <= 0 {
```
davidby-influx, line 25: Do we want a larger minimum, like 1K or something?  I cold see a 3-byte buffer causing problems.

alert/topics.go
```diff
@@ -11,19 +11,23 @@ import (
...
-	mu sync.RWMutex
-
-	topics map[string]*Topic
+	mu              sync.RWMutex
+	eventBufferSize int
+	topics          map[string]*Topic
 }
 
-func NewTopics() *Topics {
+func NewTopics(bufferSize int) *Topics {
```
davidby-influx, line 24: Should this be a method of the Topics struct so the buffer size is automagically available?

alert/topics.go
```diff
@@ -385,10 +389,13 @@ type bufHandler struct {
 	wg       sync.WaitGroup
 }
 
-func newHandler(h Handler) *bufHandler {
+func newHandler(h Handler, bufferSize int) *bufHandler {
+	if bufferSize == 0 {
```
davidby-influx, line 393: Why the difference in safety checks: above you use `<= 0`

davidby-influx (approved):
A few non-blocking suggestions.

alert/topics.go
ayang64: +1 -- this is actually a bug imho since `bufferSize` is an `int`.  this would panic 100% if someone passed in, say, -1.

alert/topics.go
docmerlin: I don't think that would be useful, as it is only ever instantiated once at startup and never changed.

alert/topics.go
```diff
@@ -185,24 +191,24 @@ func PatternMatch(pattern, id string) bool {
...
+	sorted       []*EventState
 
 	collected *expvar.Int
 	statsKey  string
 
 	handlers []*bufHandler
 }
 
-func newTopic(id string) *Topic {
+func newTopic(id string, bufferLength int) *Topic {
```
davidby-influx, line 206: Sorry, this was the one I thought should be a method of the Topics struct, so `bufferLength` is magically available.  My mistake on `NewTopics()` being a method of the thing that it created.

alert/topics.go
```diff
@@ -385,10 +391,13 @@ type bufHandler struct {
 	wg       sync.WaitGroup
 }
 
-func newHandler(h Handler) *bufHandler {
+func newHandler(h Handler, bufferSize int) *bufHandler {
+	if bufferSize < MinimumEventBufferSize {
```
davidby-influx, line 395: Nice!

davidby-influx (approved):
One more suggestion because I was wrong the last review.

alert/topics.go
docmerlin: That's actually a good idea.

davidby-influx (approved):
LGTM

timhallinflux:
@docmerlin -- docs will need the name of the config parameter, along with its default value....and where it appears in the config file.  Essentially, the config documentation needs to be updated to describe how customers can set this.

PR #2664, opened 2021-12-10
build: Darwin fix

docmerlin:
Only adjust cgo stack size on linux.

### DIFF
--- a/build.py
+++ b/build.py
@@ -540,10 +540,11 @@ def build(version=None,
         build_command += "-ldflags=\""
         if static:
             build_command +="-s "
-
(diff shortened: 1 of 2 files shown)


build.py
```diff
@@ -540,10 +540,11 @@ def build(version=None,
         build_command += "-ldflags=\""
         if static:
             build_command +="-s "
-
-        build_command += r'-extldflags \"-fno-PIC -Wl,-z,stack-size=8388608\"  -X main.version={} -X main.branch={} -X main.commit={} -X main.platform=OSS" '.format(version,
-                                                                                                               get_current_branch(),
-                                                                                                               get_current_commit())
+        if platform == "linux":
```
lesam, line 543: Might be worth a FIXME comment to get the right args to bump up the stack on non-linux platforms?

lesam (approved):

build.py
docmerlin: The default on those platforms is already reasonable.  It was just linux musl that was too small.

PR #2665, opened 2021-12-10
feat: Better point ingestion

docmerlin:
This causes kapacitor to ignore writes to databases that it doesn't actually need to listen to.


docmerlin:
This appears to be the wrong PR, closing this.

PR #2677, opened 2022-03-09
feat(kafka): this adds SASL support to kafka alerts

docmerlin:
We added SASL support, and changed the kafka client to use sarama.
There is a lot of code added to maintain compatability.

This adds the following fields to the kafka stanza:
```
	SASLUsername  string `toml:"sasl-username" override:"sasl-username"`
	SASLPassword  string `toml:"sasl-password" override:"sasl-password"`
	SASLMechanism string `toml:"sasl-mechanism" override:"sasl-mechanism"`
	SASLVersion   *int   `toml:"sasl-version" override:"sasl-version"`

	// GSSAPI config
	SASLGSSAPIServiceName        string `toml:"sasl-gssapi-service-name" override:"sasl-gssapi-service-name"`
	SASLGSSAPIAuthType           string `toml:"sasl-gssapi-auth-type" override:"sasl-gssapi-auth-type"`
	SASLGSSAPIDisablePAFXFAST    bool   `toml:"sasl-gssapi-disable-pafxfast" override:"sasl-gssapi-disable-pafxfast"`
	SASLGSSAPIKerberosConfigPath string `toml:"sasl-gssapi-kerberos-config-path" override:"sasl-gssapi-kerberos-config-path"`
	SASLGSSAPIKeyTabPath         string `toml:"sasl-gssapi-key-tab-path" override:"sasl-gssapi-key-tab-path"`
	SASLGSSAPIRealm              string `toml:"sasl-gssapi-realm" override:"sasl-gssapi-realm"`

	// OAUTHBEARER config. experimental. undoubtedly this is not good enough.
	SASLAccessToken string `toml:"sasl-access-token" override:"sasl-access-token"`
```

Also, this fixes a long standing bug where our fnv-1a hasher was different from Sarama's


services/kafka/config.go
```diff
@@ -82,58 +84,59 @@ type WriteTarget struct {
...
 		switch target.PartitionAlgorithm {
-		case "crc32c", "":
-			balancer = &kafka.CRC32Balancer{}
+		case "crc32", "":
+			partitioner = func(topic string) kafka.Partitioner { return crcpartitioner{} }
 		case "murmur2":
-			balancer = &kafka.Murmur2Balancer{}
+			partitioner = func(topic string) kafka.Partitioner { return murmur2{} }
+		case "murmur3":
+			// because murmur3 is a new feature we don't need to do anything wild to keep compatibility
```
lesam, line 101: would be better to describe the old wildness above, than call out this as being normal.

services/kafka/hash.go
```diff
@@ -0,0 +1,116 @@
...
+// with earlier versions.
+func (b murmur2) Partition(msg *sarama.ProducerMessage, numPartitions int32) (int32, error) {
+	if numPartitions == 0 {
+		return 0, ErrNonPositivePartitions
+	}
+
+	// NOTE: the murmur2 balancers in java and librdkafka treat a nil key as
+	//       non-existent while treating an empty slice as a defined value.
+	if msg.Key == nil {
+		return rand.New(rand.NewSource(time.Now().UTC().UnixNano())).Int31n(numPartitions), nil
```
lesam, line 26: Building a new source every time seems excessive - should we construct it once and keep it in the murmur2 struct maybe?

services/kafka/hash.go
```diff
@@ -0,0 +1,116 @@
...
+
+// Partition here is based on the murmur2 balance function in github.com/segmentio/kafka-go to not break compatibility
+// with earlier versions.
+func (b murmur2) Partition(msg *sarama.ProducerMessage, numPartitions int32) (int32, error) {
+	if numPartitions == 0 {
+		return 0, ErrNonPositivePartitions
+	}
+
+	// NOTE: the murmur2 balancers in java and librdkafka treat a nil key as
+	//       non-existent while treating an empty slice as a defined value.
```
lesam, line 24: Is this consistent with the previous library behaviour that kapa was using? nil vs empty slices in go don't often have meaning, so I'm suprised the sarama parsing code would make a distinction, regardless of what the java representation is.

services/kafka/hash.go
```diff
@@ -0,0 +1,116 @@
...
+	const (
+		seed uint32 = 0x9747b28c
+		// 'm' and 'r' are mixing constants generated offline.
+		// They're not really 'magic', they just happen to work well.
+		m = 0x5bd1e995
+		r = 24
+	)
+
+	// Initialize the hash to a random value
+	h := seed ^ uint32(length)
```
lesam, line 54: Not truly random right?

services/kafka/hash.go
```diff
@@ -0,0 +1,116 @@
...
+
+	h ^= h >> 13
+	h *= m
+	h ^= h >> 15
+
+	return h
+}
+
+// crcpartitioner is a partitioner for crc32 we use it instead of just using sarama.NewCustomHashPartitioner(crc32.NewIEEE),
+// bo maintain compatibility, because the way that sarama distributes the result of the hash causes us to lose compatibility
```
lesam, line 88: should specify (I assume compatibility with segmentio?)

services/kafka/hash.go
```diff
@@ -0,0 +1,116 @@
...
+type crcpartitioner struct{}
+
+// Partition takes a message and partition count and chooses a partition
+// we made a custom partitioner, because
+func (crcpartitioner) Partition(msg *sarama.ProducerMessage, numPartitions int32) (int32, error) {
+	if numPartitions == 0 {
+		return 0, ErrNonPositivePartitions
+	}
+	// NOTE: the crc32 balancers in librdkafka don't differentiate between nil
+	//       and empty keys.  both cases are treated as unset.
```
lesam, line 98: Same - does this match kapa's previous behaviour? I assume we weren't using librdkafka previously.

services/kafka/hash_test.go
```diff
@@ -0,0 +1,76 @@
...
+		{key: "hello", numPartitions: 1000, want: 229, partitioner: murmur2{}},
+		{key: "hello", numPartitions: 0, want: 0, wantErr: true, partitioner: murmur2{}},
+		{key: "hello2", numPartitions: 1000, want: 907, partitioner: murmur2{}},
+		{key: "hello3", numPartitions: 1000, want: 759, partitioner: murmur2{}},
+		{key: "hello4", numPartitions: 1000, want: 299, partitioner: murmur2{}},
+		{key: "hello5", numPartitions: 1000, want: 841, partitioner: murmur2{}},
+		{key: "", numPartitions: 1000, want: 681, partitioner: murmur2{}},
+		{key: "hello5", numPartitions: 1, want: 0, partitioner: murmur2{}},
+	}
+	for _, tt := range tests {
```
lesam, line 60: Should we test the nil case as well?

services/kafka/sasl.go
```diff
@@ -0,0 +1,108 @@
...
+	SASLGSSAPIKerberosConfigPath string `toml:"sasl-gssapi-kerberos-config-path" override:"sasl-gssapi-kerberos-config-path"`
+	SASLGSSAPIKeyTabPath         string `toml:"sasl-gssapi-key-tab-path" override:"sasl-gssapi-key-tab-path"`
+	SASLGSSAPIRealm              string `toml:"sasl-gssapi-realm" override:"sasl-gssapi-realm"`
+
+	// OAUTHBEARER config. experimental. undoubtedly this is not good enough.
+	SASLAccessToken string `toml:"sasl-access-token" override:"sasl-access-token"`
+}
+
+// SetSASLConfig configures SASL for kafka (sarama)
+func (k *SASLAuth) SetSASLConfig(config *sarama.Config) error {
```
lesam, line 28: Rather than mutating, any chance we can just return the `config.Net.SASL` structure? That would make the call site more obvious.

services/kafka/service.go
```diff
@@ -73,36 +73,41 @@ func (w *writer) Open() {
...
+
+	if err != nil {
+		w.diagnostic.Error("error in kafka client shutdown", err)
+	}
 	w.wg.Wait()
 }
 
-// pollStats periodically reads the writer Stats and accumulates the results.
+// pollErrors periodically reads the writer Stats and accumulates the results.
 // A read operation on the kafka.Writer.Stats() method causes the internal counters to be reset.
```
lesam, line 99: Looks like this comment is out of data since we changed libraries?

services/kafka/service.go
```diff
@@ -73,36 +73,41 @@ func (w *writer) Open() {
 		w: w,
 	}
 	statsMap.Set(statWriteMessageCount, writeMessages)
-
 	w.ticker = time.NewTicker(time.Second)
```
lesam, line 76: Do we use this ticker anymore?

services/kafka/service.go
```diff
@@ -143,10 +148,12 @@ func (c *Cluster) WriteMessage(diagnostic Diagnostic, target WriteTarget, key, m
 	if err != nil {
 		return err
 	}
-	return w.kafka.WriteMessages(context.Background(), kafka.Message{
-		Key:   key,
-		Value: msg,
-	})
+	w.kafka.Input() <- &kafka.ProducerMessage{
+		Topic: target.Topic,
+		Key:   kafka.ByteEncoder(key),
```
lesam, line 153: Can we even get a nil key here?

services/kafka/service.go
```diff
@@ -210,7 +224,8 @@ func configChanged(old, new Config) bool {
 	return old.UseSSL != new.UseSSL ||
 		old.SSLCA != new.SSLCA ||
 		old.SSLCert != new.SSLCert ||
-		old.SSLKey != new.SSLKey
```
lesam, line 213: Have we tested that the config change for kafka works properly (that the new producer is correctly restarted?)

Edit: I see the automated test below but this seems like a big enough change to manually try it out as well.

lesam (changes requested):

lesam:
Looks good! mostly comment stuff, and the unused `ticker`

services/kafka/hash.go
docmerlin: The library we are using mimics librdkafka behavior and we promise librdkafka compat for it.

services/kafka/hash.go
docmerlin: The library we were using before made that distinction, because they were trying to mimic the default java client behavior.

server/server_test.go
```diff
@@ -10702,12 +10702,20 @@ func TestServer_AlertHandlers(t *testing.T) {
...
-				time.Sleep(2 * time.Second)
-				ts.Close()
+				defer ts.Close()
 				got, err := ts.Messages()
 				if err != nil {
 					return err
 				}
+
+				for i := 0; i < 100 && len(got) == 0; i++ {
+					time.Sleep(time.Second)
```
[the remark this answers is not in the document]
docmerlin: checking if the server has gotten the messages, we need to keep checking for the test until it gets them, otherwise the test will be flaky.

services/kafka/sasl.go
docmerlin: the config struct isn't clean when it is created, and defaults have already been populated in it by the constructor.

services/kafka/service.go
docmerlin: I doubt it.

services/kafka/hash.go
docmerlin: correct its just the default "random" way to start the seed for the algo. Nothing in what we do here is truly random.

server/server_test.go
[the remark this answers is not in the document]
docmerlin: ts.Messages() isn't guaranteed to have messages.  Its possible without the loop and sleep to get a flaky test.  The sleep just makes sure we are waiting till the messages show up.

services/kafka/config.go
```diff
@@ -82,58 +84,59 @@ type WriteTarget struct {
...
+			partitioner = kafka.NewHashPartitioner
 		default:
-			return kafka.WriterConfig{}, fmt.Errorf("invalid partition algorithm: %q", target.PartitionAlgorithm)
+			return cfg, fmt.Errorf("invalid partition algorithm: %q", target.PartitionAlgorithm)
 		}
-	} else {
-		balancer = &kafka.LeastBytes{}
+		cfg.Producer.Partitioner = partitioner
 	}
+	cfg.Producer.CompressionLevel = 0
```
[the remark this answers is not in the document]
docmerlin: Good eye. I needed to delete that line.

services/kafka/hash.go
```diff
@@ -0,0 +1,116 @@
...
+
+func (murmur2) RequiresConsistency() bool { return true }
+
+// murmur2 is taken from github.com/segmentio/kafka-go
+// We copy here because it was in a private method.
+// It is a go port of the Java library's murmur2 function.
+// https://github.com/apache/kafka/blob/1.0/clients/src/main/java/org/apache/kafka/common/utils/Utils.java#L353
+// Go port of the Java library's murmur2 function.
+// https://github.com/apache/kafka/blob/1.0/clients/src/main/java/org/apache/kafka/common/utils/Utils.java#L353
+func murmur2Hash(data []byte) uint32 {
```
gwossum, line 43: Does the Java library have any murmur2 test vectors we can use to make sure our port is correct?

services/kafka/config.go
```diff
@@ -82,58 +84,59 @@ type WriteTarget struct {
...
+			partitioner = newCRCPartitioner
 		case "murmur2":
-			balancer = &kafka.Murmur2Balancer{}
+			// can't use kafka.NewCustomHashPartitioner here either, because we need to maintain compatibility
+			partitioner = newMurmur2
+		case "murmur3":
+			partitioner = kafka.NewCustomHashPartitioner(murmur3.New32)
 		case "fnv-1a":
-			balancer = &kafka.Hash{}
+			partitioner = kafka.NewHashPartitioner
```
lesam, line 105: Is this hasher also the same as before? I didn't see tests for that.

services/kafka/hash_test.go
```diff
@@ -0,0 +1,76 @@
+package kafka
+
+import (
+	"reflect"
+	"testing"
+
+	"github.com/Shopify/sarama"
+)
+
+// This tests the nil case of the partitioners to ensure they are randomish.  We aren't testing for cryptografphic
```
lesam, line 10:
```suggestion
// This tests the nil case of the partitioners to ensure they are randomish.  We aren't testing for cryptographic
```

lesam (commented):
Just another question and a typo

services/kafka/config.go
docmerlin: Yes, the segmentio library claims it is the same.

lesam: This test doesn't pass for me:

```
func partitions(n int) []int {
	p := make([]int, n)
	for i := range p {
		p[i] = i
	}
	return p
}

func TestHash(t *testing.T) {
	old := kafka.Hash{}
	newP := sarama.NewHashPartitioner

	for i := 0; i < 1000; i++ {
		partitionN := 10
		ps := partitions(partitionN)
		topic := fmt.Sprintf("foo-topic-%d", i)
		key := fmt.Sprintf("foo-key-%d", i)

		oldPartition := old.Balance(kafka.Message{
			Topic:     topic,
			Partition: 0,
			Offset:    0,
			Key:       []byte(key),
			Value:     nil,
			Time:      time.Time{},
		}, ps...)

		newPartitioner := newP(topic)
		newPartition, err := newPartitioner.Partition(&sarama.ProducerMessage{
			Topic: topic,
			Key:   sarama.StringEncoder(key),
		}, int32(len(ps)))
		require.NoError(t, err)

		assert.Equal(t, int(newPartition), oldPartition)
	}
}
```

docmerlin: Then it was a bug in the old hasher, because we guarantee sarama compatibility.
https://github.com/influxdata/kapacitor/blob/4c2b9656fd7435c58ce1570c77058995b97b64f9/pipeline/alert.go#L2232

And in the docs.  
https://docs.influxdata.com/kapacitor/v1.6/event_handlers/kafka/#kafka-message-partitioning

server/server_test.go
```diff
@@ -11675,19 +11684,28 @@ stream
...
 				TICKscript: tick,
 				Status:     client.Enabled,
-			}); err != nil {
+			})
+			if err != nil {
 				t.Fatalf("%s: %v", kind, err)
 			}
 
+			if err != nil {
+				t.Fatalf(" Error deleting task: %s: %v", kind, err)
```
davidby-influx, line 11693: This looks like unreachable code.  The check for `err != nil` above means this can never execute.  Also, the error is in task creation, not task deletion.

services/kafka/config.go
```diff
@@ -82,58 +84,59 @@ type WriteTarget struct {
...
+	cfg.Metadata.Full = false // we only want to grab metadata for the topics we care about
+
 	if c.UseSSL {
-		t, err := tlsconfig.Create(c.SSLCA, c.SSLCert, c.SSLKey, c.InsecureSkipVerify)
+		var err error
+		cfg.Net.TLS.Enable = true
+		cfg.Net.TLS.Config, err = tlsconfig.Create(c.SSLCA, c.SSLCert, c.SSLKey, c.InsecureSkipVerify)
 		if err != nil {
-			return kafka.WriterConfig{}, err
+			return cfg, err
```
davidby-influx, line 122: Why return a config if there is an error?

services/kafka/config.go
```diff
@@ -82,58 +84,59 @@ type WriteTarget struct {
...
-		Balancer: balancer,
+	cfg.Producer.Flush.MaxMessages = c.BatchSize
+	if cfg.Producer.Flush.MaxMessages <= 0 {
+		cfg.Producer.Flush.MaxMessages = DefaultBatchSize
 	}
+	cfg.Producer.Flush.Frequency = time.Duration(c.BatchTimeout)
 
-	return baseConfig, nil
+	// SASL
+	c.SASLAuth.SetSASLConfig(cfg)
```
davidby-influx, line 138: This can return an error.  Probably better to return that than to hope that `Validate` returns the same error.

services/kafka/service.go
```diff
@@ -159,16 +160,23 @@ func (c *Cluster) writer(target WriteTarget, diagnostic Diagnostic) (*writer, er
...
-				kafka:   kw,
-				cluster: c.cfg.ID,
-				topic:   topic,
+				requestsInFlightMetric: metrics.GetOrRegisterCounter("requests-in-flight", wc.MetricRegistry),
+				kafka:                  kp,
+				cluster:                c.cfg.ID,
+				topic:                  topic,
+				diagnostic:             diagnostic,
 			}
 			w.Open()
```
davidby-influx, line 181: Can `w.Open` fail?  Should we check for an error?

services/kafka/service_test.go
```diff
@@ -33,7 +32,10 @@ func TestWriter_UpdateConfig(t *testing.T) {
 			PartitionAlgorithm: "",
 		}, []byte{1, 2, 3, 4}, []byte{1, 2, 3, 4}))
 
-	// Update the config in a way that requires shutting down all active writers.
 	c.UseSSL = true
 	require.NoError(t, cluster.Update(c))
+	diag.Close()
```
davidby-influx, line 37: Can any of the `Close` calls fail?  If so, let's `require.NoError` from them.

davidby-influx (changes requested):
Only must-fix are the unreachable code and the unchecked error from `SetSASLConfig`.  Other questions may lead to code changes, but some are based on my ignorance.

services/kafka/service_test.go
docmerlin: Only one of those can return an error, but good idea.

services/kafka/service.go
docmerlin: It can't fail, and there is no error returned

davidby-influx (approved):
Yay!

Issue #2659, opened 2021-12-06
Generate random ID for influxdbOutNode


RBrothersBSI:
Is there a way to generate a random ID through a lambda function or anything of the sort? Something Like

    |influxDBOut()
        .create()
        .database(outputDB)
        .retentionPolicy(outputRP)
        .measurement(outputMeasurement)
        .tag('alertName', name)
        .tag('triggerType', triggerType)
        .tag('acked', 'false')
        .tag('location', 'Building 1')
        .tag('id', rand.Intn(100))

There was an issue raised about this before but no answer. Any suggestions?

docmerlin:
Currently there isn't a way to make a random number.  We really should have one.

RBrothersBSI:
Yeah it would definitely be beneficial. I am pushing all of the alerts to an Influxdb instance and need a way to delete them. It would be much easier to be able to delete if we get generate a random UUID for each alert. Maybe coming soon? Fingers crossed.

docmerlin:
https://github.com/influxdata/kapacitor/pulls

RBrothersBSI:
Awesome! So in the example, it would be something like: 

|influxDBOut()
    .create()
    .database(outputDB)
    .retentionPolicy(outputRP)
    .measurement(outputMeasurement)
    .tag('alertName', name)
    .tag('triggerType', triggerType)
    .tag('acked', 'false')
    .tag('location', 'Building 1')
    .tag('id', rand())

?

PR #2680, opened 2022-03-11
feat(lambda): rand function for tick lambdas

docmerlin:


docmerlin:
closes #2659

tick/stateful/functions.go
```diff
@@ -1085,6 +1089,43 @@ func (c *count) Signature() map[Domain]ast.ValueType {
 	return countFuncSignature
 }
 
+func newRand() *rand {
+	return (*rand)(mwc.Rand())
+}
+
+// rand gives you a ranom number.  It needs to be initalized before being run
```
davidby-influx: "ranom" => "random"

davidby-influx (approved):
LGTM

tick/stateful/functions.go
docmerlin:
```suggestion
// rand gives you a random number.  It needs to be initialized before being run
```

PR #2683, opened 2022-03-15
security(TLS): properly depreciate TLS<1.2, DES and RC4 based ciphers

docmerlin:
DES based ciphers were depreciated in v1.6.3 but due to a bug this only applied
to when those ciphers were specified in the config file.
This fully removes them, and also removes RC4 based ciphers and TLS<1.2.


davidby-influx (approved):
LGTM

PR #2684, opened 2022-03-15
chore: update changelog

docmerlin:
Update the changelog to reflect the autobuilt changelog.  We really should have the changelog file just be a link to the uploaded changelog.


davidby-influx (approved):
So, commit numbers instead of issue numbers henceforth?

PR #2687, opened 2022-03-25
chore: update go to 1.18 and refactor to take advantage of some of it…

docmerlin:


circularqueue.go
```diff
@@ -84,17 +76,17 @@ func (q *timeMessageCircularQueue) Dequeue(n int) {
...
 		q.head = 0
 		q.tail = 0
 	}
 	return
 }
 
 // Peek peeks i ahead of the current head of queue.  It should be used in conjunction with .Len() to prevent a panic.
-func (q *timeMessageCircularQueue) Peek(i int) timeMessage {
-	if i < 0 || i >= q.l {
+func Peek[x any](q *CircularQueue[x], i int) x {
```
davidby-influx, line 88: Can this be a receiver function?

services/load/service.go
```diff
@@ -601,11 +601,11 @@ func (s *Service) removeHandlers() error {
 		return err
 	}
 	for _, id := range diff(s.handlers, loadedHandlers) {
-		pair := strings.Split(id, "/")
-		if len(pair) != 2 {
-			return errors.New("expected id to be topicID/handlerID")
+		topicID, handlerID, found := strings.Cut(id, "/")
+		if !found {
+			return errors.New("expected id to be topicID/handlerID but it was: \"" + id + "\"")
```
davidby-influx, line 606: Perhaps use `fmt.Errorf` here.

davidby-influx (changes requested):
A few questions, nothing major.

services/load/service.go
```diff
@@ -603,7 +603,7 @@ func (s *Service) removeHandlers() error {
 	for _, id := range diff(s.handlers, loadedHandlers) {
 		topicID, handlerID, found := strings.Cut(id, "/")
 		if !found {
-			return errors.New("expected id to be topicID/handlerID but it was: \"" + id + "\"")
+			return fmt.Errorf("expected id to be topicID/handlerID but it was: \"%s\"", id)
```
davidby-influx, line 606: non-blocking - I like `%q` over `\"%s\"` because it escapes bad characters and looks cleaner in the formatting string.

davidby-influx (approved):
LGTM
One non-blocking suggestion.

PR #2689, opened 2022-04-08
Upgrade influxql

docmerlin:


Gopkg.lock
```diff
@@ -1114,7 +1114,7 @@
     "github.com/influxdata/influxdb",
     "github.com/influxdata/influxdb/client",
     "github.com/influxdata/influxdb/client/v2",
-    "github.com/influxdata/influxdb/influxql",
```
lesam, line 1117: Why do we still have a Gopkg if we also have a `go.mod` ? Can we delete this file?

go.mod
```diff
@@ -233,8 +240,6 @@ require (
 
 replace gopkg.in/fsnotify.v1 => github.com/fsnotify/fsnotify v1.4.2
 
-replace github.com/influxdata/influxdb => github.com/influxdata/influxdb v1.1.4
```
lesam, line 236: 👍  Wow

server/server.go
```diff
@@ -12,11 +12,14 @@ import (
 	"sync"
 	"time"
 
-	"github.com/influxdata/influxdb/influxql"
+	"go.uber.org/zap"
+
+	"github.com/influxdata/influxdb/query"
```
lesam, line 17: Does kapa use `gci` to enforce import grouping yet?

task_master.go
```diff
@@ -267,6 +269,10 @@ type TaskMaster struct {
 	wg      sync.WaitGroup
 }
 
+func (tm *TaskMaster) WritePointsPrivileged(ctx tsdb.WriteContext, database, retentionPolicy string, consistencyLevel imodels.ConsistencyLevel, points []imodels.Point) error {
```
lesam, line 272: Are we adding this to implement an interface? Can we adjust whatever requires that interface to not require unused methods?

lesam (approved):
Wow, it's shocking how little changed here - score one for API stability

task_master.go
docmerlin: It is an interface in influxdb 1.9.6 where it is actually used.

Gopkg.lock
docmerlin: I think we can.   I'll make a separate PR to do that.

task_master.go
lesam: Does kapa allow `SELECT INTO` ? Do you get a good error message if you try to use that?

docmerlin: It does allow it but it just the same as having run it on the target influx machine

lesam (approved):

PR #2699, opened 2022-06-14
fix(flux): update flux to 0.171 for: 'interface {} is nil, not string'

docmerlin:
We had a bug in flux that was causing errors in Kapacitor.  This is an update to flux fix that.


task/fluxlocal/query.go
```diff
@@ -53,17 +54,31 @@ func (f *fluxQueryer) injectDependencies(ctx context.Context) context.Context {
...
-	return deps.Inject(ctx)
+	return dependency.Inject(ctx, deps)
 }
 
 func (f *fluxQueryer) Query(ctx context.Context, compiler flux.Compiler) (flux.ResultIterator, error) {
 	f.logger.Info("executed flux query")
-	ctx = f.injectDependencies(context.Background())
+	var span *dependency.Span
+	ctx, span = f.injectDependencies(context.Background())
+	defer span.Finish()
```
davidby-influx, line 64: Should you capture the error from `Finish` and return it from `Query` if there were no other errors inside `Query`?

Or is the `defer Finish` wrong here, because the caller of `Query` will call `Release` when it is done?

task/fluxlocal/query.go
```diff
@@ -53,17 +54,31 @@ func (f *fluxQueryer) injectDependencies(ctx context.Context) context.Context {
...
 	// inject the dependencies to the context.
-	return deps.Inject(ctx)
+	return dependency.Inject(ctx, deps)
 }
 
 func (f *fluxQueryer) Query(ctx context.Context, compiler flux.Compiler) (flux.ResultIterator, error) {
 	f.logger.Info("executed flux query")
-	ctx = f.injectDependencies(context.Background())
+	var span *dependency.Span
+	ctx, span = f.injectDependencies(context.Background())
```
davidby-influx, line 63: Why use a new `context.Background` when a `context.Context` is passed in?  Should that be used instead of being shadowed?

task/fluxlocal/query.go
```diff
@@ -53,17 +54,31 @@ func (f *fluxQueryer) injectDependencies(ctx context.Context) context.Context {
...
+		span:           span,
+	}, nil
+}
+
+type spannedResultIterator struct {
+	flux.ResultIterator
+	span *dependency.Span
+}
+
+func (s *spannedResultIterator) Release() {
```
davidby-influx, line 81: Two questions: can this be safely called multiple times (per the Flux interface documentation for `ResultIterator.Release`)?

Should the return code from `span.Finish` be stored to be returned from the `ResultIterator.Err` interface method?  This would require implementing the method and passing back the `s.ResultIterator.Err` if non-nil, or the `s.span.Finish` return if the underlying iterator had no error.

davidby-influx (changes requested):
Some questions from my ignorance, and a few changes requested.

task/fluxlocal/query.go
docmerlin: the defer is completely wrong,

task/fluxlocal/query.go
davidby-influx: `Span.Finish` throws away all the errors from its nested `io.Closer`s!!!

I am shocked.

davidby-influx (approved):
LGTM

PR #2717, opened 2022-09-12
fix(security): update go to 1.18.6

jdstrand:
This fixes the following CVEs in go:
* CVE-2022-32189 - encoding/gob & math/big: decoding big.Float and big.Rat can panic
* CVE-2022-27664 - net/http: handle server errors after sending GOAWAY
* CVE-2022-32190 - net/url: JoinPath does not strip relative path components in all circumstances


docmerlin (approved):

jdstrand:
Thanks for the review and merge!

PR #2718, opened 2022-09-13
refactor: refactored code for correctness and staticcheck

docmerlin:
we refactored code for correctness and staticcheck.
We ignored ST1005 to keep backwards compatibility in log messages

This allows us to use staticcheck in this project, and cleans up the code.


services/load/service.go
```diff
@@ -335,7 +335,7 @@ func (s *Service) loadTask(f string) error {
 	}
 	s.tasks[id] = true
 
-	return nil
+	return err
```
jeffreyssmith2nd, line 338: Won't err always be nil here?

services/load/service.go
```diff
@@ -472,7 +474,7 @@ func (s *Service) loadVars(f string) error {
 	}
 	s.tasks[id] = true
 
-	return nil
```
jeffreyssmith2nd, line 475: Same thing here?

udf/server_test.go
```diff
@@ -337,6 +337,7 @@ func TestUDF_SnapshotRestore(t *testing.T) {
 		req = <-u.Requests
 		if req == nil {
 			t.Error("expected snapshot message got nil")
+			return
```
jeffreyssmith2nd, line 340: Alternatively, you could do t.Fatal(...)

udf/server_test.go
```diff
@@ -352,6 +353,7 @@ func TestUDF_SnapshotRestore(t *testing.T) {
 		req = <-u.Requests
 		if req == nil {
 			t.Error("expected restore message got nil")
```
jeffreyssmith2nd, line 355: Same here

tick/ast/lex.go
```diff
@@ -513,7 +513,7 @@ func isSpace(r rune) bool {
 const durationUnits = "uµsmhdw"
 
 func isDurUnit(r rune) bool {
-	return strings.IndexRune(durationUnits, r) != -1
+	return !strings.ContainsRune(durationUnits, r)
```
jeffreyssmith2nd, line 516: Should this be negated?

udf/server_test.go
docmerlin: Not supposed to do a t.Fatal in a goroutine that isn't the main one.

services/load/service.go
docmerlin: The idea was to return err from the defer, but I removed that.

jeffreyssmith2nd (approved):

PR #2726, opened 2022-10-13
chore(build): update go version and fix builds

docmerlin:
This updates the version of go and also should fix builds, which are failing.

This required go fmt changes and some refactoring due to staticcheck.


davidby-influx (approved):
LGTM

PR #2765, opened 2023-05-15
chore: update Go to 1.20.4

tomklapka:
based on latest edge image: go1.20.4-147b512cba0dbd7ea86df9d4e3c288e36c11e05a


bnpfeife (approved):

jdstrand (approved):
This has a couple of things that don't seem strictly related to the update to 1.20.4, but they seem ok. If these were needed, it would've been nice to have additional explanation in the commit message. Thanks for the PR!

PR #2767, opened 2023-05-30
fix: HttpPost AlertTemplateFile and RowTemplateFile work with either …

TiagoMHCSantana:
…unix- or windows-like paths

### Required checklist
- [ ] Sample config files updated (both `/etc` folder and `NewDemoConfig` methods) (influxdb and plutonium)
- [ ] openapi swagger.yml updated (if modified API) - link openapi PR
- [ ] Signed [CLA](https://influxdata.com/community/cla/) (if not already signed)

### Description
As stated on the docs of the package path (https://pkg.go.dev/path), at the overview section, path package does not deal with backslashes and drive letters. Instead, one should use the filepath package (https://pkg.go.dev/path/filepath) to handle operating system specific path separators.

### Context
With this change, windows users will be able to run Kapacitor using either the alert template file or the row template file of the httppost service, without getting the error "must use an absolute path for alert-template-file".

### Affected areas (if applicable):
N/A

### Severity (optional)
upgrade at your leasure

### Note for reviewers:
Check the semantic commit type:
 - Feat: a feature with user-visible changes
 - Fix: a bug fix that we might tell a user “upgrade to get this fix for your issue”
 - Chore: version bumps, internal doc (e.g. README) changes, code comment updates, code formatting fixes… must not be user facing (except dependency version changes)
 - Build: build script changes, CI config changes, build tool updates
 - Refactor: non-user-visible refactoring
 - Check the PR title: we should be able to put this as a one-liner in the release notes

### DIFF
--- a/services/httppost/config.go
+++ b/services/httppost/config.go
@@ -5,7 +5,7 @@ import (
 	"encoding/json"
 	"net/url"
 	"os"
-	"path"
+	"path/filepath"
 	"strings"
 	"text/template"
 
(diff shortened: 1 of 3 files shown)


TiagoMHCSantana:
Hello @nathanielc , could you please take a look at this PR? It is quite simple, but fixes an annoying bug.

bnpfeife (approved):
LGTM!

PR #2803, opened 2024-03-19
fix: do not migrate events with empty ID

alespour:
### Description
Closes [#4984](https://github.com/influxdata/EAR/issues/4984)


davidby-influx (approved):
LGTM

PR #2823, opened 2024-06-11
feat: add missing deps for crossbuilder

bednar:
This PR add missing builders toolchain to our custom builder.

### Required checklist
- [x] Signed [CLA](https://influxdata.com/community/cla/) (if not already signed)

### DIFF
--- a/builder/Dockerfile_build
+++ b/builder/Dockerfile_build
@@ -8,25 +8,28 @@ MAINTAINER support@influxdb.com
 ARG DEBIAN_FRONTEND=noninteractive
 
 RUN apt-get -qq update && apt-get -qq install -y \
-    software-properties-common \
-    unzip \
+    autoconf \
+    automake \
+    clang \
+    gcc \
+    gcc-aarch64-linux-gnu \
+    git \
+    libtool \
     mercurial \
     make \
-    ruby \
-    ruby-dev \
-    rpm \
-    zip \
+    pkg-config \
     python \
     python-setuptools \
     python3 \
     python3-setuptools \
     python3-boto \
-    autoconf \
-    automake \
-    libtool \
+    ruby \
+    ruby-dev \
+    rpm \
+    software-properties-common \
+    unzip \
     wget \
-    git \
-    pkg-config
+    zip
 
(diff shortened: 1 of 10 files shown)


builder/Dockerfile_build
```diff
@@ -54,7 +60,7 @@ ENV PATH $GOPATH/bin:$PATH
 #
 # src: https://github.com/influxdata/edge/blob/4677c285014ac27727e5a1ae9bf2c1633afc6ea6/dockerfiles/cross-builder/install-rust.sh#L7
 #
-ENV RUST_LATEST_VERSION=1.78.0
+ENV RUST_LATEST_VERSION=1.63.0
```
jdstrand, line 63: I see that this had to be downgraded for flux. Note that https://github.com/influxdata/flux/pull/5484 updates the minimum version to 1.68 and allows building with 1.78.

I think it probably makes sense to pull in the latest flux (it is at 1.95.1 while kapacitor seems to be at 1.171.0). Flux has had a lot of various fixes and this would allows building with newer toolchains. However, I suspect this will require some work and should be down as a followup PR. Cc @powersj

jdstrand (commented):
Thanks for the PR! It's surprising to me how many hoops we need to jump through to build this now, but I understand that's the case and this is largely based on what we were already doing in the `edge` repo.

That said, looking at this PRs CircleCI checks, I don't see anything about ARM. Where can we see the new builds using the updated crossbuilder? (I feel like I'm missing something obvious...)

builder/Dockerfile_build
bednar: The current CI checks do not build the multiplatform packages. I’ve tested this by customizing the CI configuration, and you can review the results here: https://app.circleci.com/pipelines/github/influxdata/kapacitor/2313/workflows/78c876a7-3379-4297-af9c-41c3615db430/jobs/9151

jdstrand (approved):

builder/Dockerfile_build
bednar: I would prefer to keep Flux at version `1.63.0` to simplify this PR. The new version of Flux contains breaking changes that could potentially cause issues for customers. Am I correct in this assumption? Upgrading Flux should be treated as a separate task.

PR #2811, opened 2024-04-02
fix: bump flux to 0.194.5 to fix rust build failure

chenrui333:
redo https://github.com/influxdata/kapacitor/pull/2738

relates to:
- https://github.com/Homebrew/homebrew-core/pull/167599
- https://github.com/influxdata/flux/issues/5438

fixes:
- https://github.com/influxdata/influxdb/issues/24367
- https://github.com/influxdata/kapacitor/issues/2769
- https://github.com/influxdata/kapacitor/pull/2796

PR #2859, opened 2025-06-26
chore: update flux to build against rust 1.83

chenrui333:
chore: update flux to build against rust 1.83

----

- https://github.com/Homebrew/homebrew-core/pull/228225
- https://github.com/influxdata/kapacitor/pull/2811


chenrui333:
cc @bednar

PR #2889, opened 2026-03-30
feat: upgrade google.golang.org/api package

NguyenHoangSon96:
### Required checklist
- [ ] Sample config files updated (both `/etc` folder and `NewDemoConfig` methods) (influxdb and plutonium)
- [ ] openapi swagger.yml updated (if modified API) - link openapi PR
- [ ] Signed [CLA](https://influxdata.com/community/cla/) (if not already signed)

### Description
- Change from this truct `DefaultCredentialsOptions{}` in "google.golang.org/grpc@v1.79.3/credentials/google/google.go" file will make code in the "google.golang.org/api" package not working anymore, so we need to upgrade the "google.golang.org/api" to the latest version v0.273.0.

### Note for reviewers:
- This PR will fix this dependabot [PR](https://github.com/influxdata/kapacitor/pull/2887)


NguyenHoangSon96:
Hi @bnpfeife, @jdstrand 
Any reason why we should not update the "quay.io/influxdb/cross-builder" image environment variable "PROTOC_GEN_GO_VER" from "v1.34.1" to "v1.36.11"?
I'm asking because I'm upgrading  "grpc" package to "1.79.3" and It will need other packages to upgrade also.
Currently, the error in "static-code-checks" can be fixed by upgrading "PROTOC_GEN_GO_VER" to "v1.36.11" and re-run `go generate ./...` command.

jdstrand:
> Hi @bnpfeife, @jdstrand Any reason why we should not update the "quay.io/influxdb/cross-builder" image environment variable "PROTOC_GEN_GO_VER" from "v1.34.1" to "v1.36.11"? I'm asking because I'm upgrading "grpc" package to "1.79.3" and It will need other packages to upgrade also. Currently, the error in "static-code-checks" can be fixed by upgrading "PROTOC_GEN_GO_VER" to "v1.36.11" and re-run `go generate ./...` command.

More than just kapacitor uses the cross-builder, so if updating, would have to be sure the other repos are still ok. @bnpfeife, thoughts?

bednar (changes requested):
We have to use the version of protobuf which is in cross-builder.

